PR #27, opened 2018-04-13
Login screen/bookmarks

javiergonzper:
BUGS & IMPROVEMENTS

- [X] (1) No connection with server https://github.com/owncloud/ios-app/pull/27#issuecomment-385328933 [FIXED]
- [X] (2) Error in issue when URL is re-typed: https://github.com/owncloud/ios-app/pull/27#issuecomment-385365747 [FIXED]
- [X] (3) Error with no credentials https://github.com/owncloud/ios-app/pull/27#issuecomment-385368008 [FIXED]
- [X] (4) Empty file list if user name contains blanks https://github.com/owncloud/ios-app/pull/27#issuecomment-385369451 [FIXED]
- [X] (5) Bookmark deleted from list before confirming https://github.com/owncloud/ios-app/pull/27#issuecomment-385372071   [FIXED]
- [x] (6) Edit option on top-left is not used to edit https://github.com/owncloud/ios-app/pull/27#issuecomment-385376586 -> sepparate issue
- [X] (7) Changing credentials in web UI, makes the app show cached contents with no feedback to the user https://github.com/owncloud/ios-app/pull/27#issuecomment-385380860 [WONT FIX HERE] -> https://github.com/owncloud/ios-app/issues/35
- [x] (8) URL not changed in edit mode https://github.com/owncloud/ios-app/pull/27#issuecomment-385392145 [FIXED]
- [X] (9) No username https://github.com/owncloud/ios-app/pull/27#issuecomment-385880116 -> improvement2do
- [X] (10) App crashes attaching old server https://github.com/owncloud/ios-app/pull/27#issuecomment-385888162  [FIXED] . Needs to improve error from SDK
- [X] (11) Cells duplicated in edit mode https://github.com/owncloud/ios-app/pull/27#issuecomment-386252226 [FIXED]


ownCloud/Bookmarks/BookmarkManager.swift
```diff
@@ -113,6 +113,26 @@ class BookmarkManager: NSObject {
...
 	}
+
+    func removeAuthDataOfBookmark(_ bookmark: OCBookmark) -> OCBookmark {
+        
+        if bookmark.authenticationMethodIdentifier == OCAuthenticationMethodBasicAuthIdentifier {
+            let username = OCAuthenticationMethodBasicAuth.userName(fromAuthenticationData: bookmark.authenticationData)
+            let password = ""
+            
+            do {
+                try bookmark.authenticationData = OCAuthenticationMethodBasicAuth.authenticationData(forUsername: username, passphrase: password, authenticationHeaderValue: nil)
```
felix-schwarz: Don't use this. From the docs / comment for this method:

> Generates authentication data for basic auth without reaching out to the server. Useful for building tests. **Should not be used for anything other than implementing tests.**

javiergonzper: Ok I used it because do exactly what I need. "Generates authentication data for basic auth without reaching out to the server".
In that point I am deleting the password for a Basic credentials but I want to still storing the username.

I will use then the (ignoring errors):
```
- (void)generateAuthenticationDataWithMethod:(OCAuthenticationMethodIdentifier)methodIdentifier options:(OCAuthenticationMethodBookmarkAuthenticationDataGenerationOptions)options completionHandler:(void(^)(NSError *error, OCAuthenticationMethodIdentifier authenticationMethodIdentifier, NSData *authenticationData))completionHandler;
```

javiergonzper: The behavior of the Edit was implemented as you say. It just a problem on the behavior of the button "Delete Authentication Data".

I will remove it on the case of Basic Auth.

felix-schwarz: > A "Delete bookmark" button instead the "Delete Authentication Data" it is not necessary because we have a Delete button in the top right.
> 
> screen shot 2018-04-20 at 10 30 58

Right 😅

javiergonzper: Ops, ok. I will change the presentation 👌

nasli:
For edit bookmarks and keeps the uuid it was implemented a replace bookmark method but seem it was removed in some point if you find useful here https://github.com/owncloud/ios-app/pull/27/commits/9d3b65680fcaaa1ced15983e8de1043b866f321e#diff-46429d884fc491a91512336226403cfbR75
This was tested editing just the name of a bookmark and works fine.

jesmrec:
### (1)

Not able to connect with servers:

apr-30-2018 09-29-22


going back and adding ownCloud demo server, works only for the first time. If other server is added, no connection and no issue displayed

@javiergonzper @pablocarmu

jesmrec:
### (2)

Check these steps:

1. Input an https URL -> Certificate displayed 
2. Cancel certificate approval
3. Input a different URL, so in 1. step was typed another server
4. Click on continue

Current: It is shown the issue regarding the server in 1.
Expected: It is shown the certificate of the server in 3.

screen shot 2018-04-30 at 12 50 49

tested with iPhoneX v11.3

jesmrec:
### (3)

Improvement:

Letting the credential fields empty, the error feedbacked is "Review connection":

screen shot 2018-04-30 at 13 03 17

Do you think it is the best option? which reason to show it?

jesmrec:
### (4)

Maybe not an issue of authentication (@felix-schwarz), but related

Steps:

Log in with an account which username contains blanks, for example: "my account is mine", or "i am jesmrec"

Current: File list is shown empty
Expected: Correct file list

jesmrec:
### (5)

Steps:

1. Create a new bookmark (mo matter auth method)
2. Tap on "Disconnect" to go to the server list
3. Swipe left on the bookmark to display the "Delete" option

Current: The bookmark is removed at the time the confirmation Dialog to remove is shown. If you select "Cancel" (to keep the bookmark), the dialog is vanished but the bookmark is not already there.

Expected: The bookmark is only removed from the list when the user selects "Delete"

apr-30-2018 13-26-54

jesmrec:
### (6)

1. Add a bookmark
2. Tap on `Edit`, top left of the server list view.

Current:

screen shot 2018-04-30 at 13 45 55

Within such view the only option you can currently perform is deleting the bookmark. So the name `Edit` is confusing, because you can not edit anything. In order to edit, you have to swipe right the bookmark row.

Expected:

Open to discussion, different options:

- Remove the `Edit` option and set a long press action to open the mode of the screenshot above
- Rename to `More` or similar, that points to additional options related with the bookmarks in the list.

jesmrec:
### (7) improvement

Look at this scenario (file list feature involved)

In web UI, change password of the user:

Current: File list is shown from cache (`Contents from cache` displayed)
Expected:  User should be pointed to edit the credentials, or redirected anywhere to perform it (more ideas welcome).

If the user types an incorrect password in Edit mode, happens the same as described in improvement (3)

jesmrec:
### (8) 

Steps:

1. Swipe to edit an existing bookmark
2. Change URL for another valid one without, changing authentication method.
3. Tap on `connect`

Current: URL is not edited. Name is edited if it was changed.
Expected: URL edited and connection with new URL

tested in cases:

http -> http
http -> https
https -> http
https -> https

apr-30-2018 14-52-49

jesmrec:
### (9)

Improvement

I miss the username somewhere after authenticating. The user authenticates, but his/her username does not appear anywhere. The choice of using the server name exists, but i think it is not the best option.

What do you think about?

jesmrec:
### (10)

i have tried to attach an 8.2 server

`{"installed":true,"maintenance":false,"version":"8.2.6.2","versionstring":"8.2.6","edition":"Enterprise"}`

After typing the URL, the app crashes. As we discussed with @felix-schwarz, a minimum server version to be set for attached accounts, but crashing is not a good scenario.

felix-schwarz:
> ### (4)
> 
> Maybe not an issue of authentication (@felix-schwarz), but related

This could be an issue in the SDK. I'll check.

> Not sure if it was discussed before:
> 
> - Create an OAuth2 bookmark
> - Try to create a second one
> 
> You are redirected to the "Authorize" view, not posible to enter credentials. I think it should be pointed to the server/core as in Android: https://github.com/owncloud/android/pull/2157#issuecomment-384585121

Yes, we discussed this at length on RocketChat a while back. The consensus was that this needs to be addressed on the server side.

> ### (6)
> 
> 1. Add a bookmark
> 2. Tap on `Edit`, top left of the server list view.
> 
> Current:
> 
> screen shot 2018-04-30 at 13 45 55
> 
> Within such view the only option you can currently perform is deleting the bookmark. So the name `Edit` is confusing, because you can not edit anything. In order to edit, you have to swipe right the bookmark row.

Not saying it's good, but it's a convention on iOS that people are used to and familiar with. That label therefore shouldn't be changed. However, in "Edit" mode, tapping the bookmark should ideally allow the user to edit the bookmark.

> ### (7) improvement
> 
> Look at this scenario (file list feature involved)
> 
> In web UI, change password of the user. Then open the file list of any bookmark.
> 
> Current: File list is shown from cache (`Contents from cache` displayed)
> Expected:  User should be pointed to edit the credentials, or redirected anywhere to perform it (more ideas welcome).

As the Add/Edit branch also contains a changed view controller for displaying issues, the client view controller (with the file view in it) does not yet implement showing issues it gets from the core during connection.

Please file this as an issue of its own, so it can be addressed once "Login screen/bookmarks" has been merged. It's nothing this branch can/should fix.

> ### (9)
> 
> Improvement
> 
> I miss the username somewhere after authenticating. The user authenticates, but his/her username does not appear anywhere. The choice of using the server name exists, but i think it is not the best option.
> 
> What do you think about?

Ideally, if the user doesn't provide a name for the bookmark, it should default to "User @ Host.name".

> ### (10)
> 
> i have tried to attach an 8.2 server
> 
> `{"installed":true,"maintenance":false,"version":"8.2.6.2","versionstring":"8.2.6","edition":"Enterprise"}`
> 
> After typing the URL, the app crashes. As we discussed with @felix-schwarz, a minimum server version to be set for attached accounts, but crashing is not a good scenario.

I added a minimum version check to the SDK last week, defaulting to requiring at least version 9.0 or later. This branch is not yet using the latest SDK version, though. Updating it may or may not solve this issue already. Please send me a crash log if it doesn't, so I can have a look.

jesmrec:
Is this ready or different for iPads. When i removed an account from an iPad, the app crashes and i noticed the following in XCode:

`2018-05-02 10:25:22.689318+0200 ownCloud[3742:274644] *** Terminating app due to uncaught exception 'NSGenericException', reason: 'Your application has presented a UIAlertController (<UIAlertController: 0x7f855d00d600>) of style UIAlertControllerStyleActionSheet. The modalPresentationStyle of a UIAlertController with this style is UIModalPresentationPopover. You must provide location information for this popover through the alert controller's popoverPresentationController. You must provide either a sourceView and sourceRect or a barButtonItem.  If this information is not known when you present the alert controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.'`

CC @felix-schwarz @javiergonzper

felix-schwarz:
> Is this ready or different for iPads. When i removed an account from an iPad, the app crashes and i noticed the following in XCode:

Going by the error message, it's likely a bug on all devices.

felix-schwarz:
> > ### (4)
> > 
> > Maybe not an issue of authentication (@felix-schwarz), but related
> 
> This could be an issue in the SDK. I'll check.

This was caused by the SDK not percent-escaping the username when building the root path. Just fixed it in https://github.com/owncloud/ios-sdk/commit/7de69d534145d7bca4fa4c167facf7db3ef5f897 . Good catch! 💯

javiergonzper:
> (2)
> Check these steps:
> 
> Input an https URL -> Certificate displayed
> Cancel certificate approval
> Input a different URL, so in 1. step was typed another server
> Click on continue
> Current: It is shown the issue regarding the server in 1.
> Expected: It is shown the certificate of the server in 3.

Fixed 👍 

> (3)
> Improvement:
> 
> Letting the credential fields empty, the error feedbacked is "Review connection":

I added the information than comes from the SDK, now it could support any error:

screen shot 2018-05-02 at 12 03 31

I will continue with the following errors taking into account your previous comments

javiergonzper:
> (6)
> Add a bookmark
> Tap on Edit, top left of the server list view.
> Current:
> 
> 
> 
> Within such view the only option you can currently perform is deleting the bookmark. So the name Edit is confusing, because you can not edit anything. In order to edit, you have to swipe right the bookmark row.

My solution is change "Edit" -> "Order" (to just allow to change the order of the Bookmarks) and the remove the "Delete" option from the "Order" view.

This is a mockup:

screen shot 2018-05-02 at 13 53 44
screen shot 2018-05-02 at 13 53 48

javiergonzper:
> (9)
> Improvement
> 
> I miss the username somewhere after authenticating. The user authenticates, but his/her username does not appear anywhere. The choice of using the server name exists, but i think it is not the best option.
> 
> What do you think about?

Do you mean in Oauth?

javiergonzper:
> (10)
> i have tried to attach an 8.2 server
> 
> {"installed":true,"maintenance":false,"version":"8.2.6.2","versionstring":"8.2.6","edition":"Enterprise"}
> 
> After typing the URL, the app crashes. As we discussed with @felix-schwarz, a minimum server version to be set for attached accounts, but crashing is not a good scenario.
> 
> I added a minimum version check to the SDK last week, defaulting to requiring at least version 9.0 or later. This branch is not yet using the latest SDK version, though. Updating it may or may not solve this issue already. Please share a crash log if it doesn't, so I can have a look.

jesmrec:
(3) 

Fixed, but the errors from SDK should be more friendly, for example: if the user types incorrect credentials, right now it is displayed:

`The operation couldn't be completed (OC Error error 2.)`

Something more friendly like  `Credentials error` or other similar would fit better for whose users who gets frightened with technical errors. So, a non-technical translation could be included in the issue object yo be displayed in the app (just a suggestion).

I think it is something to improve in next iterations.

(6)

> Not saying it's good, but it's a convention on iOS that people are used to and familiar with. That label therefore shouldn't be changed. However, in "Edit" mode, tapping the bookmark should ideally allow the user to edit the bookmark.

You mean, the `Edit` option has to do the same action as the swipe right. Maybe we should merge them and avoid the mode mentioned in the screenshot. 

(7)

> Not saying it's good, but it's a convention on iOS that people are used to and familiar with. That label therefore shouldn't be changed. However, in "Edit" mode, tapping the bookmark should ideally allow the user to edit the bookmark.

i will, thanks for the clarification.

(9)

> Ideally, if the user doesn't provide a name for the bookmark, it should default to "User @ Host.name".

i agree with that. The point here is OAuth2, where i guess the username is not retrieved. Am i wrong? Any global fix for both auth methods here?

jesmrec:
(7)

>As the Add/Edit branch also contains a changed view controller for displaying issues, the client view controller (with the file view in it) does not yet implement showing issues it gets from the core during connection.

> Please file this as an issue of its own, so it can be addressed once "Login screen/bookmarks" has been merged. It's nothing this branch can/should fix.

but the Add/Edit bookmark branch is rebased against master, including the file list and its logic, so it should work with the issues in such branch. Is it correct or am i missing something?

CC @javiergonzper

javiergonzper:
> (7)
> 
> As the Add/Edit branch also contains a changed view controller for displaying issues, the client view controller (with the file view in it) does not yet implement showing issues it gets from the core during connection.
> 
> Please file this as an issue of its own, so it can be addressed once "Login screen/bookmarks" has been merged. It's nothing this branch can/should fix.
> 
> but the Add/Edit bookmark branch is rebased against master, including the file list and its logic, so it should work with the issues in such branch. Is it correct or am i missing something?
> 
> CC @javiergonzper

It should works. Uhm... maybe the issue it is no the list. I can debug it if you want.

jesmrec:
### (11)

@javiergonzper 

After the last fixes, look at these steps:

1. Create a new bookmark
2. Then, swipe to edit the bookmark
3. Set incorrect credentials or URL (force a OC Error 2)

Current: Authentication cells are duplicated
Expected: No duplication

tested with iPhoneX v11

jesmrec:
From me it is ok to move forward, taking in account the following considerations:

(6) -> i will open an issue to discuss about. For the moment, we keep the `Edit` option. I think that beta testers can give also their thoughts about.
(7) -> moved to https://github.com/owncloud/ios-app/issues/35
(9) -> same as (6). I will open an issue with such improvement.
(10) -> Another improvement2do

@felix-schwarz @pablocarmu, please code review before merging.

javiergonzper:
> This pull request, and BookmarkViewController.swift in particular:
> 
> is hard-coded for Basic Auth and OAuth2 instead of making use of the abstraction provided by the SDK

Where do you mean that it is hardcoded? Maybe here?:

> if preferredAuthMethods.count > 0 {
>             if OCAuthenticationMethod.registeredAuthenticationMethod(forIdentifier: preferedAuthMethod).type() == .passphrase {
>                 authMethod = OCAuthenticationMethodBasicAuthIdentifier
>             } else {
>                 authMethod = OCAuthenticationMethodOAuth2Identifier
>             }
>         }

Let me know how to get that info from the SDK.

> is thereby leading to various possible security issues and/or authentication not working down the road
> has structural issues that make the code hard to read and maintain
> breaks bookmark deletion by dropping vault removal from the source code, following the merge/rebase (?) with master
> adds several stray files and directories that aren't used anywhere

This branch was updated several times since the beginning. When I come back the project did not compile and I fixed it.

> has unresolved Swift Lint and Swift Compiler warnings regarding the force unwrapping of values

I asked @pablocarmu and he said me that we have to remove that concrete check. Because of that I did not spend time on that check

> BookmarkViewController has so many issues that I don't think its fixable. If @michaelstingl agrees, I'll re-write that one from scratch while reusing the many other (good) parts like IssuesViewController, in order to bring this feature to completion.

I am sorry if you think that but I tried to do my best after took that task started and break.
The Pull Request was open on the day that you asked in order to start to check it (13 Apr) and after three weeks of team work (develop, QA and bug fixing) did you notice that the code it nos good enough.

We have to cover many different possibilities on that view and because of that I asked about divide this view in several files (add/edit) as we did on the Old App (when we detected that one view is so huge to make all the things that a login view have to do) in order to make it more clear and you said "No".

jesmrec:
From the points that mentioned @felix-schwarz, the following one i think deserves some attention, so the other ones can be reviewed and fixed in the current branch (tell me if i am wrong):

> has structural issues that make the code hard to read and maintain

knowing such structural issues, can help to improve this one or the following ones. Let's work as a team, code reviewing should be the way to be all together in the same line. 

I think this would be profitable than re-writing and passing other QA.

felix-schwarz:
> > This pull request, and BookmarkViewController.swift in particular:
> > 
> > is hard-coded for Basic Auth and OAuth2 instead of making use of the abstraction provided by the SDK
> 
> Where do you mean that it is hardcoded? Maybe here?:

Please make sure to click on "11 hidden conversations. Load more." to see the entire feedback (that link is is easy to miss). I provide a lot of examples in the places I commented.

> Let me know how to get that info from the SDK.

It's all in the header file for OCAuthenticationMethod:

- getting the class for an authentication method:

```objc
+ (Class)registeredAuthenticationMethodForIdentifier:(OCAuthenticationMethodIdentifier)identifier; //!< Returns the OCAuthenticationMethod class for identifier
```

- getting info on the authentication method using the retrieved class:

```objc
#pragma mark - Identification
+ (OCAuthenticationMethodType)type;
+ (OCAuthenticationMethodIdentifier)identifier;
+ (NSString *)name;
```

- extracting info from authentication data using the retrieved class:

```objc
#pragma mark - Passphrase-based Authentication Only
+ (BOOL)usesUserName; //!< This authentication method uses a user name
+ (NSString *)userNameFromAuthenticationData:(NSData *)authenticationData; //!< Returns the user name stored inside authenticationData
+ (NSString *)passPhraseFromAuthenticationData:(NSData *)authenticationData; //!< Returns the passphrase stored inside authenticationData
```

There is also sample code with additional comments on how bookmark creation works at https://github.com/owncloud/ios-sdk/blob/master/doc/RECIPES.md#bookmark-setup-and-authentication

> I asked @pablocarmu and he said me that we have to remove that concrete check. Because of that I did not spend time on that check

Hmm. @pablocarmu showed me how to fix force_cast warnings here (unfortunately, the individual commit seems lost to squashing): https://github.com/owncloud/ios-app/commit/148bce127d42e8c9fc7729cc86a75bb2ccdb44f8

In a nutshell instead of

```swift
self.passwordTextFieldDidChange(newPassword: row.value as! String)
```

write

```swift
if let rowValue = row.value as? String {
    self.passwordTextFieldDidChange(newPassword: rowValue)
}
```

No force casting needed (as that can result in a crash when things go wrong).

> I am sorry if you think that but I tried to do my best after took that task started and break.
> The Pull Request was open on the day that you asked in order to start to check it (13 Apr) and after three weeks of team work (develop, QA and bug fixing) did you notice that the code it nos good enough.

14 days ago, on April 20, I did point out that Authentication Methods shouldn't be referenced directly, but the abstractions should be used instead (https://github.com/owncloud/ios-app/pull/27#discussion_r182974681):

> UI for authentication should not be tied to a specific method implementation (like Basic Auth), but instead target the type of authentication (passphrase-based, token-based), so authentication methods can be plugged into the SDK and all apps using it automatically benefit of it without changes

That advice wasn't followed.

I also provided feedback on the structure that should be pursued:

> - OCBookmark should be used as backing store in the view controller
>    - for new bookmarks, that backing store is a new OCBookmark instance
>    - for editing bookmarks, that backing store is a copy of the existing OCBookmark instance (the copy can be obtained f.ex. via `bookmarkCopyForEditing = [OCBookmark bookmarkFromBookmarkData:[bookmark bookmarkData]]`). This way, only changes to the authenticationData property are saved directly
> - input handlers for the respective fields can then write directly there
> - handlers should be able to call a single method that
>    - checks the contents of the bookmark
>    - updates, which fields are visible depending on the bookmark's content
>    - takes care of setting the first responder to the next required field in case none is currently having first responder status
> - only part where add/edit differences really need to be considered then is:
>    - wording of buttons in the navigation bar
>    - whether or not the "Remove authentication data" option should be displayed

I don't see that reflected in the code, either.

> We have to cover many different possibilities on that view and because of that I asked about divide this view in several files (add/edit) as we did on the Old App (when we detected that one view is so huge to make all the things that a login view have to do) in order to make it more clear and you said "No".

I didn't just say "No". I provided a [long, detailed reply](https://github.com/owncloud/ios-app/pull/27#discussion_r182974681) pointing out why this is a bad idea and also provided above blueprint for an implementation.

I didn't see a reply of yours asking questions about it or expressing a need for discussion, so I assumed that you'd make the changes accordingly.

If I can't assume that, my feedback is largely ignored, and I'm supposed to check every commit, however, what's the point of asking me to provide feedback?

felix-schwarz:
> From the points that mentioned @felix-schwarz, the following one i think deserves some attention, so the other ones can be reviewed and fixed in the current branch (tell me if i am wrong):
> 
> > has structural issues that make the code hard to read and maintain
> 
> knowing such structural issues, can help to improve this one or the following ones. Let's work as a team, code reviewing should be the way to be all together in the same line. 
> 
> I think this would be profitable than re-writing and passing other QA. This is the power of code peer reviewing and we must take advantage of it.

Examples for structural issues:
- https://github.com/owncloud/ios-app/pull/27#discussion_r185910609
- https://github.com/owncloud/ios-app/pull/27#discussion_r185917154
- the whole class being structured around specific authentication methods

This view controller has been in the works since March. It's May now. It has been discussed and feedback been given. A lot. I don't see a benefit in putting even more time into this at this point. I'd rather prefer spending 1-2 days implementing this from scratch using the structure I outlined [in a comment 2 weeks ago](https://github.com/owncloud/ios-app/pull/27#discussion_r182974681).

PR #33, opened 2018-05-02
feature/settings

pablocarmu:
## Settings Screen

The way i'm coding this is:

- I've created a parent settings screen called `SettingsViewController` which inherits from `StaticTableViewController`.

- Instead of coding all the logic in this view controller, I've split the settings into three different categories:
    - **Security**: In this category should be coded:
           - **frequency**:  the frequency wich the security methods should be prompted to the user, the options I proposed here are: _allways, 1 minute, 5 minutes, 30 minutes, 1 hour_ .
           - **Passcode**:  custom passcode, a new ViewController should be implemented for this, there is some open source libraries but I prefer to code this instead of using an external dependency.
           - **TouchID/FaceID**: if the device support any kind of biometric ID prompt the user with this option **only** if the passcode option is enabled.

    - **Instant Uploads**: In this category should be coded: 
           - **Photo uploads**: This should allow to instant upload photos. 
           - **Video  uploads**: This should allow to instant upload videos. 
           - **Background uploads**: Allow background uploads for the selected video or photo options. This option should be presented only if one of the video of photo uploads are selected.
           - **Wifi only uploads**: Like the above option, this should be presented only if one of the video of photo uploads are selected.
           - **Accout to upload**: This should present to the user a list of bookmarks you have in the app and select the account **you want to be the one** in which the instant uploads should be put.
           - **Path for instant uploads**: Here you can select the folder you want to be the container for the instant uploads.   
          
    - **Various Things**: Here we should code:
           - **Open source licenses**: Is required to put the license or a link to the license for every third party libreary we use. 
           - **About ownCloud**
           - **Feedback**
           - **More things...**
          
         
Coding each category into a separate class may look as an overhead, but this allow us to:

- Select what settings we want to be shown: Maybe in the future, some users/customers want to ban instant uploads, so this way in a single line of code you avoid the creation and presentation of this settings.
- Decouple the settings: Instead of one massive class with all the settings you have 3 classes (each one has a `StaticTableViewSection`) so you have a more decoupled settings system. This should make us able to have the code isolated and add in the future more settings/categories without make the parent settings view a massive class.

The issue to comment more settings or another way to abord this feature is linked to this pull request so if you have any requirements or any things to discuss please comment there and ping me (@pablocarmu) in the comment so we can get this pull request comment as clean as posible.


felix-schwarz:
@pablocarmu:
> - Decouple the settings: Instead of one massive class with all the settings you have 3 classes (each one has a `StaticTableViewSection`) so you have a more decoupled settings system. This should make us able to have the code isolated and add in the future more settings/categories without make the parent settings view a massive class.

Unless there's a reason not to do so, please consider making these classes subclasses of `StaticTableViewSection`.

ownCloud/Settings/SecuritySettingsSection.swift
```diff
@@ -0,0 +1,161 @@
...
+            self.frequency = .allways
+        }
+        self.passcodeEnabled = userDefaults.bool(forKey: SecuritySettingsPasscodeKey)
+        self.biometricalSecurityEnabled = userDefaults.bool(forKey: SecuritySettingsBiometricalKey)
+
+        super.init()
+        self.headerTitle = "Security".localized
+        self.identifier = SecuritySectionIdentifier
+
+        updateUI()
```
felix-schwarz: Instead of breaking the creation of the rows into methods, and then checking if they exist using their identifier lateron, you could generate all rows here in the init-method and save them in instance variables.

Benefits:
- you don't need to keep track of whether or not they were already created (=> they're all just there already, ready for use).
- you can use the row objects directly and don't have to find and fetch them through their identifiers (=> you also don't need to generate symbols for the identifiers (like f.ex. `SecurityBiometricsRowIdentifier`) then, simplifying the implementation further)
- to change the interface dynamically, you could either keep using add/remove methods on the `StaticTableViewSection` - or put together a new `rows` array and ask the tableView to reload just that section. Ideally, you could extend `StaticTableViewSection` with a "`replaceRows`" method that replaces the rows instance var with a passed in array, and triggers a section reload. I also noticed that `StaticTableViewSection` doesn't support "live" additions/removals and animation yet, so it may be a good idea to:
   - add animation support to `add(rows:)` (like `add(rows:animated:)` where animated defaults to false and only calls `insertRows(at indexPaths: [IndexPath], with animation: UITableViewRowAnimation)` if the `viewController` property is set)
   - add a `remove(rows:animated:)` method that does the same, just for removal
   - add `add(row:animated:)` and `remove(row:animated:)` convenience methods that wrap the passed row into an array and forward it to the `add(rows:animated:)` and `remove(rows:animated:)` methods.
   - in the future, an `insert(rows:indexes:animated:)` method could be added, too - and `add(rows:animated:)` be changed to become just a convenience method to call that. But if it's not needed at the moment, this can surely be saved until we need it.

ownCloud/Settings/MoreSettingsSection.swift
```diff
@@ -0,0 +1,115 @@
...
+                <a href="https://itunes.apple.com/es/app/owncloud/id543672169?mt=8">Download here</a>
+                """
+            self.sendMail(to: nil, subject: "Try ownCloud on your smartphone!", message: message)
+        }, title: "Recommend to a friend".localized, accessoryType: .disclosureIndicator)
+
+        privacyPolicyRow = StaticTableViewRow(rowWithAction: { (_, _) in
+            let url = URL(string: "https://owncloud.org/privacy-policy/")
+            let title = "Privacy Policy".localized
+            self.viewController?.navigationController?.pushViewController(WebViewController(url:url!, title: title), animated: true)
+        }, title: "Privacy Policy".localized, accessoryType: .disclosureIndicator)
```
felix-schwarz: I'm not sure this is the right privacy policy for the app. There's another privacy policy shown in the App Store (that can also be modified on a per-app basis, but needs to meet certain minimum requirements).

Also, opening that page in-app may be problematic under GDPR, as the user might not expect network access / a full browser session to occur there. If the app keeps showing a web page there, the user should probably be asked, if it's ok to show that page. (Disclaimer: I'm not a lawyer and this is not legal advice, just my two cents.)

We'll likely need some legal advice / legal guidance from ownCloud HQ to get this one right. @michaelstingl

ownCloud/Settings/MoreSettingsSection.swift
```diff
@@ -0,0 +1,115 @@
...
+
+            if message != nil {
+                mail.setMessageBody(message!, isHTML: true)
+            }
+
+            self.viewController?.present(mail, animated: true)
+        } else {
+            if let vc = self.viewController {
+                let alert = UIAlertController(title: "Please configure an email account".localized,
+                                              message: "You need to configure an email account first to be able to send emails".localized,
```
felix-schwarz: The sentence is missing a closing point at its end.

ownCloud/Settings/MoreSettingsSection.swift
```diff
@@ -0,0 +1,115 @@
...
+                                              preferredStyle: .alert)
+
+                let okAction = UIAlertAction(title: "Ok", style: .default, handler: nil)
```
felix-schwarz: On Apple platforms, "Ok" should always be "OK".

ownCloud/Settings/SecuritySettingsSection.swift
```diff
@@ -0,0 +1,177 @@
...
+import LocalAuthentication
+
+// MARK: - Security UserDefaults keys
+public let SecuritySettingsKey: String = "security-settings"
+public let SecuritySettingsFrequencyKey: String =  "security-settings-frequency"
+public let SecuritySettingsPasscodeKey: String = "security-settings-usePasscode"
+public let SecuritySettingsBiometricalKey: String = "security-settings-useBiometrical"
+
+// MARK: - Section identifier
+private let SecuritySectionIdentifier: String = "settings-security-section"
```
felix-schwarz: Please consider just using it in place if an identifier is only used once and private to a class.

ownCloud/Settings/MoreSettingsSection.swift
```diff
@@ -0,0 +1,115 @@
...
+ * You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+ *
+ */
+
+import UIKit
+import WebKit
+import MessageUI
+
+// MARK: - Section identifier
+private let MoreSectionIdentifier: String = "settings-more-section"
```
felix-schwarz: Please consider just using it in place if an identifier is only used once and private to a class.

ownCloud/Settings/SecuritySettingsSection.swift
```diff
@@ -0,0 +1,177 @@
...
+                    radioButtons.append([frequency.toString() : frequency.rawValue])
+                }
+
+                frequencySection.add(radioGroupWithArrayOfLabelValueDictionaries: radioButtons, radioAction: { (row, _) in
+                    if let rawFrequency = row.value! as? Int, let frequency = SecurityAskFrequency.init(rawValue: rawFrequency) {
+                        self.frequency = frequency
+                        self.frequencyRow?.cell?.detailTextLabel?.text = frequency.toString()
+                    }
+                }, groupIdentifier: "frequency-group-identifier", selectedValue: SecurityAskFrequency.always.rawValue, animated: true)
+
```
felix-schwarz: If I pick a duration for the passcode lock (i.e. 30 mins) and go back to the settings screen, it shows the 30 mins. But if I tap on that entry again, the selection defaults back to "Always".

You can either use `func setSelected(_ value: Any, groupIdentifier: String)`to change the value after creation - or pass in the current setting as `selectedValue` during creation.

ownCloud/Settings/SettingsViewController.swift
```diff
@@ -0,0 +1,37 @@
...
+import UIKit
+
+class SettingsViewController: StaticTableViewController {
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        self.navigationItem.title = "Settings".localized
+
+        let standardUserDefaults = UserDefaults.standard
```
felix-schwarz: IIRC we'll need to grab a different UserDefaults instance via UserDefaults(suiteName: appGroupIdentifier) here. However, since it'll be needed in more places, I'd really like to add that to OCAppIdentity, so SettingsViewController can grab it from there in the future. 

For now, please use
```swift
// TODO: Use OCAppIdentity-provided user defaults in the future
let userDefaults = UserDefaults(suitename: OCAppIdentity.sharedAppIdentity().appGroupIdentifier)
```

ownCloud/Settings/UploadsSettingsSection.swift
```diff
@@ -0,0 +1,234 @@
...
+            let videoUploadIndex = rows.index(of: videosRow!)!
+
+            if !rows.contains(videosWifiOnlyRow!) {
+                insert(row: videosWifiOnlyRow!, at: videoUploadIndex + 1, animated: true)
+                insert(row: videosSelectedPathRow!, at: videoUploadIndex + 2, animated: true)
+            }
+
+        } else {
+            isVideoWifiOnlyUploadsEnabled = false
+            userDefaults.removeObject(forKey: UploadsVideoSelectedPathKey)
```
felix-schwarz: Please move this to the willSet/didSet closures for isVideoUploadEnabled, too, to keep UI code and defaults manipulation separate.

ownCloud/UI Elements/StaticTableViewSection.swift
```diff
@@ -182,4 +182,12 @@ class StaticTableViewSection: NSObject {
 
 		return nil
 	}
+
+    private func selectAnimation(_ animated: Bool) -> UITableViewRowAnimation {
```
felix-schwarz: This doesn't seem to be used anywhere. Please remove it.

ownCloud/UIKit Extensions/LAContext+Extension.swift
```diff
@@ -0,0 +1,34 @@
...
+import LocalAuthentication
+
+extension LAContext {
+
+    func supportedBiometricsAuthenticationNAme() -> String? {
+
+        if  canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil) {
+            switch self.biometryType {
+            case .faceID : return "Face ID"
+            case .touchID: return "Touch ID"
```
felix-schwarz: Please add ".localized" to make use of localizations in Localizable.strings.

ownCloud/UIKit Extensions/LAContext+Extension.swift
```diff
@@ -0,0 +1,34 @@
...
+ * For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
+ * You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+ *
+ */
+
+import LocalAuthentication
+
+extension LAContext {
+
+    func supportedBiometricsAuthenticationNAme() -> String? {
```
felix-schwarz: Tiny typo here: "NAme" vs "Name" at the end of the method.

ownCloud/Settings/UploadsSettingsSection.swift
```diff
@@ -0,0 +1,234 @@
...
+            if !rows.contains(photosWifiOnlyRow!) {
+
+                insert(row: photosWifiOnlyRow!, at: photoUploadIndex + 1, animated: true)
+                insert(row: photosSelectedPathRow!, at: photoUploadIndex + 2, animated: true)
+            }
+
+        } else {
+            isPhotoWifiOnlyUploadsEnabled = false
+            remove(rows: [photosWifiOnlyRow!, photosSelectedPathRow!], animated: true)
+            userDefaults.removeObject(forKey: UploadsPhotoSelectedPathKey)
```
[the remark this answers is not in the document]
pablocarmu: What I'm going to do here is to remove the line:
```Swift
userDefaults.removeObject(forKey: UploadsPhotoSelectedPathKey)
```

Should be enough setting the property ```isPhotoWifiOnlyUploadsEnabled``` to false instead of deleting it.

ownCloud/Settings/GlobalSettingsViewController.swift
```diff
@@ -25,113 +25,113 @@ class GlobalSettingsViewController: StaticTableViewController {
```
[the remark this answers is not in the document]
pablocarmu: Maybe worth to keep this file in the repo as some kind of "recipes". It really helped me some times to understand how the StaticTableView... structure works. As we are open source maybe itcould be valuable for contributors.

ownCloud/Settings/MoreSettingsSection.swift
```diff
@@ -0,0 +1,115 @@
...
+        }, title: "Help".localized, accessoryType: .disclosureIndicator)
+
+        sendFeedbackRow = StaticTableViewRow(rowWithAction: { (_, _) in
+            self.sendMail(to: "apps@owncloud.com", subject: nil, message: nil)
+        }, title: "Send feedback".localized, accessoryType: .disclosureIndicator)
+
+        recommendRow = StaticTableViewRow(rowWithAction: { (_, _) in
+            let message = """
+                <p>I want to invite you to use ownCloud on your smartphone!</p>
+                <a href="https://itunes.apple.com/es/app/owncloud/id543672169?mt=8">Download here</a>
```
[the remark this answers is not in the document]
pablocarmu: Good point, I'll delete it.

ownCloud/Settings/MoreSettingsSection.swift
pablocarmu: Should I prompt the user with an AlerView asking to launch the webview or leave it as it is for the moment?

ownCloud/Settings/MoreSettingsSection.swift
```diff
@@ -16,15 +16,29 @@
...
 
+    var appVersion: String {
+        if let version = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String {
+            return version
+        }
+
+        return ""
+    }
+
+    var appBuildNumber: String {
```
felix-schwarz: Great idea to make these "properties".

ownCloud/Settings/SecuritySettingsSection.swift
```diff
@@ -164,6 +161,8 @@ class SecuritySettingsSection: SettingsSection {
 
             var rowsToRemove: [StaticTableViewRow] = []
             frequencyRow?.cell?.detailTextLabel?.text = SecurityAskFrequency.always.toString()
+            frequency = .always
```
felix-schwarz: Might want to move this to isPasscodeSecurityEnabled/willSet.

michaelstingl:
Suggested changes:

- Show Git hash (on bottom with version?)
- Hide/explain unfinished settings
- Open Links in SFSafariViewController
- Make entries like "Recommend to a friend" brandable/MDM

jesmrec:
A couple of minor details regarding the settings view

- In `Send feedback` option, the subject of the mail has a lack of one `)`
- In the bottom of the view you wrote ' Owncloud'  instead of `ownCloud` . The content of such label has to be brandable as well.

jesmrec:
Approved

PR #55, opened 2018-05-23
File list view sorting

pablocarmu:
This PR aims to add support in the file list screen to different sorting methods.

The way I'm coding this is:

- I've created an enum with the possible sorting options, in our case I've coded:

```Swift
    case alphabeticallyAscendant
    case alphabeticallyDescendant
    case type
    case size
```

Taking advantage of the Swift's enums, the `SortMethod` enum has a method called `comparator()` which return a `Comparator`. This allows us to simply call
```Swift 
SortMethod.alphabeticallyAscendant.comparator()
```
to get the comparator for each sort method option in the enum.

Inside this method, there is a switch that returns the correct `Comparator` for each case.

- Next, I'll create a custom view with three buttons:
       
     - ~~**Create folder button**: like in the files app, this button in this pull request is not connected with any action but in the future should present some screen for creating a folder.~~

     - **Sort button**: the title of this button should be __sort by <sort method>__ and should present an action sheet with the different available sorting methods.

     - ~~**Change the view to List/Grid**: again, like in the files app, this button should change the layout to a Grid/List layout. In this pull request button is not connected with any action.~~

- Then I'll connect the sort button with the action sheet presentation method and present all the possible options and finally, when the user selects some option, we should change the `comparator` of the query and then relaunch the query. 

____

bugs & improvements:

- [x] (1) iPad sort menu with crash https://github.com/owncloud/ios-app/pull/55#issuecomment-394637594 [FIXED]

### DIFF
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -55,12 +55,7 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		progressSummarizer = ProgressSummarizer.shared(forCore: inCore)
 
 		query?.delegate = self
-		query?.sortComparator = { (left, right) in
-			let leftItem = left as? OCItem
-			let rightItem = right as? OCItem
-
-			return (leftItem?.name.compare(rightItem!.name))!
-		}
+        query?.sortComparator = SortMethod.alphabeticallyAscendant.comparator()
 
 		query?.addObserver(self, forKeyPath: "state", options: .initial, context: observerContext)
 		core?.addObserver(self, forKeyPath: "reachabilityMonitor.available", options: .initial, context: observerContext)
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -111,8 +111,8 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		// Uncomment the following line to display an Edit button in the navigation bar for this view controller.
 		// self.navigationItem.rightBarButtonItem = self.editButtonItem
 
-        sortBar = SortBar(frame: CGRect(x: 0, y: 0, width: self.tableView.frame.width, height: 40))
-        sortBar?.sortButton.addTarget(self, action: #selector(presentSortOptions), for: .touchUpInside)
+		sortBar = SortBar(frame: CGRect(x: 0, y: 0, width: self.tableView.frame.width, height: 40), sortMethod: sortMethod)
+		sortBar?.delegate = self
 		sortBar?.sortButton.setTitle("sort by \(sortMethod.localizedName()) ▼", for: .normal)
 
         tableView.tableHeaderView = sortBar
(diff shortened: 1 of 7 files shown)


ownCloud/Client/SortBar.swift
```diff
@@ -0,0 +1,83 @@
...
+		for method in SortMethod.all {
+			let action = UIAlertAction(title: method.localizedName(), style: .default, handler: {(_) in
+				self.sortMethod = method
+			})
+			controller.addAction(action)
+		}
+
+		let cancel = UIAlertAction(title: "Cancel".localized, style: .cancel)
+		controller.addAction(cancel)
+		UIApplication.shared.keyWindow?.rootViewController?.presentedViewController?.present(controller, animated: true, completion: nil)
```
felix-schwarz: You could add a method to ask the delegate to present this view controller for you like:

```
func sortBar(_ sortBar: SortBar, presentViewController: UIViewController, animated: Bool, completionHandler: () -> Void)
```

to give the SortBar a way to present this and - in the future - other view controllers via its delegate.

ownCloud/Client/SortMethod.swift
```diff
@@ -0,0 +1,123 @@
...
+* You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+*
+*/
+
+import Foundation
+import ownCloudSDK
+
+typealias OCSort = Comparator
+
+public enum SortMethod: Int {
```
felix-schwarz: I still like this approach so much 👍

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -226,6 +229,8 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		// Make sure we don't request the thumbnail multiple times in that case.
 		if cell?.item?.versionIdentifier != newItem.versionIdentifier {
 			cell?.item = newItem
+		} else {
+			cell?.updateWith(newItem)
```
felix-schwarz: I assume this is about the issue you noticed where file name changes don't change eTag and fileID and therefore don't update the cell. With this change, however, the cell would always be updated (cell?.item = .. calls through to updateWith(item) as well).

To update the cell only if the contents or the filename changed, something like this should be sufficient:
```swift
if (cell?.item?.versionIdentifier != newItem.versionIdentifier) || (cell?.item?.name != newItem.name) {
   cell?.item = newItem
}
```

jesmrec:
### (1)

In iPad, open the file list and tap on the sort menu

Current: App crashes
Expected: options displayed

Tested with iPadAir v11.3

pablocarmu:
@jesmrec  https://github.com/owncloud/ios-app/pull/55#issuecomment-394637594 this is addressed in the last commit --> https://github.com/owncloud/ios-app/pull/55/commits/55f4d233c67abc4dcec860bc7d7bcebe372310f6 please check and test it.

PR #57, opened 2018-05-25
Bitrase/integration

javiergonzper:


javiergonzper:
@felix-schwarz I added an `Equatable` function on the `ProgresSummary` because without it bitrise do not pass:

`❌  /Users/vagrant/git/ownCloud/UI Elements/Progress/ProgressSummarizer.swift:22:8: type 'ProgressSummary' does not conform to protocol 'Equatable'`

I added this `Equatable` func, but maybe it should have an ID or something similar to check if two `ProgresSummary` are exactly the same.

PR #63, opened 2018-06-05
Updated the README.md

javiergonzper:
- Updated the image of the Build status to use the bitrise one
- Changed the real time chat from irc freenode to the open Rocketchat for collaborators

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@ Not yet available for production neither in the store, use previous iOS app:
 
 **Build status:**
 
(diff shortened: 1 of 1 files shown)


jesmrec:
A tiny change in the README, not in the set of changes of the current PR. 

We have the following sentence as presentation:

`Not yet available for production neither in the store, use previous iOS app:`

i think is not correct, i would change for something like:

`Not available in store yet. Current iOS app`: [](https://itunes.apple.com/app/owncloud/id543672169)

PR #64, opened 2018-06-05
Local search

pablocarmu:
_____

BUGS & IMPROVEMENTS

- [X] (1) Landscape weird effect https://github.com/owncloud/ios-app/pull/64#issuecomment-398660973 [FIXED]
- [X] (2) Search not cleared https://github.com/owncloud/ios-app/pull/64#issuecomment-398665920 [FIXED]
- [X] (3) Browsing back makes the search bar invisible https://github.com/owncloud/ios-app/pull/64#issuecomment-398669190 [FIXED]


ownCloud/Tools/UIDevide+UIUserInterfaceIdiom.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  UIDevide+UIUserInterfaceIdiom.swift
```
felix-schwarz: There's a typo in that file's name: UIDevi**d**e vs. UIDevi**c**e

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -116,6 +117,16 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		sortBar?.updateSortMethod()
 
         tableView.tableHeaderView = sortBar
+
+		searchController.searchResultsUpdater = self
+		searchController.obscuresBackgroundDuringPresentation = false
+		searchController.hidesNavigationBarDuringPresentation = true
+		searchController.searchBar.placeholder = "Search".localized
```
felix-schwarz: Since the search bar in the "Files" app seems to do a global search, but this one doesn't, we should maybe rename this to "Search this folder" to make the scope of the search clearer.

Please also add that string to Localizable.strings.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -116,6 +117,16 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+		navigationItem.searchController = searchController
+		navigationItem.hidesSearchBarWhenScrolling = false
```
felix-schwarz: I noticed Files app has the search bar present all the time, but I'm not sure it's also the right choice for us. It's not a code issue, of course, so this can stay as is. We should just track this in an issue maybe, and discuss it there.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -389,30 +404,36 @@ extension ClientQueryViewController : OCQueryDelegate {
...
-				self.tableView.reloadData()
 
 				switch query.state {
-					case .contentsFromCache, .idle:
-						if self.items?.count == 0 {
-							self.message(show: true, imageName: "folder", title: "Empty folder".localized, message: "This folder contains no files or folders.".localized)
+				case .contentsFromCache, .idle:
+					if self.items?.count == 0 {
+						if self.searchController.searchBar.text != "" {
+							self.message(show: true, imageName: "folder", title: "No matches".localized, message: "There is no results for this search".localized)
```
felix-schwarz: We should maybe use a different image for search. Do we have a SVG of a search icon somewhere else already @michaelstingl? One that we could reuse here to show when search didn't bring up any results?

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -389,30 +404,36 @@ extension ClientQueryViewController : OCQueryDelegate {
...
-						self.message(show: true, imageName: "folder", title: "Folder removed".localized, message: "This folder no longer exists on the server.".localized)
+				case .targetRemoved:
+					self.message(show: true, imageName: "folder", title: "Folder removed".localized, message: "This folder no longer exists on the server.".localized)
+					self.tableView.reloadData()
 
-					default:
-						self.message(show: false)
+				default:
+					self.message(show: false)
+					self.tableView.reloadData()
```
felix-schwarz: I'm not sure if calling `reloadData` is needed here.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -425,6 +446,36 @@ extension ClientQueryViewController : SortBarDelegate {
...
+}
+
+// MARK: - UISearchResultsUpdating Delegate
+extension ClientQueryViewController: UISearchResultsUpdating {
+	func updateSearchResults(for searchController: UISearchController) {
+		let searchText = searchController.searchBar.text!
+
+		let filterHandler: OCQueryFilterHandler = { (_, _, item) -> Bool in
+			if let item = item {
+				if item.name.lowercased().contains(searchText.lowercased()) {return true}
```
felix-schwarz: Please use `localizedCaseInsensitiveContains()` instead. We got a lot of bonus functionality for free this way. F.ex. searching for `o` should also match`ö`, `ó`, `ò` etc.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: It was called always for all the cases in line 392 before. What I did is to take in mind the search and not update the `tableView` for this specific use-case and maintain the call in the rest of the cases in the switch statement.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: I've tried
```Swift
if item.name.localizedCaseInsensitiveContains(searchText.localizedLowercase) {return true}
```

and 

```Swift
if item.name.localizedCaseInsensitiveContains(searchText) {return true}

```

and I can't achieve what you said:

> F.ex. searching for o should also matchö, ó, ò etc.

There is something that I'm missing?

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: Definitely, I'll change it.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: I've just copied the Files app design. Yes, we could open an issue and talk about it.
@michaelstingl  what do you think?

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: If you guys have the SVG file send it to me and I'll change it. Another option is to search for some SVG with public domain license.

SamuAlfageme: Core uses: 

screenshot 2018-06-12 18 04 00

https://github.com/owncloud/core/blob/master/core/img/actions/search.svg

pablocarmu: So if everyone is fine with it, I'll use it then.

ownCloud/Client/ClientQueryViewController.swift
michaelstingl: Fine for first iteration 👍 Let’s wait for user feedback and change later…

ownCloud/Client/ClientQueryViewController.swift
michaelstingl: 👍

pablocarmu: Here it's the updated screen:
image-1

felix-schwarz: This is great! :)

jesmrec:
functionally works fine, no problems found in the searching. The only issue to review is the location of the search bar related to the spinner when pulling down, that moves the bar to the cell below:

jun-13-2018 14-11-51

maybe better to set the spinner below the search bar in order not to move it? what do you think?

CC @michaelstingl @javiergonzper

pablocarmu:
This behavior is how iOS works, it's the default behavior in other apps so we should maintain it as it is to maintain the system consistency.

michaelstingl:
Files: Search bar stays on top (no spinner)
Mail: Search bar moves with pull-to-refresh gesture (with spinner)
Dropbox ("Home" tab): Search bar moves with pull-to-refresh gesture (no spinner)

pablocarmu:
The spinner is for pull-to-refresh. It has nothing to do with the local search feature.

> Files: Search bar stays on top (no spinner)

Files app does not support pull to refresh. If there are changes on your underlying files you should notify files app and then the files app asks you for the changes.

If we want to maintain the spinner we should follow iOS conventions and let the system place the spinner as it is right now (like mail app).

jesmrec:
For me it is OK the spinner. The effect of movement of the searching bar is a bit weird, imho.

javiergonzper:
The problem it is that the animation it is not smooth enough when the spinner is hidden. Looks like the search bar "jump" when the spinner is hidden.

jesmrec:
In the Mail app, the search bar is "integrated" in the view where the pull down is performed, with a better visual effect:

jun-13-2018 18-34-28

felix-schwarz:
Essentially what @pablocarmu said, plus:

- this is a UIKit bug
- I just spent 2 hours trying to work around the issue. I found a workaround, but it makes another UIKit bug visible.

We're using the refreshControl, searchController and the tableHeaderView (which seems to have been the place to assign the UISearchController pre-iOS 11, maybe that's the issue). There's a slight chance this is fixed in iOS 12. Can you check in the iOS 12 simulator @pablocarmu?

pablocarmu:
We have one iPhone 6s with iOS 12 beta 1 in the office. I’ll test this on it tomorrow’s morning and try to post a little video/gif here.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

felix-schwarz:
Thanks! What happens if you slide down, so the Search Bar is visible - and then tap on "Photos"?

pablocarmu:
> Thanks! What happens if you slide down, so the Search Bar is visible - and then tap on "Photos"?

jun-14-2018 08-46-59

felix-schwarz:
Thanks again! So it looks like they didn't fix this in iOS 12.

What happens if you put the following line back in?

```
navigationItem.hidesSearchBarWhenScrolling = false
```

Does it lead to an abrupt stop of the pull-to-refresh spinner animation on iOS 12, too?

What happens when navigating to a subfolder with that line?

Thanks!

jesmrec:
Fixed in iOS 11 devices.

felix-schwarz:
@pablocarmu and I just spent some time on Jitsi playing with the current state of affairs on an iOS 12 device and also go through possible workarounds. But while each work around can fix one of the issues, it breaks something in an other, unrelated place.

Reaching the conclusion that there's currently no way of getting everything work smoothly here, we agreed on taking a page from other apps, which put two icons on the upper right: one for local search (which could grow into a server-wide search when the server adds that capability) - and one for multi-selection.

If the UIKit bugs around using `UIRefreshControl` and `UISearchController` while also pushing new view controllers onto the `UINavigationController` get fixed in UIKit in the future, we can always change it back. For now, this makes for a workable solution.

jesmrec:
My review after latest changes:

### (1)

Weird effect in the pull down has been fixed with the new design, but only in portrait orientation. In landscape:

jun-20-2018 10-02-12

jesmrec:
### (2)

More landscape stuff

Steps:

1. Set device in landscape
2. Enter a search pattern. At least one folder should be included
3. Browse into the folder

Current: Search is not cleared. If you tap on (x) and enter a new search, it is not performed. Yo need to cancel and enter a new one.
Expected: Search cleared (at least in was in the first dev iteration).

jesmrec:
### (3)

Steps:

1. Enter a search pattern. At least one folder should be included
2. Browse into the folder (At this point (2) happens)
3. Cancel
4. Browse back

Current: Search bar vanished
Expected: Search bar available

jesmrec:
### (4)

Improvement:

Search bar is visible in empty folders, maybe not needed in this case.

jesmrec:
(4) was designed following files app design, so for the moment we can go ahead with it.

From my side is all ok.

Issue #11, opened 2018-03-16
Files app 


jesmrec:
__As user, i want to manage files using 3rp party apps through Files app__

Epic for all features related with Files App. Smaller issues will be created to build the feature in an incremental way

- [x] Research about performance
- [x] Research Swift vs Obj-C in File Provider to handle memory
- [x] Add extension Files app
- [x] New App Id
- [x] Update set up file
- [x] Show list of files 
- [x] Save files by opening from 3rd party
- [X] Create files from 3rd party app

Features related:

- [x] Browse through ownCloud folders (including multiaccount)
- [x] Open ownCloud file
- [x] Save opened file
- [x] Save in oC created file
- [x] Multiaccount https://github.com/owncloud/ios-app/issues/7#issuecomment-388300629
- [x] Basic operations: Copy, Duplicate, Move, Delete, ~~Share~~ , Download, Rename
- [ ] Copy files between storage providers https://github.com/owncloud/ios-app/issues/361
- [ ] oC Sharing via iOS Files app
- [ ] oC Favorites via iOS Files app
- [ ] Working Set ([developer docs](https://developer.apple.com/documentation/fileprovider/content_and_change_tracking/defining_your_file_provider_s_content#2897861), wait for major changes in FileProvider in Spring 2020, open PR: https://github.com/owncloud/ios-app/pull/338)
  - [ ] Recent items
  - [ ] iOS Tags
  - [ ] iOS Favorites
  - [ ] Recently deleted items

michaelstingl:
Multi account: https://github.com/owncloud/ios-app/issues/7#issuecomment-388300629

felix-schwarz:
Files app integration is implemented at this point.

PR #67, opened 2018-06-07
[feature/fileprovider] File Provider support

felix-schwarz:
### Description of the Pull Request
- Starting implementation of a file provider for the new ios-app
- WIP!

### Related issue
https://github.com/owncloud/ios-app/issues/11

____

BUGS & IMPROVEMENTS

- [X] (1) Removed accounts in Files App https://github.com/owncloud/ios-app/pull/67#issuecomment-432708892 [FIXED]
- [x] (2) zip files (maybe more extensions) points to an error https://github.com/owncloud/ios-app/pull/67#issuecomment-432713207
- [X] (3) Upload a new file from a 3rd party https://github.com/owncloud/ios-app/pull/67#issuecomment-432720003 [MOVED]
- [X] (4) Place of certificates within security settings https://github.com/owncloud/ios-app/pull/67#issuecomment-432929462 [FIXED]
- [X] (5) Blinking list file https://github.com/owncloud/ios-app/pull/67#issuecomment-433006672
- [x] (6) Download large file https://github.com/owncloud/ios-app/pull/67#issuecomment-433019897
- [x] (7) Duplication in files app
- [x] (8) Dragging out of oC https://github.com/owncloud/ios-app/pull/67#issuecomment-436049786
- [X] (9) Empty list of files in second bookmark of same server https://github.com/owncloud/ios-app/pull/67#issuecomment-436975945 [NOT ALWAYS REPRODUCIBLE]
- [x] (10) Rename with no connection https://github.com/owncloud/ios-app/pull/67#issuecomment-440365223 [SDK OFFLINE]
- [X] (11) Duplication of folders https://github.com/owncloud/ios-app/pull/67#issuecomment-440369524 [FIXED]
- [X] (12) Copy from another location https://github.com/owncloud/ios-app/pull/67#issuecomment-440749262 [MOVED]


jesmrec:
I designed a first draft of test plan to check the funcionality developed here. In the test plan is included (basically):

- Tests about bookmark checking and browsing
- Tests about uploads and downloads via Files App 
- Tests about modification of files (via Files app or 3rd party)
- Tests about issues (lack of internet connection, lack of server connection)

i noticed a very huge work in the current branch, so it is very posible i am missing other features to test. If you please can share your thought about, i will complete a more accurate test plan.

@michaelstingl @felix-schwarz @pablocarmu

jesmrec:
I'm starting to check this using the test plan posted above. I will release my feedback as usual

jesmrec:
### (1)

Steps:

1. Add an account to ownCloud app
screen shot 2018-10-24 at 17 31 41

2. Go to Files App and enable this app 

screen shot 2018-10-24 at 17 32 25

3. Go to ownCloud app and remove the account

Current: No accounts attached to the app. But, the app is in Files App as an option 
screen shot 2018-10-24 at 17 34 12

By clicking it, appears blank

Expected: No accounts in oC app, no option in Files app

jesmrec:
### (2)

Try to download a .zip file inside a Files App ownCloud account. This message is displayed:

`Couldn’t communicate with a helper application`

after such error happens, it is not posible to download/open files inside Files App anymore. Files remain with an spinner forever.

In this GIF. `Archive` is a zip file:

oct-24-2018 17-44-35

jesmrec:
### (3)

To reproduce the following one, i used MS Office Word, but i guess other app that integrates Files App support will reproduce as well.

1. Create a new file
2. Save the file, using as location the ownCloud account in Files app

Current: Document appears in file list with 0 bytes. 
Expected: File with content

iOS12

jesmrec:
### (5)

While browsing in Files App, file list there blinks in the following way:

oct-25-2018 12-54-55

This is not something that happens always, but it is easier to reproduce the first time a folder is browsed after enabling the account in Files App

iOS12 iPhoneX

jesmrec:
### (6)

1. Open Files App
2. Browse into  ownCloud account which contains a huge file (i tested with one larger than 500Mb)
3. Tap on it

Current: File does not finish to download and after some time with the spinner, points to same error as https://github.com/owncloud/ios-app/pull/67#issuecomment-432713207

Expected: file dowloaded and opened 

iOS12 iPhoneX

felix-schwarz:
Thanks for your feedback @jesmrec!

Regarding the issues:
(1) Unfortunately it doesn't seem to be possible to make the FileProvider "invisible" when there are no accounts.
(2) This happens when the File Provider crashed. I couldn't reproduce the issue with the latest commits. Please retest with the latest commit.
(4) Fixed in the latest commits.
(5) I saw this myself a lot earlier, but not since completing some internal consolidation in the SDK. If you still find a way to reproduce it, I'd be very interested how, so I can try as well.

I've yet to try to reproduce (3) + (6).

michaelstingl:
> - [X] (1) Removed accounts in Files App https://github.com/owncloud/ios-app/pull/67#issuecomment-432708892 [WONT FIX]

Item in iOS Files app can't be removed, but other apps show a prompt to authenticate if no account is configured:



I've seen this in various app, so maybe there's something in the framework that can be used here?

michaelstingl:
`Files > Select > Duplicate (icon)` doesn't work. It creates identical files that disappear a few seconds later.

michaelstingl:
iOS Files Favorites and Tags aren't persistent.

As I understood, Favorites and Tags in iOS Files are independent from Favorites and Tags on the ownCloud server. Where does this information need to be stored?

Would it make sense to sync Favorites and Tags with the ownCloud server if there would be an API for it in the future?

michaelstingl:
Another report from a tester:

> Dragging out files from an ownCloud server out of Files app never completes. Ways to produce (on iPad):
> 
> 1. configure an ownCloud server in the app
> 2. activate the Files app ownCloud provider
> 3. open an app accepting file drops in split screen besides ownCloud ([Gladys](https://itunes.apple.com/app/gladys/id1257526927) is a good
> choice)
> 4. drag one or more files not downloaded yet (important!) to the second app
> 
> Expected result: Files app downloads the dragged files so that the target app can
> receive them.
> 
> Actual result: depending on your target app, placeholders appear, and Files app starts
> signalling a download (the dotted cloud icon is overlaid by the circular download
> activity indicator). The download never completes, leaving both apps in limbo.

jesmrec:
> iOS Files Favorites and Tags aren't persistent.

> As I understood, Favorites and Tags in iOS Files are independent from Favorites and Tags on the ownCloud server. Where does this information need to be stored?

> Would it make sense to sync Favorites and Tags with the ownCloud server if there would be an API for it in the future?

In case we want favorites and tags for oC, i would deal in a separate issue to face it also in the app itself

michaelstingl:
> In case we want favorites and tags for oC, i would deal in a separate issue to face it also in the app itself

Yeah, I'll open new issues – also for support of Favorites and Tags in Files App **independent from ownCloud** Favorites and Tags

jesmrec:
### (9)

1. Add a bookmark (basic auth)
2. Add a second bookmark from same server

Current: file list as empty folder. A pull2refresh needed to see the list
Expected: file list of the second bookmark

nov-08-2018 13-20-02

felix-schwarz:
@michaelstingl 

> Item in iOS Files app can't be removed, but other apps show a prompt to authenticate if no account is configured:
> 
> I've seen this in various app, so maybe there's something in the framework that can be used here?

With the latest commit, the ownCloud app's file provider now shows this, too:
bildschirmfoto 2018-11-08 um 16 33 45

felix-schwarz:
> ### (9)
> 1. Add a bookmark (basic auth)
> 2. Add a second bookmark from same server
> 
> Current: file list as empty folder. A pull2refresh needed to see the list
> Expected: file list of the second bookmark
> 
> nov-08-2018 13-20-02

I can't reproduce this. A minimal log file of this happening would be awesome @jesmrec !

felix-schwarz:
> ### (7)
> `Files > Select > Duplicate (icon)` doesn't work. It creates identical files that disappear a few seconds later.
> 
> 

What happens here under the hood is a bit wild:
- iOS asks the FP to download the file
- iOS then points the FP to its own file, asking it to upload it (with **same name** as original, to the **same folder** as the original)
- FP performs upload, returns with an error
- since the FP has no error UI yet, the FP automatically picks the default (cancel upload)
- at this point, the placeholder is removed again

The documentation says that in those cases, the FP should automatically pick a different name. I'll check our options there (including a shortcut where the FP figures the file already exists on the server, skips the download and does a COPY directly on the server).

felix-schwarz:
@michaelstingl 

> iOS Files Favorites and Tags aren't persistent.
 
Yes, both aren't yet implemented in the File Provider.

> As I understood, Favorites and Tags in iOS Files are independent from Favorites and Tags on the ownCloud server. Where does this information need to be stored?
> Would it make sense to sync Favorites and Tags with the ownCloud server if there would be an API for it in the future?

This is Apple's recommendation for storing tags:
```
 The file provider is asked to persist the new tag data on disk, then call the
 completion callback with the updated tagged data.  At a later point, the file
 provider should sync the new tag data to their server.
```

Tags are not provided as strings to the file provider. Instead, the file provider is given a binary blob of data whose content is private.

iOS expects the same binary blob back later.

Does the ownCloud server offer any way to store a blob of binary data alongside any item? (much like extended attributes in filesystems)

And this is for storing favorites:
```
 The file provider is asked to persist the new favorite rank on disk, then call
 the completion callback with the updated favorite rank.  At a later point, the
 file provider should sync the new favorite rank to their server.
```

For favorites, iOS doesn't provide a simple boolean `true`/`false`, it provides a number representing the item's "favorite ranking".

How are favorites implemented on the ownCloud server? Are they boolean `true`/`false` - or favorite rankings as well?

felix-schwarz:
@michaelstingl I added support for locally stored favorite ranking and tag data with the latest commit.

Regarding the contents of `tagData`: it's a `NSKeyedArchiver`-encoded `NSDictionary` that contains a version number and the tag labels alongside a color number:
- sync *to* server: it's possible to extract the tag labels and colors from the `tagData` without much effort
- sync *from* server: I did a quick test creating `tagData` on my own, using labels that were not previously created in the Files app and… it worked. The color dots matched the color numbers I provided and when I brought up the tags for the test item, the Files app showed my additional labels.

Conclusion:
- technically, there's nothing that keeps us from adding tag support that syncs with the server
- since the tagData is versioned, at least providing our self-composed tagData to iOS should continue to work in future iOS versions.
- if Apple decides to change the format or contents of the tagData, however, we may either no longer be able to extract tag information from it, or need to adapt our code to the new format. At that point, things could become a bit of a mess. The decision to go down that route would therefore be coupled with a plan to test this feature early and often on new iOS releases, so we can adapt in time before new iOS releases hit production devices.

felix-schwarz:
@michaelstingl Also regarding favorites, we could do a "lite" approach:

- store the favorite ranking locally
- add/remove favorite flag on the server depending on ranking (there's a special value `NSFileProviderFavoriteRankUnranked` that can be used to determine if an item is still a favorite based on the ranking)

felix-schwarz:
> @michaelstingl:
> 
> > ### (7)
> > `Files > Select > Duplicate (icon)` doesn't work. It creates identical files that disappear a few seconds later.
> > 
> 
> What happens here under the hood is a bit wild:
> 
> * iOS asks the FP to download the file
> * iOS then points the FP to its own file, asking it to upload it (with **same name** as original, to the **same folder** as the original)
> * FP performs upload, returns with an error
> * since the FP has no error UI yet, the FP automatically picks the default (cancel upload)
> * at this point, the placeholder is removed again
> 
> The documentation says that in those cases, the FP should automatically pick a different name. I'll check our options there (including a shortcut where the FP figures the file already exists on the server, skips the redundant upload and does a COPY directly on the server).

Happy to report that with the latest commit, Duplicate in the Files app now works.

jesmrec:
Review of the already reported bugs, using commit https://github.com/owncloud/ios-app/pull/67/commits/e29adb2f528e42416b13741960b7949aac24957f

(1) -> lack of accounts points to ownCloud app to create a new bookmark. Fixed.
(2)
(3) -> Still reproducible. I 'll try to get it with other 3rd party app apart of MS Office
(5) -> Still reproducible, i know this is a weird one.
(6) -> Still reproducible. I can ship you the file and the account to get it.
(7) -> By duplicating a file, i get a disconnection:

nov-16-2018 13-11-15



(8) -> Checked with different files, not always reproducible using Gladys to drag into. I will try to accurate more here.
(9) -> Still reproducible, this is the log caught in the app itself (starts in the moment of the second bookmark addition): [ownCloud App Log.txt](https://github.com/owncloud/ios-app/files/2589165/ownCloud.App.Log.1.txt)

I will perform another complete QA round

jesmrec:
New checks with latest commits:

app: `e0ec124`
ios-sdk: `1a82aa0`

(2) Open ZIP -> Fixed
(6) Download big file -> Fixed

jesmrec:
About (7) -> Duplication of files in Files App

Bug reported by @michaelstingl does not happen but:

- Files with specific extensions raise an error when they are duplicated in Files App. I have found with `png`and `pdf`:

nov-19-2018 14-37-20

- After duplicating, sometimes the screen blinks  (like in https://github.com/owncloud/ios-app/pull/67#issuecomment-433006672).

kopischke:
Apologies for asking slightly OT, but will there be a new TestFlight release for those who, like me, cannot simply build and load from source, in the foreseeable future?

jesmrec:
### (10)

1. Open an account in Files app
2. Switch device connection off
3. Rename an item

Current: Done button does not anything
Expected: Done button does something (save the new name and update when connection is back, alert of no connection...)

NOTE: iCloud saves locally the name and when connection is back, send the changes

jesmrec:
### (11) 

Duplication of folders in Files App points to an error 405 OCHTTPStatus ErrorDomain.

michaelstingl:
> Apologies for asking slightly OT, but will there be a new TestFlight release for those who, like me, cannot simply build and load from source, in the foreseeable future?

Thanks for asking. New TestFlight build with many fixes thanks to awesome user feedback planned for later this week…

jesmrec:
### (12)

1. Enter in another Files app location (f.ex iCloud)
2. Select a huge file (i used a 500MB one), or a big amount of files (i check with 60 files, over 300MB)
3. Move them to oC location (i used an OAuth2 account  to reproduce), with the option related in the bottom bar

Current: it does not finish never, and finally it crashes
Expected: File uploaded to oC from the another location

iPhoneX v12

felix-schwarz:
> ### (12)
> 1. Enter in another Files app location (f.ex iCloud)
> 2. Select a huge file (i used a 500MB one), or a big amount of files (i check with 60 files, over 300MB)
> 3. Move them to oC location (i used an OAuth2 account  to reproduce), with the option related in the bottom bar
> 
> Current: it does not finish never, and finally it crashes. From this point the app becomes inestable, with enqueued operations that are not finished, or `Authorization failed` alerts
> 
> Expected: File uploaded to oC from the another location
> 
> iPhoneX v12

I tried the following:
- upload 673 MB file to iCloud (free Big Buck Bunny film at 4K) on the Mac
- Basic Auth: drag & drop the file to test@demo.owncloud.org within the Files app:
    - the progress circle shows the Files app's progress downloading the file from iCloud
    - once Files app has downloaded the file, it tells the File Provider to import it. At that point, it replaces the progress circle with a cloud symbol
    - once the upload finishes, the File Provider reports success and the cloud symbol disappears
- OAuth2: drag & drop the file to a local ownCloud instance within the Files app:
    - Files app takes advantage of the already downloaded file
    - Files app directly tells the File Provider to import the file. At that point, it replaces the progress circle with a cloud symbol
    - once the upload finishes, the File Provider reports success and the cloud symbol disappears

Ideas:
- This may be an issue in the OAuth2 support that has yet to be identified. Please try reproducing this issue with a freshly created OAuth2 bookmark and see if it helps.
- I've yet to try with a lot of files. If it crashes, it could be related to the 15MB memory limit imposed by iOS. How many files did you try with? Where does it crash?

jesmrec:
(3) and (12) moved to next stage.

PR #92, opened 2018-06-27
Configure the release as ad-hoc

javiergonzper:
## Description
This Pull Request set the right config for adhoc distribution in order to be executed by Bitrise automatically

## Motivation and Context
The idea it is use Bitrise as continue deployment system to send Ad Hoc binaries to a certain number of users

## How Has This Been Tested?
- Executing the job of Bitrise
- An email will be received by the Bitrise members
- On the email there is an install button and you will be able to install the app on a device added on the Apple Developer Portal

## Screenshots

img_0223


michaelstingl:
Is there an option to configure the email text? Could the PR and the Git hash be included?

javiergonzper:
Yes @michaelstingl there is a possibility custom the emails adding a new step on our Bitrise workflow and using `Mailgun ` to send the email instead the automatic mail that send the `Deploy` step.
But the URL of Bitrise will be always the same and will looks like the screenshot that I uploaded.

Anyway it is not necessary the PR or the Git hash because on the mail/web we have the Bitrise build:
img_0224

With that number we can access on Bitrise to all the necessary information:
screen shot 2018-06-27 at 17 39 26

michaelstingl:
ah, great. works for me

felix-schwarz:
@ChrisEdS Please check if this collides with the Fastlane match settings.

michaelstingl:
@ChrisEdS is in training Wed to Fri. No blocker. Please merge, and if something breaks, we will fix it…

ChrisEdS:
As recommended by @michaelstingl, I will react if there is a problem. Your process has absolute priority.

PR #107, opened 2018-08-20
Fix/sortbar hidden when no items

pablocarmu:
## Description

This PR fixes the bug with the sort bar when there were no items inside a folder.

## Motivation and Context

When there are no items inside a folder, the sort bar was hidden by the "No items" view. So you can't tap on the create folder button.

## Types of changes

## Checklist:

_____

BUGS & IMPROVEMENTS

- [ ] Default name of a folder https://github.com/owncloud/ios-app/pull/107#issuecomment-415337603

### DIFF
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -160,8 +160,8 @@ class ClientQueryViewController: UIViewController {
 		view.addSubview(tableView)
 		NSLayoutConstraint.activate([
 			tableView.topAnchor.constraint(equalTo: sortBar.bottomAnchor),
-			tableView.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor),
-			tableView.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor),
+			tableView.leftAnchor.constraint(equalTo: view.leftAnchor),
+			tableView.rightAnchor.constraint(equalTo: view.rightAnchor),
 			tableView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor)
 			])
 
@@ -341,8 +341,8 @@ class ClientQueryViewController: UIViewController {
 
 				self.tableView.addSubview(rootView)
 
-				rootView.leftAnchor.constraint(equalTo: self.tableView.safeAreaLayoutGuide.leftAnchor).isActive = true
-				rootView.rightAnchor.constraint(equalTo: self.tableView.safeAreaLayoutGuide.rightAnchor).isActive = true
+				rootView.leftAnchor.constraint(equalTo: self.tableView.leftAnchor).isActive = true
+				rootView.rightAnchor.constraint(equalTo: self.view.rightAnchor).isActive = true
 				rootView.topAnchor.constraint(equalTo: self.tableView.safeAreaLayoutGuide.topAnchor).isActive = true
 				rootView.bottomAnchor.constraint(equalTo: self.tableView.safeAreaLayoutGuide.bottomAnchor).isActive = true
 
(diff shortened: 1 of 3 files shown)


jesmrec:
### (1)

Suggestion: as hint for the new folder, i would set `New folder`(as web UI) better than `Unknown folder`

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -19,9 +19,13 @@
 import UIKit
 import ownCloudSDK
 
-class ClientQueryViewController: UITableViewController, Themeable {
-	var core : OCCore?
-	var query : OCQuery?
+class ClientQueryViewController: UIViewController {
```
felix-schwarz: Why did you change the superclass from `UITableViewController` to `UIViewController` and set up the tableView yourself?

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -123,20 +132,46 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+			tableView.topAnchor.constraint(equalTo: sortBar.bottomAnchor),
+			tableView.leftAnchor.constraint(equalTo: view.leftAnchor),
+			tableView.rightAnchor.constraint(equalTo: view.rightAnchor),
+			tableView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor)
+			])
+
+		tableView.delegate = self
+		tableView.dataSource = self
+
+		self.navigationController?.navigationBar.setValue(true, forKey: "hidesShadow")
```
felix-schwarz: This is private API and shouldn't be used.

What effect does it have / why is it needed?

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -550,18 +590,43 @@ extension ClientQueryViewController : OCQueryDelegate {
 // MARK: - SortBar Delegate
 extension ClientQueryViewController : SortBarDelegate {
 	func sortBar(_ sortBar: SortBar, leftButtonPressed: UIButton) {
-		print("LOG ---> left button pressed")
+		let createFolderVC = NamingViewController(with: core, defaultName: "Unknown Folder".localized, stringValidator: { name in
```
felix-schwarz: Like @jesmrec I think that "New folder" is what a user would usually expect here.

ownCloud/Client/SortBar.swift
```diff
@@ -97,7 +97,7 @@ class SortBar: UIView, Themeable {
 	func applyThemeCollection(theme: Theme, collection: ThemeCollection, event: ThemeEvent) {
 		self.sortButton.applyThemeCollection(collection)
 		self.sortSegmentedControl.applyThemeCollection(collection)
-		self.applyThemeCollection(collection)
+		self.backgroundColor = Theme.shared.activeCollection.navigationBarColors.backgroundColor
```
felix-schwarz: Please use `collection` instead of `Theme.shared.activeCollection` here.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: The problem this PR solves is that when you don't have any files, the sort bar was hidden by the _No Files_ view. 

There were two approaches:

-  Make the _No Files_ view top constraint constant to be 40 pt (the height of the sort bar) to not hide it with this view. Then the scrolling effect to reload scrolls the sort bar under the _No files_ view and the effect was not beauty.

- Make the `ClientQueryViewController` be a child of `UIViewController` and make the sort bar part of the view controller view.
 The only downside of this is that you need to create the `tableView`  yourself and implement the `UITableViewDelegate` and `UITableViewDataSource`. This way you can have the sort bar independent of the `tableView` and present the _No Files_ view above only the `tableView` allowing the user to push the _create folder_ button and not making any rare scroll with the sort bar.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: The effect was to hide the line between the `NavigationBar` and the sortBar. It's only a visual effect so if this API shouldn't be used we can remove it without any problem.

ownCloud/Client/SortBar.swift
pablocarmu: I'll change it.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: Definitely right. 
I just copied here what files app does but _New Folder_ is a lot better. I'll change it.

jesmrec:
works for me. merging.

PR #113, opened 2018-09-04
Fix flexible header height in MoreViewHeader / cards

felix-schwarz:
- fixes an issue where the title in the header of a card view using a MoreViewController would be cut off.
- adds a useful NSLayoutConstraint extension to change the priority on the fly
- removes line_length limit from SwiftLint rules as a higher limit of 300 doesn't work - and the limit of 200 often makes for worse formatting

screenshot 2018-09-04 at 11 06 42


jesmrec:
Let's start a new review over the `More`

jesmrec:
In landscape, the options in the card are cut:

sep-06-2018 11-01-20

felix-schwarz:
@jesmrec Thanks!

Does that happen when
a) in landscape mode and tapping the "more-dots" .. or ..
b) switching from portrait to landscape mode when already in the "more" view?

jesmrec:
@felix-schwarz with the following steps i can reproduce it:

1.- Device in portrait. Open the card
2.- With card opened, switch to landscape -> Card is visible correctly
3.- Swipe the card to the top of the screen and release like here: https://github.com/owncloud/ios-app/pull/113#issuecomment-419020392

felix-schwarz:
@jesmrec Thanks!

I tried with an iPhone 6s, 6s Plus and X and couldn't reproduce the issue: 
owncloudcardresizingafterrotation

IIRC a previous commit had that issue. Can you try again with the latest commit?

jesmrec:
i will try

jesmrec:
i have tried out, and also reviewed your gif @felix-schwarz. Just in the moment you change the orientation to landscape, the `Delete` option is not visible, only the `Rename` one. This is the frame: 

screen shot 2018-09-28 at 12 52 02

After that, each time the card is pulled up/down, the options are correctly displayed.

Checking in iOS12, i noticed that in the reverse case (switching from landscape -> portrait): If the name is long, by pulling up the card, the options are not visible:

ezgif com-optimize

felix-schwarz:
@jesmrec With the latest commit, the issue should now be fixed and from my point of view can be merged.

jesmrec:
fixed. Moving forward...

Issue #112, opened 2018-08-28
File Viewing / Editing structure


felix-schwarz:
Here's my suggested approach for extensible, modular file viewer / editor support:

- Two primitives
	- thumbnail generator (future)
	- interactive: viewer / editor
	
- Based on extensions

- DisplayViewController: Viewer / Editor
	- Source as URL
		- allows viewing downloaded files
		- allows streaming from URL provided by SDK
	- Implemented as View Controllers
	- Indicate features via extension
		- streaming: ability to stream content / requirement of local file
		- editing/saving: capability of saving using a delegate

- DisplayHostController
	- designed to view a list of items
		- can be single file
		- can be several files (=> slideshow, quickly run through files)

	- taps extensions to determine and instantiate viewer/editor view controller
		- uses DisplayViewControllers as child view controller to display content
		
	- uses edge swipe gesture (recognizer) to navigate between files
	- shows thumbnails of the files at the bottom (like Photos app)

felix-schwarz:
Implemented.

Issue #85, opened 2018-06-20
[Previews] Image gallery


javiergonzper:
_As a user, i want the possibility see the images of a folder on the ownCloud iOS App_

### AC:
 - [ ] When the user select an image on the `File list` the gallery should be open showing the selected image
 - [ ] The gallery load all the images in the order of the `File list`
 - [ ] The gallery should have in memory the current image, the next and the previous one
 - [ ] The files should be cached
 - [ ] Pinch to zoom
 - [ ] The gallery should allow full screen with one tap on the screen and exit from background with one tap
- [ ] The image have to be loaded touching the limits of the screen on the higher size (including under `Navigation bar` and `Status bar` to prevent blinks when  enter on full screen)
- [ ] The title on the `Navigation bar` have to change once the new image are over the middle of the transition
- [ ] The transition should have a separation between images

### Mockups:
Image preview
<kbd>
![01 image preview loading](https://user-images.githubusercontent.com/1815050/41655644-bfdc1c24-748d-11e8-88bb-d6cadb872d4e.png)
</kbd>
<kbd>
![image preview 1](https://user-images.githubusercontent.com/1815050/41652473-6ded6ae4-7483-11e8-8b01-4ffccb916390.png)
</kbd>

Transition
<kbd>
![image preview 2 - transition](https://user-images.githubusercontent.com/1815050/41652542-aa1d93b8-7483-11e8-9736-6b89f51106ee.png)
</kbd>

Transition loading
<kbd>
![02 image preview transition loading](https://user-images.githubusercontent.com/1815050/41655621-ab24f044-748d-11e8-91f2-8ff9ac2c391f.png)
</kbd>

Image portrait
![image preview 3](https://user-images.githubusercontent.com/1815050/41652555-b9acf580-7483-11e8-8895-9619143e4b26.png)

Image portrait full screen
![image preview 4](https://user-images.githubusercontent.com/1815050/41652571-c86ebbf8-7483-11e8-8b7f-c276db5005ef.png)

Image landscape full screen
![image preview 5](https://user-images.githubusercontent.com/1815050/41652578-cf39d706-7483-11e8-81ef-e1a40caece3e.png)

Figma design:
https://www.figma.com/file/Ui4YzD3oQcxJSbzBEr80O9eU/Working-Draft-UI?node-id=1746%3A228

jesmrec:
the easier, the better. Works for me.

michaelstingl:
Cool idea from @atwm : 

[...]

PR #114, opened 2018-09-24
Feature/file previewing

pablocarmu:
## Description

Make the app able to automatically select the view controller to present based on the mime-type of the item to open.

## Related Issue

https://github.com/owncloud/ios-app/issues/112

## Motivation and Context

This PR takes advantage of the Extension mechanism in the SDK so we can change between implementations easily.

## Types of changes

- [] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

______

BUGS & IMPROVEMENTS

- [x] (1) No connection https://github.com/owncloud/ios-app/pull/114#issuecomment-428902523
- [x] (2) Formats https://github.com/owncloud/ios-app/pull/114#issuecomment-428915636
- [x] (3) GIFs


ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -0,0 +1,89 @@
...
+	override func loadView() {
+		super.loadView()
+		iconImageView = UIImageView()
+		iconImageView.translatesAutoresizingMaskIntoConstraints = false
+		iconImageView.contentMode = .scaleAspectFit
+		view.addSubview(iconImageView)
+		NSLayoutConstraint.activate([
+			iconImageView.centerXAnchor.constraint(equalTo: view.safeAreaLayoutGuide.centerXAnchor),
+			iconImageView.centerYAnchor.constraint(equalTo: view.safeAreaLayoutGuide.centerYAnchor),
+			iconImageView.heightAnchor.constraint(equalToConstant: 200),
```
mneuwert: @pablocarmu avoid using magic numbers

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -0,0 +1,89 @@
...
+			iconImageView.widthAnchor.constraint(equalTo: iconImageView.heightAnchor)
+			])
+	}
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+		iconImageView.image = item.icon(fitInSize:CGSize(width: 200.0, height: 200.0))
```
mneuwert: @pablocarmu avoid using magic numbers

ownCloud/Viewer/DisplayViewProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+import UIKit
+import ownCloudSDK
+
+protocol DisplayViewProtocol where Self: DisplayViewController {
+
+	static var features: [String : Any]? {get}
+	static var supportedMimeTypes: [String] {get}
+}
+
+struct FeatureKeys {
```
mneuwert: @pablocarmu why not use enum?

ownCloud/Viewer/OCExtension Subclasses/OCDisplayExtension.swift
```diff
@@ -0,0 +1,95 @@
...
+		self.locations = [location]
+		self.features = features
+		self.objectProvider = objectProvider
+	}
+
+	override func matchesContext(_ context: OCExtensionContext!) -> OCExtensionPriority {
+
+		var matchPriority: OCExtensionPriority = .noMatch
+
+		if context.location.type == self.type {
```
mneuwert: @pablocarmu For me the below code is not self-explaining without comments

pablocarmu: it's a Swift translation of [this](https://github.com/owncloud/ios-sdk/blob/feature/filemgmt/ownCloudSDK/Extensions/OCExtension.m).

It does the same but also in the location comparison it compares the mime-type of the file you want to present and if it matches, the context match and it's returned as a possible matched value.

it goes through the location, the features and the priority.

pablocarmu:
This branch had previously inside all the commits from the `feature/fileprovider` branch so comparing it agains master created a lot of diffs that aren't made in this branch. I've cleaned all the "non-wanted" commits and now it's only my changes.

As a side effect on this clean, every time the app try to download an element to preview it the app crashes because there is no file provider inside. This could seem like there is a problem in the branch but it isn't.

@mneuwert  I reviewed all the changes you proposed and made all but the 2 I commented. Please check again the code and feel free to propose more changes.

ownCloud/Viewer/OCExtension Subclasses/OCDisplayExtension.swift
```diff
@@ -0,0 +1,95 @@
...
+//  Created by Pablo Carrascal on 03/09/2018.
+//  Copyright © 2018 ownCloud GmbH. All rights reserved.
+//
+
+import UIKit
+import ownCloudSDK
+
+class OCDisplayExtension: OCExtension {
+
+	init(identifier: OCExtensionIdentifier!, type: OCExtensionType!, location: OCExtensionLocation!, features: [String : Any]!, objectProvider: @escaping OCExtensionObjectProvider) {
```
felix-schwarz: I was about to say that there's `+[OCExtension extensionWithIdentifier:type:location:features:objectProvider:]` already, but then noticed, that doesn't seem to get translated to Swift properly. 😬

I just added a pair of init-methods to the SDK (`feature/filemgmt`) now, so that this can be removed/replaced by them.

ownCloud/Viewer/OCExtension Subclasses/OCDisplayExtension.swift
```diff
@@ -0,0 +1,95 @@
...
+			if priority != .noMatch {
+				matchPriority = priority
+			} else {
+				if context.preferences != nil, context.preferences.count > 0, self.features != nil {
+					for preferencekey in context.preferences {
+						let feature = features[preferencekey.key] as AnyObject
+						let requirement = context.preferences[preferencekey.key] as AnyObject
+						if feature.isEqual(requirement) {
+							// TODO : Talk with felix because this is not allowed in Swift.
+							//matchPriority += Float(OCExtensionPriority.featureMatchPlus.rawValue)
```
felix-schwarz: You can achieve this by working solely with raw values, like f.ex.:

```swift
matchPriority = OCExtensionPriority.init(rawValue: matchPriority.rawValue + OCExtensionPriority.featureMatchPlus.rawValue)!
```

ownCloud/Viewer/OCExtension Subclasses/OCDisplayExtension.swift
felix-schwarz: If you put the MIME-Type into the locationIdentifier, there's no longer any need to override `matchesContext`.

As far as I can see, the locationIdentifier is not really in use for anything right now, so that's what I'd propose.

ownCloud/Viewer/DisplayViewProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+	static let canEdit: String = "featureKeyCanEdit"
+	static let canSave: String = "featureKeyCanSave"
+	static let showPDF: String = "featureKeyShowPDF"
```
felix-schwarz: As far as I can see, we don't need `showPDF` as a feature. Instead, PDF viewer(s) can be found by querying `OCExtensionManager` with the PDF MIME-Type `application/pdf`.

ownCloud/Viewer/OCExtension Subclasses/OCDisplayExtension+DisplayView.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Pablo Carrascal on 30/08/2018.
+//  Copyright © 2018 ownCloud GmbH. All rights reserved.
+//
+
+import ownCloudSDK
+
+extension OCDisplayExtension {
+
+	static func normalPDFExtension(identifier: String) -> OCDisplayExtension {
```
felix-schwarz: This static/class method could be moved to the `PDFViewerViewController` class method/property `extension` to make it more self-contained.

ownCloud/Viewer/OCExtension Subclasses/OCDisplayExtension+DisplayView.swift
```diff
@@ -0,0 +1,42 @@
...
+		let features: [String : Any] = PDFViewerViewController.features!
+
+		let normalPDFExtension = OCDisplayExtension(identifier: rawIdentifier, type: .viewer, location: location, features: features) { (_ rawExtension, _ context, _ error) -> Any? in
+			return PDFViewerViewController()
+		}
+
+		return normalPDFExtension
+	}
+
+	static func webViewExtension(identifier: String) -> OCDisplayExtension {
```
felix-schwarz: This static/class method could be moved to the `WebViewDisplayViewController` class method/property `extension` to make it more self-contained.

ownCloud/Viewer/WebViewDisplayViewController.swift
```diff
@@ -0,0 +1,127 @@
...
+
+class WebViewDisplayViewController: DisplayViewController, DisplayViewProtocol {
+	static var supportedMimeTypes: [String] =
+		["image/jpeg",
+		 "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+		 "application/vnd.openxmlformats-officedocument.wordprocessingml.document"]
+
+	static var features: [String : Any]? = [FeatureKeys.canEdit : true]
+
+	static func webViewExtension(identifier: String) -> OCExtension {
```
felix-schwarz: Seeing how this method and `normalPDFExtension` are almost identical, I can see how this could be unified and simplified, f.ex. with an extension to `DisplayViewProtocol` like this:

```swift
extension DisplayViewProtocol {
	static var displayExtension: OCExtension {
            // Code generating OCExtension from `supportedMIMETypes`, `features` and `displayExtensionIdentifier` goes here.
	}
}
```

Then, adding the extension to the extension manager can be done with one line

```swift
OCExtensionManager.shared.addExtension(PDFViewerViewController.displayExtension)
```

and everything related to the respective viewer (incl. ID) is isolated in its own class.

ownCloud/Viewer/WebViewDisplayViewController.swift
```diff
@@ -0,0 +1,127 @@
...
+		let rawIdentifier: OCExtensionIdentifier =  OCExtensionIdentifier(rawValue: identifier)
+		var locationIdentifiers: [OCExtensionLocationIdentifier] = []
+		for mimeType in supportedMimeTypes {
+			let locationIdentifier: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier(rawValue: mimeType)
+			locationIdentifiers.append(locationIdentifier)
+		}
```
felix-schwarz: Given that OCExtensionLocationIdentifiers are also just Strings… do you know of a more direct way to cast this? I couldn't immediately find any.

pablocarmu: I've been looking for some other option and I found some info about how `NS_TYPED_EXTENSIBLE_ENUM` is imported into Swift: [Declare Typed Extensible Enumerations](https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/grouping_related_objective-c_constants)

The solution they propose is to create an extension with a static computed variable for each new case you want to add. Something like:

```Swift
extension OCExtensionLocationIdentifier {
    static var pdf: OCExtensionLocationIdentifier {
        return OCExtensionLocationIdentifier(rawValue: "application/pdf") 
    }
}
```

I think this is not extensible for our use case because we could have a lot of mime types.

I coded the `OCExtension` creation in a `DisplayExtension` extension that provides the default implementation, so this code is only written once. It could be overridden in the future by any class that implements this protocol.

```Swift
static var displayExtension: OCExtension {
		let rawIdentifier: OCExtensionIdentifier =  OCExtensionIdentifier(rawValue: displayExtensionIdentifier)
		var locationIdentifiers: [OCExtensionLocationIdentifier] = []
		for mimeType in supportedMimeTypes {
			let locationIdentifier: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier(rawValue: mimeType)
			locationIdentifiers.append(locationIdentifier)
		}

		let displayExtension = OCExtension(identifier: rawIdentifier, type: .viewer, locations: locationIdentifiers, features: features) { (_ rawExtension, _ context, _ error) -> Any? in
			return Self()
		}

		return displayExtension!
	}
}
```

ownCloud/Viewer/PDFViewerViewController.swift
```diff
@@ -0,0 +1,56 @@
...
+
+import UIKit
+import ownCloudSDK
+import PDFKit
+
+class PDFViewerViewController: DisplayViewController, DisplayExtension {
+
+	static var displayExtensionIdentifier: String = "normalPDFViewer"
+	static var supportedMimeTypes: [String] = ["application/pdf"]
+	static var features: [String : Any]? = [FeatureKeys.canEdit : true]
```
felix-schwarz: This should be `false`, since it can't perform any editing.

ownCloud/Viewer/WebViewDisplayViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+
+class WebViewDisplayViewController: DisplayViewController, DisplayExtension {
+	
+	static var displayExtensionIdentifier: String = "imageViewer"
+	static var supportedMimeTypes: [String] =
+		["image/jpeg",
+		 "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+		 "application/vnd.openxmlformats-officedocument.wordprocessingml.document"]
+
+	static var features: [String : Any]? = [FeatureKeys.canEdit : true]
```
felix-schwarz: This should be `false`, since it can't perform any editing.

ownCloud/Viewer/PDFViewerViewController.swift
```diff
@@ -0,0 +1,56 @@
...
+//  Copyright © 2018 ownCloud GmbH. All rights reserved.
+//
+
+import UIKit
+import ownCloudSDK
+import PDFKit
+
+class PDFViewerViewController: DisplayViewController, DisplayExtension {
+
+	static var displayExtensionIdentifier: String = "normalPDFViewer"
```
felix-schwarz: Let's use reverse-domain-name-style identifiers here, like f.ex. `org.owncloud.pdfViewer.default` (Michael's PDF viewer could then be added as `org.owncloud.pdfViewer.extended`, and something else as `com.othercorp.pdfViewer`).

ownCloud/Viewer/WebViewDisplayViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Copyright © 2018 ownCloud GmbH. All rights reserved.
+//
+
+import UIKit
+import ownCloudSDK
+import WebKit
+
+class WebViewDisplayViewController: DisplayViewController, DisplayExtension {
+	
+	static var displayExtensionIdentifier: String = "imageViewer"
```
felix-schwarz: Analogous to `PDFViewerViewController`, a reverse-domain-name-style identifier here could be `org.owncloud.webview`.

ownCloud/Viewer/WebViewDisplayViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+	static var supportedMimeTypes: [String] =
+		["image/jpeg",
+		 "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+		 "application/vnd.openxmlformats-officedocument.wordprocessingml.document"]
```
felix-schwarz: ~~Please also add MIME Types for other supported formats. As far as I can tell, https://github.com/owncloud/core/blob/master/core/js/mimetypelist.js should be an exhaustive list of what to expect from the OC server.~~

Thinking about it, this isn't as future-proof as it could be.

So in the latest commit I added support for a `customMatcher` to OCExtension. That block gets the `OCExtensionContext` and the `OCExtensionPriority` computed by the default implementation. To add wild-card matching for f.ex. `image/*`, `video/*`, `audio/*` and `text/*`, all one would need to do is add a `customMatcher` block to the `OCExtension` instance representing WebViewDisplayViewController.

In that block, you'd need to check if the provided `OCExtensionPriority` is `OCExtensionPriorityNoMatch`. And if it is, check if the context's location.locationIdentifier begins with `image/`, `video/`, `audio/` or `text/`. If it is, the block could return `OCExtensionPriorityLocationMatch`, otherwise return `OCExtensionPriorityNoMatch`.

That block could be provided to `DisplayExtension.displayExtension` by adding another property to return a `customMatcher`.

That said, adding those MIME Types from https://github.com/owncloud/core/blob/master/core/js/mimetypelist.js that do not match one of the "wildcards" should still be done, especially for Office formats like PowerPoint.

felix-schwarz:
I've made a few minor changes that I thought were quicker to write in code than in text:
- Removed ClientItemViewController.swift
- Added progress bar and cancel button to DisplayViewController
- Added theming support to DisplayViewController (for button, progress bar)
- Extended NSObject+ThemeApplication with support for theming UIProgressViews
- Extended WebCViewDisplayViewController regex to also match audio and video MIME types
- Added missing copyright texts and other minor cleanups

Using it for a bit, I also noticed a few other things that still need attention:
- the name of the viewed file isn't shown anywhere right now (the navigation bar might be a good fit).
- the PDF viewer starts at a scroll position that's not the top of the PDF (probably nothing that needs fixing as the PDF viewer will be replaced by @mneuwert's PDF viewer eventually)
- what happens when trying to view a filetype that isn't supported by a viewer extension?
- error handling

pablocarmu:
>what happens when trying to view a file type that isn't supported by a viewer extension?

Right now the item it's downloaded and the thumbnail or icon (if any) is presented. Like when you are trying to open a supported file but without opening the file.

There are some possibilities here:

**A**.  Show the thumbnail with a "non-supported file" message below the thumbnail, without download the file.
**B**.  Show the thumbnail with a "non-supported file" message below the thumbnail and download the file.
**C**.  Not even show the thumbnail and warn the user with an alert that this file is not supported by the app.

IMO the option **A** should be the one to go, what do you guys think? @felix-schwarz @mneuwert @jesmrec @javiergonzper @michaelstingl

mneuwert:
@pablocarmu I would vote against **C** at least. I think that the decision if the file has to be downloaded shall depend on what actions can be performed with the downloaded file (share, copy into Files app, print or whatever). Besides, even if we can’t show file preview we could still try to show some interesting information along with the thumbnail icon: e.g. file size, meta-data etc since we have more room for that in the detail view.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

felix-schwarz:
I'd also go with A. If a file is tapped for preview and can't be shown, there's no benefit in downloading it.

Regarding further actions to act on a local copy of the file: I think that once the user triggers such an action would be the right time to trigger the download.

Conveniently, you can always just call `OCCore`'s download method and avoid two separate code paths: if a copy of the file already exists locally, `OCCore` will promptly call the `resultHandler`. And if the file doesn't have a local copy, it'll download the file from the server and then call the `resultHandler` when done.

@pablocarmu @mneuwert @jesmrec @javiergonzper @michaelstingl

jesmrec:
Lest's QA a bit

### (1)

What do we expect in the following scenario?

1. Switch connection off
2. Tap on a file

Current: Download view static, no notification to the user
Expected: Alert the user when he taps on "Open file" (other ideas are welcome)

jesmrec:
### (2)

Formats that are not opened:

- txt/readable formats: xml, tex, odt
- media formats: flv, wmv, avi, ogv

jesmrec:
Pics are shown alligned to the top on the screen. We should improve such behaviour when we develop https://github.com/owncloud/ios-app/issues/85

ownCloud/Viewer/ImageDisplayViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+		scrollView?.bounces = false
+		scrollView?.bouncesZoom = false
+		scrollView?.delegate = self
+		self.view.addSubview(scrollView!)
+
+		NSLayoutConstraint.activate([
+			scrollView!.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor),
+			scrollView!.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor),
+			scrollView!.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+			scrollView!.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor)
```
felix-schwarz: By using the `safeAreaLayoutGuide`, the scroll view does only fill part of the available space and is "cut off" on the bottom on iPhone X:
bildschirmfoto 2018-10-17 um 23 13 40

ownCloud/Viewer/ImageDisplayViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+			imageView?.translatesAutoresizingMaskIntoConstraints = false
+			imageView?.backgroundColor = .black
+			scrollView!.addSubview(imageView!)
+			NSLayoutConstraint.activate([
+				imageView!.centerXAnchor.constraint(equalTo: scrollView!.centerXAnchor),
+				imageView!.centerYAnchor.constraint(equalTo: scrollView!.centerYAnchor),
+				imageView!.leftAnchor.constraint(equalTo: scrollView!.leftAnchor),
+				imageView!.rightAnchor.constraint(equalTo: scrollView!.rightAnchor),
+				imageView!.bottomAnchor.constraint(equalTo: scrollView!.bottomAnchor),
+				imageView!.topAnchor.constraint(equalTo: scrollView!.topAnchor)
```
felix-schwarz: Using constraints to center the imageView inside the scrollview works initially, but the zoom will also scale up the blank space around the image later. Scrolling around while zoomed into the image, then, can give you this result:
bildschirmfoto 2018-10-17 um 23 20 18

You can use `UIScrollView.contentInsets` and adapt these in `UIScrollViewDelegate`'s `scrollViewDidZoom(_ scrollView: UIScrollView)` as zoom levels change to increase or decrease insets value. [This Stackoverflow question](https://stackoverflow.com/questions/14069571/keep-zoomable-image-in-center-of-uiscrollview) gives a few examples on how that could be done.

ownCloud/Viewer/ImageDisplayViewController.swift
```diff
@@ -0,0 +1,87 @@
+//
+//  ImageDisplayViewController.swift
+//  ownCloud
+//
+//  Created by Pablo Carrascal on 17/10/2018.
+//  Copyright © 2018 ownCloud GmbH. All rights reserved.
+//
+
```
felix-schwarz: Please add the standard copyright notice.

jesmrec:
> * media formats: flv, wmv, avi, ogv
These formats are not supported by the server based on what I can see here 

ok, the point is that such formats (but .avi) try to be opened in the app, pointing to a white screen: 

oct-18-2018 12-40-52

same behaviour in case of .odt files

xml, tex, avi -> correct behaviour for non-supported formats

jesmrec:
On the other side, GIF files are not reproduced anymore with the latest commits. This is not a priority and can be faced in a new ticket.

jesmrec:
Issues detected:

- Media formats lead to blank screen: https://github.com/owncloud/ios-app/pull/114#issuecomment-430962069
- .odt files lead to blank screen
- GIF files not reproduced
- SVG files: sometimes crashes, other times, tries to be downloaded but unsuccesfully. Crash happens with `Thread 1: Fatal error: Unexpectedly found nil while unwrapping an Optional value` in https://github.com/owncloud/ios-app/blob/666dc8cf73fd6d36ce4335f72dc61a0d51eb353b/ownCloud/Viewer/ImageDisplayViewController.swift#L31

jesmrec:
From my side this is OK to move forward. Finally, some unsupported formats were displayed in web view (to improve incrementally), and SVG crash was fixed.

Issue #62, opened 2018-06-05
[File list] Display info about items


jesmrec:
Within the sorting options in #55 , we are including the possibility of sorting by date and size, but such info is not being displayed. 

in a first design iteration, the info about size and date is displayed just below the name:

![screen shot 2018-06-05 at 12 18 08](https://user-images.githubusercontent.com/14894746/40970361-aa364392-68ba-11e8-8364-1faf28aa0bce.png)

right now it is displayed the MIME type. Let's check for other approaches

jesmrec:
Done

PR #117, opened 2018-09-25
item data (size, last modified) in the file's list cell detail label

pablocarmu:
## Description

Instead of the mime-type this PR write in the detail label the data in a readable way i.e. `54 Mb` and the last modified date i.e. `06 sept, 2018`
## Related Issue

https://github.com/owncloud/ios-app/issues/62

## Motivation and Context

This provides more information to the user and makes sense because we can sort by date and size

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/SDK Extensions/OCItem+Extension.swift
+++ b/ownCloud/SDK Extensions/OCItem+Extension.swift
@@ -216,4 +216,21 @@ extension OCItem {
 	func fileSuffix() -> String {
 		return (self.name as NSString).pathExtension
 	}
+
+	var sizeInReadableFormat: String {
+		let byteCountFormatter = ByteCountFormatter()
+		byteCountFormatter.countStyle = .file
+		let size = byteCountFormatter.string(fromByteCount: Int64(self.size))
+		return size
+	}
+
+	var lastModifiedInReadableFormat: String {
+		let dateFormatter = DateFormatter()
+		dateFormatter.timeStyle = .none
+		dateFormatter.dateStyle = .medium
+		dateFormatter.locale = Locale.current
+		dateFormatter.doesRelativeDateFormatting = true
+		let dateString = dateFormatter.string(from: self.lastModified)
+		return dateString
+	}
 }
--- a/ownCloud/Client/ClientItemCell.swift
+++ b/ownCloud/Client/ClientItemCell.swift
@@ -122,11 +122,7 @@ class ClientItemCell: ThemeTableViewCell {
 
 		iconImage = item.icon(fitInSize: iconSize)
 
-		if item.type == .collection {
-			self.detailLabel.text = "Folder".localized
-		} else {
-			self.detailLabel.text = item.mimeType
-		}
+		self.detailLabel.text = item.sizeInReadableFormat + " - " + item.lastModifiedInReadableFormat
 
 		self.accessoryType = .none
 


jesmrec:
Regarding https://github.com/owncloud/ios-app/issues/62, in case of folders is included the number of items instead of the folder size. I personally prefer the folder size. 

What do you think? @javiergonzper @mneuwert @michaelstingl @felix-schwarz

michaelstingl:
@pablocarmu @jesmrec screenshots?

ownCloud/SDK Extensions/OCItem+Extension.swift
```diff
@@ -216,4 +216,21 @@ extension OCItem {
...
 	}
+
+	var sizeInReadableFormat: String {
+		let byteCountFormatter = ByteCountFormatter()
+		byteCountFormatter.countStyle = .file
+		let size = byteCountFormatter.string(fromByteCount: Int64(self.size))
+		return size
+	}
+
+	var lastModifiedInReadableFormat: String {
```
mneuwert: @pablocarmu Creating formatter instances on the fly is very expensive. I would propose having those lazily created and cached statically, so that they can be re-used.

jesmrec:
screen shot 2018-09-26 at 15 39 28

michaelstingl:
Display both?

Finder on macOS;
> 107.469.460.299 bytes (746,3 MB on disk) for 1.680 items

That's too much detail, but only size and number of items?

javiergonzper:
> Regarding #62, in case of folders is included the number of items instead of the folder size. I personally prefer the folder size.
> 
> What do you think? @javiergonzper @mneuwert @michaelstingl @felix-schwarz

On Files App they show the size only on the files. Not on the folders. But probably it is because that information could no be available (calculate the size of a folder with subfolders is not an easy task and maybe not all the cloud services return it).

To be consisten I also prefer the size of the folder but not sure what it is more useful for a normal user.

javiergonzper:
> Display both?
> 
> Finder on macOS;
> 
> > 107.469.460.299 bytes (746,3 MB on disk) for 1.680 items
> 
> That's too much detail, but only size and number of items?

Size, number of items and last modification date. Too much in my opinion.

jesmrec:
@mneuwert we take the size from API

ownCloud/SDK Extensions/OCItem+Extension.swift
pablocarmu: yeah, make sense. I'll change it and push it!

michaelstingl:
ownCloud server might not know the size from folders that are mounted storages. That needs to be covered as well…

pablocarmu:
From the app's perspective, I'm showing in the detail the `size` and `lastModifiedDate` properties of `OCItem` with a nice format.

> Regarding #62, in case of folders is included the number of items instead of the folder size. I personally prefer the folder size.

There is no current property in `OCItem` to know the number of child items of a folder and I think that unless you make a propfind to get the contents of this folder and count the results you can not get it.

> ownCloud server might not know the size from folders that are mounted storages. That needs to be covered as well…

Right now if there is no info about the size we're showing zero as displayed [here](https://github.com/owncloud/ios-app/pull/117#issuecomment-424718937). I wanted to mimic the behavior of the old app here, but instead of showing minute details I coded the date by days ie. _Today, Tomorrow, 3 days ago,..._

jesmrec:
i will check what happens after mounting an external storage... but the app will show what the server retrieves.

pablocarmu:
I changed the property `allowsNonnumericFormatting` to `false` to not allow non-numeric display options. Now it will show `0 bytes`

jesmrec:
Regarding external storage, we can not know the  folder and file sizes till we browse through it. PROPFIND returns:

`<oc:size>-1</oc:size>`

with such value, the web UI sets `Pending` as folder size, and set a different icon. We can distinguish such kind of folders with the permissions property, including `M`.

Icon is not in the scope of the current PR.

jesmrec:
External storage issue fixed. From my side, this is OK to move forward... after code review (@mneuwert )

Issue #23, opened 2018-04-09
Files: Duplicate item


jesmrec:
AC:

- [x] Set the option to duplicate in the file list
- [x] The item will be duplicated in the same  folder than the original
- [x] The name of the duplicated will be the same, adding "copy" to the file name (before the extension). F. ex. :  ownCloud.pdf -> ownCloud copy.pdf
- [x] In case of a folder, all the content inside will be duplicated to keep the same structure
- [x] Handle issues (for example, no connection etc..)
- [x] Refresh the list view to show the new status
- [x] QA

PR #119, opened 2018-09-26
Duplicate files and folders

pablocarmu:
## Description

With this pull request the app is now able to duplicate files and folders.
When you duplicate a folder the naming follows the next schema:
> item-name + _copy_

## Related Issue

https://github.com/owncloud/ios-app/issues/23

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/Actions/NamingViewController.swift
+++ b/ownCloud/Client/Actions/NamingViewController.swift
@@ -349,7 +349,7 @@ extension NamingViewController: UITextFieldDelegate {
 
 		if let name = nameTextField.text,
 			let fileExtension = item?.fileExtension(),
-			let range = name.range(of: fileExtension),
+			let range = name.range(of: ".\(fileExtension)"),
 			let position: UITextPosition = nameTextField.position(from: nameTextField.beginningOfDocument, offset: range.lowerBound.encodedOffset) {
 
 				textField.selectedTextRange = nameTextField.textRange(from: nameTextField.beginningOfDocument, to:position)
--- a/ownCloud/Client/Actions/NamingViewController.swift
+++ b/ownCloud/Client/Actions/NamingViewController.swift
@@ -348,9 +348,9 @@ extension NamingViewController: UITextFieldDelegate {
 	func textFieldDidBeginEditing(_ textField: UITextField) {
 
 		if let name = nameTextField.text,
-			let fileExtension = item?.fileSuffix(),
+			let fileExtension = item?.fileExtension(),
 			let range = name.range(of: fileExtension),
-			let position: UITextPosition = nameTextField.position(from: nameTextField.beginningOfDocument, offset: range.lowerBound.encodedOffset - 1) {
+			let position: UITextPosition = nameTextField.position(from: nameTextField.beginningOfDocument, offset: range.lowerBound.encodedOffset) {
 
 				textField.selectedTextRange = nameTextField.textRange(from: nameTextField.beginningOfDocument, to:position)
 
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -565,10 +565,17 @@ class ClientQueryViewController: UITableViewController, Themeable {
 	}
 
 	func duplicate(_ item: OCItem, viewDidAppearHandler: ClientActionVieDidAppearHandler? = nil, completionHandler: ClientActionCompletionHandler? = nil) {
-		let name = "\(item.fileSuffix()) copy.\(item.path.components(separatedBy: ".").last!)"
-		if let progress = self.core?.copy(item, to: query?.rootItem, withName: name, options: nil, resultHandler: { (error, _, _, _) in
+		var name: String = "\(item.name!) copy"
+
+		if item.type != .collection {
+			let suffix = item.fileSuffix()
+			let itemNameWithouExtension = item.name.dropLast(suffix.count + 1) // The extra one is for the '.' in the extension.
+			name = "\(itemNameWithouExtension) copy.\(suffix)"
+		}
+
+		if let progress = self.core?.copy(item, to: self.query?.rootItem, withName: name, options: nil, resultHandler: { (error, _, item, _) in
 			if error != nil {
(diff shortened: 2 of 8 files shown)


jesmrec:
From my side this is OK. 

@mneuwert or @felix-schwarz please review the code

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -563,6 +563,29 @@ class ClientQueryViewController: UITableViewController, Themeable {
 
 		self.present(createFolderNavigationVC, animated: true, completion: viewDidAppearHandler)
 	}
+
+	func duplicate(_ item: OCItem, viewDidAppearHandler: ClientActionVieDidAppearHandler? = nil, completionHandler: ClientActionCompletionHandler? = nil) {
+		var name: String = "\(item.name!) copy"
+
+		if item.type != .collection {
+			let suffix = item.fileSuffix()
```
mneuwert: would be extension more clear here?

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -563,6 +563,29 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+			let itemNameWithouExtension = item.name.dropLast(suffix.count + 1) // The extra one is for the '.' in the extension.
```
mneuwert: I would rather have a method in OCItem returning the name w/o extension instead of in-place string manipulation for better readability. Also what is with files without extension? Those are pretty common in the Apple ecosystem.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: you're right, i'll change it.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: I'll create a new function in `OCItem+Extension` like:
```Swift
func nameWithoutExtension() -> String {
	return (self.name as NSString).deletingPathExtension
}
```

>Also what is with files without extension? 
by using `NSString` APIs we can use `pathExtension` [NSString pathExtension](https://developer.apple.com/documentation/foundation/nsstring/1407801-pathextension)

if there is no extension it return `""` so should not be any problem with files without extension.

mneuwert: @pablocarmu still you would assume that there is a "." character in dropLast() call, right?

pablocarmu: I'll change the implementation of this part to:
```Swift
if item.type != .collection {
	let itemName = item.nameWithoutExtension()
	var fileExtension = item.fileExtension()

	if fileExtension != "" {
		fileExtension = ".\(fileExtension)"
	}

	name = "\(itemName) copy\(fileExtension)"
}
```
I can do it here or I can change `fileExtension()` implementation to return `""` in case there is no extension or `.extension` if there is. I think the second option is the way to go, do you agree?

ownCloud/SDK Extensions/OCItem+Extension.swift
```diff
@@ -213,7 +213,16 @@ extension OCItem {
 		return nil
 	}
 
-	func fileSuffix() -> String {
-		return (self.name as NSString).pathExtension
+	func fileExtension() -> String {
+		var fileExtension = (self.name as NSString).pathExtension
+		if fileExtension != "" {
+			fileExtension = ".\(fileExtension)"
```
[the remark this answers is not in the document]
pablocarmu: The property `name` in `OCItem` has the extension in it so for example it would return `foobar.pfd`. If you want to make a duplicate file, you need to `copy` the file and give it a new name. Does not work simply by adding "copy" at the end because with no extension the file loses its mime-type.

I would prefer that this method returns the dot because if not, we need to add it later with some code like the following:
```Swift
if item.type != .collection {
	let itemName = item.nameWithoutExtension()
	var fileExtension = item.fileExtension()

	if fileExtension != "" {
		fileExtension = ".\(fileExtension)"
	}

	name = "\(itemName) copy\(fileExtension)"
}
```

ownCloud/Client/Actions/NamingViewController.swift
```diff
@@ -348,9 +348,9 @@ extension NamingViewController: UITextFieldDelegate {
 	func textFieldDidBeginEditing(_ textField: UITextField) {
 
 		if let name = nameTextField.text,
-			let fileExtension = item?.fileSuffix(),
+			let fileExtension = item?.fileExtension(),
 			let range = name.range(of: fileExtension),
-			let position: UITextPosition = nameTextField.position(from: nameTextField.beginningOfDocument, offset: range.lowerBound.encodedOffset - 1) {
+			let position: UITextPosition = nameTextField.position(from: nameTextField.beginningOfDocument, offset: range.lowerBound.encodedOffset) {
```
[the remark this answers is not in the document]
pablocarmu: I'll check it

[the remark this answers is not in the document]
pablocarmu: after some testing, in the following case:
```Swift
var fileExtension: String = ""
let range = name.range(of: fileExtension)
```

`range` value is `nil` and as we use `if let` it's not goint to go through this line but it's going to select:

```Swift
textField.selectedTextRange = nameTextField.textRange(from: nameTextField.beginningOfDocument, to: nameTextField.endOfDocument)
```

so there is no going to be any problem if the `fileExtension` function returns an empty String.

ownCloud/SDK Extensions/OCItem+Extension.swift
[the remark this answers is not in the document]
mneuwert: @pablocarmu I understood that but I proposed not to add the dot for the sake of the clean API since dot is not actually a part of the file extension it is just a separator. Ideally, you should be able to understand what the API function is returning without looking up the documentation and I personally would not expect fileExtension() to return extension plus something else.

ownCloud/Client/Actions/NamingViewController.swift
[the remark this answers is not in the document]
mneuwert: @pablocarmu right, I confused it with NSString.range(..) implementation which doesn't return nil.

PR #124, opened 2018-10-09
Fix TVG rendering

felix-schwarz:
- Extend TVG and MakeTVG with new "viewBox" attribute
- Extend TVGImage to support "viewBox" attribute to correctly render whitespace on the right and bottom of vector images (was previously not honored)

bildschirmfoto 2018-10-09 um 13 06 40


pablocarmu (approved):

felix-schwarz:
That's actually unrelated to SVG/TVG rendering.

The thinking between showing icons and thumbnails that led to this choice is this:
- icons that are as large as the thumbnails feel huge and (to me, at least) don't feel right to the point that it looks like a bug
- thumbnails as small as the icons feel too small and offer too little space to be of much use

That's why, right now, the app displays icons at a smaller size than thumbnails. I'm of course open to change and discussing this in a separate issue.

jesmrec:
i did not think in a bug for such behaviour, for me it is ok to discuss in a new issue. 

This one is approved on my side.

jesmrec (approved):

PR #125, opened 2018-10-16
[fix/errorPresentation] Ensure issues and alerts are always presented

felix-schwarz:
## Description
Ensure that issues and alerts are presented even if the `ClientRootViewController` already presents a view controller

## Motivation and Context
Previously, when f.ex. trying to create a folder in the move interface with the same name as an existent one, the resulting error wouldn't be presented, but the Sync Journal get stuck instead waiting for a user interaction that can't happen.

## Types of changes


felix-schwarz:
Thanks for the feedback! That the alert is shown twice is a bug in the SDK that went previously unnoticed because it was dropped by UIKit. I already fixed it in `feature/offline` (https://github.com/owncloud/ios-sdk/commit/2f0b908b719ecc4dd82a6b25e01f598dbfe579c4) a couple of days ago but haven't yet created a separate PR for this issue in the SDK repo.

jesmrec:
if it is fixed, go ahead.

felix-schwarz:
> if it is fixed, go ahead.

Just merged back the latest changes from `master`. You can now merge the PR.

PR #131, opened 2018-10-24
Feature/previewer refinements

pablocarmu:
## Description

This PR fixes some of the pending bugs in the image previewer.
It adds some cool gesture recognizers and the ability to hide the navigation bar and the home indicator (in X layout phones).
The scrollView that contains the image now is able to bounce to have a nice effect.

## Motivation and Context

This makes our image previewer a more solid one. It has more usability and is more user-friendly.

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Viewer/DisplayHostViewController.swift
+++ b/ownCloud/Viewer/DisplayHostViewController.swift
@@ -87,10 +87,6 @@ class DisplayHostViewController: UIViewController {
 		navigationController?.tabBarController?.tabBar.isHidden = false
 	}
 
-	override func prefersHomeIndicatorAutoHidden() -> Bool {
-		return true
-	}
-
 	override func childViewControllerForHomeIndicatorAutoHidden() -> UIViewController? {
 		if let childViewController = self.childViewControllers.first {
 			return childViewController
--- a/ownCloud/Viewer/DisplayHostViewController.swift
+++ b/ownCloud/Viewer/DisplayHostViewController.swift
@@ -87,6 +87,17 @@ class DisplayHostViewController: UIViewController {
 		navigationController?.tabBarController?.tabBar.isHidden = false
 	}
 
+	override func prefersHomeIndicatorAutoHidden() -> Bool {
+		return true
+	}
+
+	override func childViewControllerForHomeIndicatorAutoHidden() -> UIViewController? {
+		if let childViewController = self.childViewControllers.first {
+			return childViewController
+		}
+		return nil
+	}
+
 	// MARK: - Host Actions
 	private func selectDisplayViewControllerBasedOn(mimeType: String) -> (DisplayViewController) {
 
--- a/ownCloud/Viewer/ImageDisplayViewController.swift
+++ b/ownCloud/Viewer/ImageDisplayViewController.swift
@@ -11,9 +11,15 @@ import ownCloudSDK
 
 class ImageDisplayViewController : DisplayViewController {
 
+	// MARK: - Instance variables
 	var scrollView: ImageScrollView?
 	var imageView: UIImageView?
 
+	// MARK: - Gesture recognizers
+	var tapToZoomGestureRecognizer : UITapGestureRecognizer!
+	var tapToHideBarsGestureRecognizer: UITapGestureRecognizer!
+
+	// MARK: - Specific view
 	override func renderSpecificView() {
 		scrollView = ImageScrollView(frame: self.view.bounds)
 		scrollView?.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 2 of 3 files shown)


ownCloud/Viewer/DisplayHostViewController.swift
```diff
@@ -87,6 +87,17 @@ class DisplayHostViewController: UIViewController {
 		navigationController?.tabBarController?.tabBar.isHidden = false
 	}
 
+	override func prefersHomeIndicatorAutoHidden() -> Bool {
+		return true
+	}
+
+	override func childViewControllerForHomeIndicatorAutoHidden() -> UIViewController? {
```
mneuwert: @pablocarmu I don’t understand how this method is related to “home indicator autohidden”? At least not from reading method signature..

ownCloud/Viewer/ImageDisplayViewController.swift
```diff
@@ -38,12 +54,59 @@ class ImageDisplayViewController : DisplayViewController {
...
+	// MARK: - Gesture recognizers actions.
+	@objc func tapToZoom() {
+		guard let scrollView = scrollView else {
+			return
+		}
+
+		if scrollView.zoomScale != scrollView.minimumZoomScale {
+			scrollView.setZoomScale(scrollView.minimumZoomScale, animated: true)
+		} else {
+			scrollView.setZoomScale(scrollView.maximumZoomScale / 3, animated: true)
```
mneuwert: @pablocarmu I would avoid using magic numbers in the code

ownCloud/Viewer/ImageDisplayViewController.swift
```diff
@@ -38,12 +54,59 @@ class ImageDisplayViewController : DisplayViewController {
...
 
+	// MARK: - Frame changes
 	override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
 		super.viewWillTransition(to: size, with: coordinator)
 		self.scrollView?.updateScaleForRotation(size: size)
 	}
+
+	// MARK: - Gesture recognizers actions.
+	@objc func tapToZoom() {
+		guard let scrollView = scrollView else {
```
mneuwert: @pablocarmu I think you can assume that scrollView is always accessible and just force unwrap it

ownCloud/Viewer/ImageDisplayViewController.swift
```diff
@@ -38,12 +54,59 @@ class ImageDisplayViewController : DisplayViewController {
...
+			return
+		}
+
+		if scrollView.zoomScale != scrollView.minimumZoomScale {
```
mneuwert: @pablocarmu direct comparison of float numbers is evil due to rounding errors

ownCloud/Viewer/ImageDisplayViewController.swift
```diff
@@ -38,12 +54,59 @@ class ImageDisplayViewController : DisplayViewController {
...
+			scrollView.setZoomScale(scrollView.minimumZoomScale, animated: true)
+		} else {
+			scrollView.setZoomScale(scrollView.maximumZoomScale / 3, animated: true)
+		}
+
+		setNeedsUpdateOfHomeIndicatorAutoHidden()
```
mneuwert: @pablocarmu method name is not self-explanatory for me

ownCloud/Viewer/DisplayHostViewController.swift
pablocarmu: The home indicator is always managed by the system, we can not hide it if we want.
But you can say to iOS that you prefer for your view controller the home indicator to be hidden.

This calls `prefersHomeIndicatorAutoHidden`. If you return true, basically you're saying to the system that you want it hidden, but is an iOS choice.

If you have child view controllers and you want a different behavior for each one, you need to implement  `childViewControllerForHomeIndicatorAutoHidden`. If you return a view controller that is your child, the system will ask for the child view controller `prefersHomeIndicatorAutoHidden` value instead of the parent one. If you return nil the system asks for the parent `prefersHomeIndicatorAutoHidden`.

`DisplayHostViewController` is the parent view controller of `ImageDisplayViewController`, and we want to hide the home indicator when we're in full screen in the image preview. So we need to override this func and return `ImageDisplayViewController` so it's called its implementation.

I should delete the `prefersHomeIndicatorAutoHidden` implementation in `DisplayHostViewController`.

ownCloud/Viewer/ImageDisplayViewController.swift
pablocarmu: you're right, I'll code this in a constant.

ownCloud/Viewer/ImageDisplayViewController.swift
pablocarmu: This is from UIKit, it asks for the value of the `prefersHomeIndicatorAutoHidden` and update the home indicator preferences.

[setNeedsUpdateOfHomeIndicatorAutoHidden](https://developer.apple.com/documentation/uikit/uiviewcontroller/2887509-setneedsupdateofhomeindicatoraut#)

ownCloud/Viewer/ImageDisplayViewController.swift
pablocarmu: I'll change the way I'm comparing those numbers, it is a preferred way of doing this kind of comparisons or simply doing a `<` or `>` works?

mneuwert: @pablocarmu I double checked the code, I think here it's fine.

ownCloud/Viewer/ImageDisplayViewController.swift
mneuwert: @pablocarmu Ok, I was not aware that it is Apple's API. Sorry:) My bad.

ownCloud/Viewer/DisplayHostViewController.swift
mneuwert: @pablocarmu Ok, got it. Checked Apple docs, was not aware that this API even exists

Issue #130, opened 2018-10-23
[File Preview] Send a file to other applications


michaelstingl:
I like the way it's done in the iOS Files app viewer:

| File list | Preview | Send … |
|---|---|---|
| ![2018-10-17 23 57 27](https://user-images.githubusercontent.com/214010/47119070-ceef7a00-d269-11e8-825b-59d13567f4bc.png) | ![2018-10-17 23 57 37](https://user-images.githubusercontent.com/214010/47119071-ceef7a00-d269-11e8-9bc6-689b34ea42e5.png) | ![2018-10-17 23 57 59](https://user-images.githubusercontent.com/214010/47119072-ceef7a00-d269-11e8-9a8a-0e8283428a57.png) |

I also like the clean way it looks in the default "Quick Look" feature:
- https://developer.apple.com/design/human-interface-guidelines/ios/system-capabilities/quick-look/
- https://developer.apple.com/documentation/quicklook

I think @felix-schwarz said a simple implementation shouldn't be that much effort? (few lines of code?) Is there something available in the system that could be used here easily?

This is part of the epic https://github.com/owncloud/ios-app/issues/68 (File Preview)

pablocarmu:
I think this _open in action sheet_ should be shown in two different ways:

1. Tapping on the 3 dots in the file list view: by doing this, the _more card view_ should show up and then there should be a new action called _open in_. When you tap in this new action, the file should start the download process and once the download finishes, the _open in action sheet_ appears.

2. We should code a new `UIToolbar` in the file's preview view controller and there a new button with the same flow than tapping in the _open in_ action.

michaelstingl:
Yeah, triggering _more card view_ in both cases sounds good 👍 

What would be the effort to add this to the file preview view controller? Would a lightweight "3-dot-only" solution make sense as a first implementation?

michaelstingl:
@pablocarmu awesome 😍 !!

jesmrec:
Done. 

Check improvements here: 

https://github.com/owncloud/ios-app/issues/178
https://github.com/owncloud/ios-app/issues/179

PR #132, opened 2018-10-24
[feature/open_in] Open in another app (via UIDocumentInteractionController) 

pablocarmu:
## Description

I coded a new action in the card view controller that is launched through the 3 dots in the file list to be able to open in a file in another app.

## Related Issue

https://github.com/owncloud/ios-app/issues/130

## Motivation and Context

With this PR you're able to open one item in another app

## Types of changes

## Checklist:

_____

BUGS & IMPROVEMENTS

- [ ] (1) Double download to open in https://github.com/owncloud/ios-app/pull/132#issuecomment-441633703
- [ ] (2) PDF with no options in preview https://github.com/owncloud/ios-app/pull/132#issuecomment-441635980
- [ ] (3) No connection https://github.com/owncloud/ios-app/pull/132#issuecomment-441984471
- [ ] (4) Bottom status bar not vanished
- [ ] (5) "open in" a removed icon https://github.com/owncloud/ios-app/pull/132#issuecomment-442037655

### DIFF
--- a/ownCloud/AppDelegate.swift
+++ b/ownCloud/AppDelegate.swift
@@ -53,6 +53,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		OCExtensionManager.shared.addExtension(WebViewDisplayViewController.displayExtension)
 		OCExtensionManager.shared.addExtension(PDFViewerViewController.displayExtension)
 		OCExtensionManager.shared.addExtension(ImageDisplayViewController.displayExtension)
+		OCExtensionManager.shared.addExtension(OpenInAction.actionExtension)
+		OCExtensionManager.shared.addExtension(DeleteAction.actionExtension)
+		OCExtensionManager.shared.addExtension(MoveAction.actionExtension)
 
 		Theme.shared.activeCollection = ThemeCollection(with: ThemeStyle.preferredStyle)
 
--- a/ownCloud/AppDelegate.swift
+++ b/ownCloud/AppDelegate.swift
@@ -57,6 +57,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		OCExtensionManager.shared.addExtension(DeleteAction.actionExtension)
 		OCExtensionManager.shared.addExtension(MoveAction.actionExtension)
 		OCExtensionManager.shared.addExtension(RenameAction.actionExtension)
+		OCExtensionManager.shared.addExtension(DuplicateAction.actionExtension)
 
 		Theme.shared.activeCollection = ThemeCollection(with: ThemeStyle.preferredStyle)
 
--- a/ownCloud/AppDelegate.swift
+++ b/ownCloud/AppDelegate.swift
@@ -50,9 +50,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
 		application.setMinimumBackgroundFetchInterval(UIApplicationBackgroundFetchIntervalMinimum + 10)
 
+		// Display Extensions
 		OCExtensionManager.shared.addExtension(WebViewDisplayViewController.displayExtension)
 		OCExtensionManager.shared.addExtension(PDFViewerViewController.displayExtension)
 		OCExtensionManager.shared.addExtension(ImageDisplayViewController.displayExtension)
+
+		// Action Extensions
 		OCExtensionManager.shared.addExtension(OpenInAction.actionExtension)
 		OCExtensionManager.shared.addExtension(DeleteAction.actionExtension)
 		OCExtensionManager.shared.addExtension(MoveAction.actionExtension)
--- a/ownCloud/AppDelegate.swift
+++ b/ownCloud/AppDelegate.swift
@@ -61,6 +61,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		OCExtensionManager.shared.addExtension(MoveAction.actionExtension)
 		OCExtensionManager.shared.addExtension(RenameAction.actionExtension)
 		OCExtensionManager.shared.addExtension(DuplicateAction.actionExtension)
+		OCExtensionManager.shared.addExtension(CreateFolderAction.actionExtension)
 
 		Theme.shared.activeCollection = ThemeCollection(with: ThemeStyle.preferredStyle)
 
--- a/ownCloud/AppDelegate.swift
+++ b/ownCloud/AppDelegate.swift
@@ -56,6 +56,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		OCExtensionManager.shared.addExtension(OpenInAction.actionExtension)
 		OCExtensionManager.shared.addExtension(DeleteAction.actionExtension)
 		OCExtensionManager.shared.addExtension(MoveAction.actionExtension)
+		OCExtensionManager.shared.addExtension(RenameAction.actionExtension)
 
 		Theme.shared.activeCollection = ThemeCollection(with: ThemeStyle.preferredStyle)
 
--- a/ownCloud/Client/Actions/Action.swift
+++ b/ownCloud/Client/Actions/Action.swift
@@ -9,234 +9,211 @@
 import UIKit
 import ownCloudSDK
 
-typealias ActionCompletion = ((_ item: OCItem, _ core: OCCore, _ vcToPresent: UIViewController) -> Void)?
-
-enum ActionType {
+enum ActionCategory {
+	case normal
 	case destructive
-	case regular
+	case informal
+	case edit
+	case save
 }
 
-struct Action {
-	var name: String
-	var type: ActionType
-	var completion: ActionCompletion
+enum ActionPosition : Int {
+	case none = -1
 
-	init(with name: String, completion: ActionCompletion, type: ActionType) {
-		self.name = name
-		self.completion = completion
-		self.type = type
+	case first = 100
+	case beforeMiddle = 200
+	case middle = 300
+	case afterMiddle = 400
+	case last = 500
 
+	static func between(_ position1: ActionPosition, and position2: ActionPosition) -> ActionPosition {
+		return ActionPosition(rawValue: ((position1.rawValue + position2.rawValue)/2))!
+	}
+
+	func shift(by offset: Int) -> ActionPosition {
+		return ActionPosition(rawValue: self.rawValue + offset)!
 	}
 }
 
-class ActionsMoreViewController: NSObject {
+enum Result<T> {
+	case success(T)
+	case failure(Error)
+}
 
-	weak var vcToPresentIn: UIViewController?
-	var moreViewController: UIViewController?
-	var core: OCCore
-	var item: OCItem
+typealias ActionCompletionHandler<T> = ((Result<T>) -> Void)
+typealias ActionProgressHandler = ((Progress) -> Void)
+typealias ActionBeforeRunHandler = () -> Void
 
-	var interactionController: UIDocumentInteractionController?
+extension OCExtensionType {
+	static let action: OCExtensionType  =  OCExtensionType("app.action")
+}
 
-	init (item: OCItem, core: OCCore, into viewController: UIViewController) {
-		self.vcToPresentIn = viewController
+extension OCExtensionLocationIdentifier {
+	static let tableRow: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("tableRow") //!< Present as table row action
+	static let moreItem: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("moreItem") //!< Present in "more" card view for a single item
+	static let moreFolder: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("moreFolder") //!< Present in "more" options for a whole folder
+	static let toolbar: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("toolbar") //!< Present in a toolbar
+}
+
+class ActionExtension: OCExtension {
+	// MARK: - Custom Instance Properties.
+	var name: String
+	var category: ActionCategory
+
+	// MARK: - Init & Deinit
+	init(name: String, category: ActionCategory = .normal, identifier: OCExtensionIdentifier, locations: [OCExtensionLocationIdentifier]?, features: [String : Any]?, objectProvider: OCExtensionObjectProvider?, customMatcher: OCExtensionCustomContextMatcher?) {
+
+		self.name = name
+		self.category = category
+
+		super.init(identifier: identifier, type: .action, locations: locations, features: features, objectProvider: objectProvider, customMatcher: customMatcher)
+	}
+}
+
+class ActionContext: OCExtensionContext {
+	// MARK: - Custom Instance Properties.
+	weak var viewController: UIViewController?
+	weak var core: OCCore?
+	weak var query: OCQuery?
+	var items: [OCItem]
+
+	// MARK: - Init & Deinit.
+	init(viewController: UIViewController, core: OCCore, query: OCQuery? = nil, items: [OCItem], location: OCExtensionLocation, requirements: [String : Any]? = nil, preferences: [String : Any]? = nil) {
+		self.items = items
+
+		super.init()
+
+		self.viewController = viewController
 		self.core = core
-		self.item = item
+		self.location = location
+
+		self.query = query
+		self.requirements = requirements
+		self.preferences = preferences
+	}
+}
+
+class Action : NSObject {
+	// MARK: - Extension metadata
+	class var identifier : OCExtensionIdentifier? { return nil }
+	class var category : ActionCategory? { return .normal }
+	class var name : String? { return nil }
+	class var locations : [OCExtensionLocationIdentifier]? { return nil }
+	class var features : [String : Any]? { return nil }
+
+	// MARK: - Extension creation
+	class var actionExtension : ActionExtension {
+		let objectProvider : OCExtensionObjectProvider = { (_ rawExtension, _ context, _ error) -> Any? in
+			if let actionExtension = rawExtension as? ActionExtension,
+				let actionContext   = context as? ActionContext {
+				return self.init(for: actionExtension, with: actionContext)
+			}
+
+			return nil
+		}
+
+		let customMatcher : OCExtensionCustomContextMatcher  = { (context, priority) -> OCExtensionPriority in
+			if let actionContext = context as? ActionContext,
+				self.applicablePosition(forContext: actionContext) == .none {
+				// Exclude actions whose applicablePosition returns .none
+				return .noMatch
+			}
+
+			if let actionContext = context as? ActionContext {
+				let priority = OCExtensionPriority(rawValue: priority.rawValue + UInt(self.applicablePosition(forContext:actionContext).rawValue))!
+				print("LOG ---> priority = \(priority.rawValue) for extension \(context.location?.identifier?.rawValue)")
+				return priority
+			}
+
+			// Additional filtering (f.ex. via OCClassSettings, Settings) goes here
+			return priority
+		}
+
+		return ActionExtension(name: name!, category: category!, identifier: identifier!, locations: locations, features: features, objectProvider: objectProvider, customMatcher: customMatcher)
+	}
+
+	// MARK: - Extension matching
+	class func applicablePosition(forContext: ActionContext) -> ActionPosition {
+		return .middle
+	}
+
+	// MARK: - Finding actions
+	class func sortedApplicableActions(for context: ActionContext) -> [Action] {
+		var sortedActions : [Action] = []
+
+		if let matches = try? OCExtensionManager.shared.provideExtensions(for: context) {
+			for match in matches {
+				if let action = match.extension.provideObject(for: context) as? Action {
+					sortedActions.append(action)
+				}
+			}
+		}
+
+		sortedActions.sort { (action1, action2) -> Bool in
+			return action1.position.rawValue < action2.position.rawValue
+		}
+
+		return sortedActions
+	}
+
+	// MARK: - Action metadata
+	var context : ActionContext
+	var actionExtension: ActionExtension
+	var core : OCCore
+
+	// MARK: - Action creation
+	required init(for actionExtension: ActionExtension, with context: ActionContext) {
+		self.actionExtension = actionExtension
+		self.context = context
+		self.core = context.core!
+
 		super.init()
 	}
 
-	func presentActionsCard(with actions: [Action], completion: () -> Void) {
-		self.moreViewController = actionsViewController(with: actions, for: item, core: core)
-		vcToPresentIn?.present(asCard: moreViewController!, animated: true)
-	}
+	// MARK: - Execution metadata
+	var progressHandler : ActionProgressHandler?     // to be filled before calling run(), provideStaticRow(), provideContextualAction(), etc. if desired
+	var completionHandler : ActionCompletionHandler<Any>? // to be filled before calling run(), provideStaticRow(), provideContextualAction(), etc. if desired
+	var beforeRunHandler: ActionBeforeRunHandler? // to be filled before calling run(), provideStaticRow(), provideContextualAction(), etc. if desired
 
-	func actionsViewController(with actions: [Action], for item: OCItem, core: OCCore) -> MoreViewController {
-
-		let header = MoreViewHeader(for: item, with: core)
-		let tableViewController = MoreStaticTableViewController(style: .grouped)
-		let moreViewController: MoreViewController = MoreViewController(item: item, core: core, header: header, viewController: tableViewController)
-
-		var rows: [StaticTableViewRow] = []
-
-		for action in actions {
-
-			var style: StaticTableViewRowButtonStyle
-			switch action.type {
-			case .destructive:
-				style = .destructive
-			default:
-				style = .plainNonOpaque
-			}
-
-			let row: StaticTableViewRow = StaticTableViewRow(buttonWithAction: { (_, _) in
-				moreViewController.dismiss(animated: true, completion: {
-					action.completion?(item, core, self.vcToPresentIn!)
-				})
-			}, title: action.name, style: style)
-
-			rows.append(row)
+	// MARK: - Action implementation
+	func run() {
+		if beforeRunHandler != nil {
+			beforeRunHandler!()
 		}
 
-		let title = NSAttributedString(string: "Actions".localized, attributes: [NSAttributedStringKey.font: UIFont.systemFont(ofSize: 20, weight: .heavy)])
-
-		let section = MoreStaticTableViewSection(headerAttributedTitle: title, identifier: "actions-section", rows: rows)
-
-		tableViewController.addSection(section)
-		return moreViewController
+		if completionHandler != nil {
+			completionHandler!(Result.success(true))
+		}
 	}
 
-	func openIn(completion: (() -> Void)? = nil) -> Action {
-		let action = Action(with: "Open in".localized, completion: { (item, core, vcToPresentIn) in
-			let controller = DownloadFileProgressHUDViewController()
-			controller.present(on: vcToPresentIn)
-
-			if let downloadProgress = core.downloadItem(item, options: nil, resultHandler: { (error, _, _, file) in
-				if error != nil {
-					Log.log("Error \(String(describing: error)) downloading \(String(describing: item.path)) in openIn function")
-				} else {
-					controller.dismiss(animated: true, completion: {
-						self.interactionController = UIDocumentInteractionController(url: file!.url)
-						self.interactionController?.presentOptionsMenu(from: .zero, in: vcToPresentIn.view, animated: true)
-					})
-				}
-				completion?()
-			}) {
-				controller.attach(progress: downloadProgress)
-			} else {
-				let alert = UIAlertController(with: "No Network connection", message: "No network connection")
-				vcToPresentIn.present(alert, animated: true)
-			}
-		}, type: .regular)
-		return action
+	// MARK: - Action UI elements
+	func provideStaticRow() -> StaticTableViewRow? {
+		return StaticTableViewRow(buttonWithAction: { (_ row, _ sender) in
+				self.run()
+		}, title: actionExtension.name, style: actionExtension.category == .destructive ? .destructive : .plain, identifier: actionExtension.identifier.rawValue)
 	}
 
-	func duplicate(completion: (() -> Void)? = nil) -> Action {
-		let action = Action(with: "Duplicate", completion: { (item, core, viewcontroller) in
-
-			guard let viewController = viewcontroller as? ClientQueryViewController else {
-				return
-			}
-
-			var name: String = "\(item.name!) copy"
-
-			if item.type != .collection {
-				let itemName = item.nameWithoutExtension()
-				var fileExtension = item.fileExtension()
-
-				if fileExtension != "" {
-					fileExtension = ".\(fileExtension)"
-				}
-
-				name = "\(itemName) copy\(fileExtension)"
-			}
-
-			if let progress = core.copy(item, to: viewController.query.rootItem, withName: name, options: nil, resultHandler: { (error, _, item, _) in
-				if error != nil {
-					Log.log("Error \(String(describing: error)) deleting \(String(describing: item?.path))")
-				}
-				completion?()
-			}) {
-				viewController.progressSummarizer?.startTracking(progress: progress)
-			}
-
-		}, type: .regular)
-
-		return action
+	func provideContextualAction() -> UIContextualAction? {
+		return UIContextualAction(style: actionExtension.category == .destructive ? .destructive : .normal, title: self.actionExtension.name, handler: { (_ action, _ view, uiCompletionHandler) in
+			uiCompletionHandler(true)
+			self.run()
+		})
 	}
 
-	func move(completion: (() -> Void)? = nil) -> Action {
-		let action = Action(with: "Move".localized, completion: { (item, core, viewController) in
-
-			guard let viewController = viewController as? ClientQueryViewController else {
-				return
-			}
-
-			let directoryPickerVC = ClientDirectoryPickerViewController(core: core, path: "/", completion: { (selectedDirectory) in
-
-				if let progress = core.move(item, to: selectedDirectory, withName: item.name, options: nil, resultHandler: { (error, _, _, _) in
-					if error != nil {
-						Log.log("Error \(String(describing: error)) moving \(String(describing: item.path))")
-					}
-					completion?()
-				}) {
-					viewController.progressSummarizer?.startTracking(progress: progress)
-				}
-			})
-
-			let pickerNavigationController = ThemeNavigationController(rootViewController: directoryPickerVC)
-			viewController.navigationController?.present(pickerNavigationController, animated: true)
-		}, type: .regular)
-
-		return action
+	// MARK: - Action metadata
+	func iconForLocation(_ location: OCExtensionLocationIdentifier) -> UIImage? {
+		return nil
 	}
 
-	func delete(completion: (() -> Void)? = nil) -> Action {
-		let action = Action(with: "Delete".localized, completion: { (item, core, viewController) in
-			let alertController = UIAlertController(
-				with: item.name!,
-				message: "Are you sure you want to delete this item from the server?".localized,
-				destructiveLabel: "Delete".localized,
-				preferredStyle: UIDevice.current.isIpad() ? UIAlertControllerStyle.alert : UIAlertControllerStyle.actionSheet,
-				destructiveAction: {
-					if let progress = core.delete(item, requireMatch: true, resultHandler: { (error, _, _, _) in
-						if error != nil {
-							Log.log("Error \(String(describing: error)) deleting \(String(describing: item.path))")
-						}
-						completion?()
-					}) {
-						if let viewController = viewController as? ClientQueryViewController {
-							viewController.progressSummarizer?.startTracking(progress: progress)
-						}
-					}
-			})
+	var icon : UIImage? {
+		if let locationIdentifier = context.location?.identifier {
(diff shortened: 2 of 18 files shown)


ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -51,6 +51,8 @@ class ClientQueryViewController: UITableViewController, Themeable {
 	private var observerContext : UnsafeMutableRawPointer
 	var refreshController: UIRefreshControl?
 
+	var interactionController: UIDocumentInteractionController?
+
 	// MARK: - Init & Deinit
 	public init(core inCore: OCCore, query inQuery: OCQuery) {
 		observerContext = UnsafeMutableRawPointer(&observerContextValue)
```
mneuwert: @pablocarmu I checked the code and I think `observerContext` is not really required, since it is used when e.g. you observe two objects for the change of same keypath and want to differentiate between them when the observer iis called.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -658,6 +660,30 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		}
 
 	}
+
+	func openInRow(_ item: OCItem, viewDidAppearHandler: ClientActionVieDidAppearHandler? = nil, completionHandler: ClientActionCompletionHandler? = nil) {
+
+		if let progress = self.core?.downloadItem(item, options: nil, resultHandler: { (error, _, _, file) in
+			if error != nil {
```
mneuwert: @pablocarmu If I remember correctly completion handler of` downloadItem()` is not guaranteed to be called on the main thread, so are we missing dispatch on main thingy here?

mneuwert: @pablocarmu I guess I had to wait until you remove [WIP] to start reviewing..

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -658,6 +655,49 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		}
 
 	}
+
+	func openInRow(_ item: OCItem, cell: ClientItemCell? = nil, viewDidAppearHandler: ClientActionVieDidAppearHandler? = nil, completionHandler: ClientActionCompletionHandler? = nil) {
+
+		let controller = DownloadFileProgressHUDViewController()
+
+		if core!.reachabilityMonitor.available {
```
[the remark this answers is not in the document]
pablocarmu: So I'll remove the check and leave a `TODO` and when the SDK has support for this, implement it. 

I'll open an issue with the info to track this if you agree with the approach.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -658,6 +655,49 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+			if let progress = self.core?.downloadItem(item, options: nil, resultHandler: { (error, _, _, file) in
+				if error != nil {
+					Log.log("Error \(String(describing: error)) downloading \(String(describing: item.path)) in openIn function")
+					completionHandler?(false)
+				} else {
+					controller.dismiss(animated: true, completion: {
+						self.interactionController = UIDocumentInteractionController(url: file!.url)
+						self.interactionController?.delegate = self
+						OnMainThread {
```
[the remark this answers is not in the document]
pablocarmu: I'll remove the main thread call.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -658,6 +655,49 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+					})
+				}
+			}) {
+				OnMainThread {
+					controller.present(on: self)
+					controller.attach(progress: progress)
+				}
+			}
+		} else {
+			OnMainThread {
```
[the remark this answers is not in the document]
pablocarmu: I'll remove the main thread call.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -658,6 +655,49 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+								self.interactionController?.presentOptionsMenu(from: cell!.frame, in: self.view, animated: true)
+							} else {
+								self.interactionController?.presentOptionsMenu(from: .zero, in: self.view, animated: true)
+							}
+						}
+						completionHandler?(true)
+					})
+				}
+			}) {
+				OnMainThread {
```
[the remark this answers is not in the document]
pablocarmu: I'll take the following approach then:
1. Present the controller.
2. Call the `if let progress = self.core.downloadItem...`
2. Attach the progress if any.
3. If the completion handler is called before the return, no strange behavior should appear. The controller should be dismissed.

What do you think?

ownCloud/UI Elements/DownloadFileProgressHUDViewController.swift
```diff
@@ -0,0 +1,223 @@
...
+* For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
+* You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+*
+*/
+
+import UIKit
+
+class DownloadFileProgressHUDViewController: UIViewController {
+
+	private let progressViewSidesConstraintConstant: CGFloat = 20
```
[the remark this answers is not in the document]
pablocarmu: I'll test this with an iPad and see how it looks. In case it looks bad, I'll attach new constraints for iPad layouts.

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -307,6 +310,79 @@ class DisplayViewController: UIViewController {
...
+		tableViewController.addSection(MoreStaticTableViewSection(headerAttributedTitle: title, identifier: "actions-section", rows: [openInRow]))
+
+		self.present(asCard: moreViewController, animated: true)
+	}
+
+	// MARK: - Actions
+	func openInRow(_ item: OCItem, button: UIBarButtonItem? = nil) {
+
+		if source == nil {
+			if !core.reachabilityMonitor.available {
```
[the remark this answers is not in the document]
pablocarmu: I've been thinking about this and tried a couple of solutions, but I've faced a lot of issues trying to make it work easily.

What I think could be abstracted it's the creation of the card with all the actions and but not the action themselves (or at least not all).

I've tried to make something like this

```Swift
typealias ActionCompletion = ((_ item: OCItem, _ core: OCCore, _ vcToPresent: UIViewController) -> Void)?

enum ActionType {
	case destructive
	case regular
}

struct Action {
	var name: String
	var type: ActionType
	var completion: ActionCompletion

	init(with name: String, completion: ActionCompletion, type: ActionType) {
		self.name = name
		self.completion = completion
		self.type = type

	}
}
```

To be able to abstract the actions into an struct, then I created a card using the more style and those actions

```Swift
	func actionsViewController(with actions: [Action], for item: OCItem, core: OCCore) -> MoreViewController {

		let header = MoreViewHeader(for: item, with: core)
		let tableViewController = MoreStaticTableViewController(style: .grouped)
		let moreViewController: MoreViewController = MoreViewController(item: item, core: core, header: header, viewController: tableViewController)

		var rows: [StaticTableViewRow] = []

		for action in actions {

			var style: StaticTableViewRowButtonStyle
			switch action.type {
			case .destructive:
				style = .destructive
			default:
				style = .plainNonOpaque
			}

			let row: StaticTableViewRow = StaticTableViewRow(buttonWithAction: { (_, _) in
				moreViewController.dismiss(animated: true, completion: {
					action.completion?(item, core, self.vcToPresentIn!)
				})
			}, title: action.name, style: style)

			rows.append(row)
		}

		let title = NSAttributedString(string: "Actions".localized, attributes: [NSAttributedStringKey.font: UIFont.systemFont(ofSize: 20, weight: .heavy)])

		let section = MoreStaticTableViewSection(headerAttributedTitle: title, identifier: "actions-section", rows: rows)

		tableViewController.addSection(section)
		return moreViewController
	}
```

This works fine for `Move` and `Duplicate` because then you can create functions like this one:

```Swift
func move(completion: (() -> Void)? = nil) -> Action {
		let action = Action(with: "Move".localized, completion: { (item, core, viewController) in

			guard let viewController = viewController as? ClientQueryViewController else {
				return
			}

			let directoryPickerVC = ClientDirectoryPickerViewController(core: core, path: "/", completion: { (selectedDirectory) in

				if let progress = core.move(item, to: selectedDirectory, withName: item.name, options: nil, resultHandler: { (error, _, _, _) in
					if error != nil {
						Log.log("Error \(String(describing: error)) moving \(String(describing: item.path))")
					}
					completion?()
				}) {
					viewController.progressSummarizer?.startTracking(progress: progress)
				}
			})

			let pickerNavigationController = ThemeNavigationController(rootViewController: directoryPickerVC)
			viewController.navigationController?.present(pickerNavigationController, animated: true)
		}, type: .regular)

		return action
	}
```

But this adds a problem when each view controller that wants the action also wants a custom implementation of what happens on every momment, maybe they want a custom UI or something that with this implementation it's not posible.

So another solution came to my mind, make all the abstraction returns a Result:

```Swift
enum Result<T> {
	case failure(Error)
	case success(T)
}
```

and make the call something like:

```Swift
func move(progress: @escaping (Progress) -> Void, result: (Result<Bool>) -> Void) {}
```

But this also has the problem of the custom implementation, so I think the best solution pass through abstract the way the card controller is created and create/attach each action inside the view controller it's going to use it. This way we avoid stupid unneded castings inside a common method and gives us the flexibility to insert the actions we want in all the possible scenarios.

I've uploaded a [gist](https://gist.github.com/pablocarmu/65c2d9f7c8af5391687f2ce09dfaf0c4) with the complete implementation of the first solution I explained here so you can take a look and judge it.

ownCloud/Client/ClientQueryViewController.swift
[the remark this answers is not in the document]
felix-schwarz: Thanks!

ownCloud/Client/ClientQueryViewController.swift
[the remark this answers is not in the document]
felix-schwarz: Sounds good to me!

ownCloud/UI Elements/DownloadFileProgressHUDViewController.swift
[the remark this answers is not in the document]
felix-schwarz: Thanks!

ownCloud/Viewer/DisplayViewController.swift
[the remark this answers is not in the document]
felix-schwarz: @pablocarmu Finally coming around to answer. Thanks a ton for the extensive feedback and thinking on this!

[That gist](https://gist.github.com/pablocarmu/65c2d9f7c8af5391687f2ce09dfaf0c4) already comes pretty close to what I had in mind last week. I didn't think of actions as their own objects or structs, though. Instead, I'd have tried to simply move over the methods we already have in `ClientQueryViewController`.

The idea of having actions as fully independent building blocks is appealing, though. And it immediately makes me think of putting them into `OCExtension`s:
- a new `OCExtension` subclass `OCActionExtension` (or just `ActionExtension`) could add infos like name and actionType
- control over where an action is present (i.e. only in the action sheet, or also in the table row actions) could leverage the `locations` feature of extensions (through different `OCExtensionLocationIdentifier`s)
- control over whether an action should be presented for an item could leverage the `customMatcher` feature of extensions. The same feature could also be used to establish an order of appearance (even taking into account the location of appearance!) via the matching score (and sorting by that score).
- the identifier of the action extension in combination with `OCClassSettings` and custom matching could be used to allow MDM-side configuration of actions (especially to remove actions). Blocked actions, f.ex. would just not match.
- a new `OCExtensionContext` subclass `OCActionContext` (or just `ActionContext`) could be used to pass along information required by the actions like items, core, query, viewController, etc. to make matching decisions, but also to present custom UI
- that context subclass could be extended as needed to accommodate for any special requirements, including completion handlers and results
- building this on `OCExtension`s of course also would allow the addition of customer-specific actions without requiring any changes in the app code itself

What do you think?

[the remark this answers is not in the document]
pablocarmu: I made a quick PoC of this in a [gist](https://gist.github.com/pablocarmu/0c3f6967676affdb273b5b38af8c81af)

I didn't had time to check if it works inside the app, but I think this almost matches with what you wrote, I'm still need to get better the idea but I think this could be a base to start a discussion about the implementation.

Even if this PR gets more time of what was initially though, I think we should address a correct implementation of the actions.

What do you think about the gist?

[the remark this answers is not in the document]
felix-schwarz: Thanks for putting the gist together!

I've forked and extended it ([gist here](https://gist.github.com/felix-schwarz/0e85ec0ddd378da221a343469cf0a507)) to include everything I wrote (except `OCClassSettings` integration, which is just a comment in the spot I'd put it) and test a few ideas.

One of these ideas was to let `Action` instances directly create `StaticTableViewRow`s and `UIContextualAction`s for use in the UI. The main idea for this was to have convenient access to these through default "builders" in the base `Action` class, but to also have the flexibility to customize them in the respective `Action` subclass as needed.

jesmrec:
Added test plan here: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Open%20In.md

ownCloud/AppDelegate.swift
```diff
@@ -51,6 +51,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		OCExtensionManager.shared.addExtension(WebViewDisplayViewController.displayExtension)
 		OCExtensionManager.shared.addExtension(PDFViewerViewController.displayExtension)
 		OCExtensionManager.shared.addExtension(ImageDisplayViewController.displayExtension)
+		OCExtensionManager.shared.addExtension(OpenInAction.actionExtension)
```
felix-schwarz: Only cosmetic: a space here would help to visually group the different kinds of extension.

ownCloud/Client/Actions/Action.swift
```diff
@@ -0,0 +1,218 @@
...
+			guard let actionContext = context as? ActionContext else {
+				return priority
+			}
+
+			if self.applicablePosition(forContext: actionContext) == .none {
+				// Exclude actions whose applicablePosition returns .none
+				return .noMatch
+			}
+
+			if actionContext.items[0].type == OCItemType.collection, identifier!.rawValue == "com.owncloud.action.openin" {
```
felix-schwarz: Since this is specific to the open-in action, it should move to `OpenInAction.applicablePosition(forContext:)` and return `.none` there.

ownCloud/Client/Actions/Action.swift
```diff
@@ -0,0 +1,218 @@
...
+				return .noMatch
+			}
+
+			let priority = OCExtensionPriority(rawValue: priority.rawValue + UInt(self.applicablePosition(forContext:actionContext).rawValue))!
+			return priority
```
felix-schwarz: I think you can just return `priority` here as `Action.sortedApplicableActions` will take care of sorting the actions according to their `applicablePosition`.

ownCloud/AppDelegate.swift
pablocarmu: Made it!

ownCloud/Client/Actions/Action.swift
pablocarmu: You're right, it didn't come to my mind, it's changed now.

ownCloud/Client/Actions/Action.swift
pablocarmu: Done, I wasn't using `Action.sortedApplicableActions`, now I am.

ownCloud/Client/Actions/Action.swift
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	// MARK: - Execution metadata
+	var progressHandler : ActionProgressHandler?     // to be filled before calling run(), provideStaticRow(), provideContextualAction(), etc. if desired
+	var completionHandler : ActionCompletionHandler? // to be filled before calling run(), provideStaticRow(), provideContextualAction(), etc. if desired
+	var beforeRunHandler: ActionBeforeRunHandler? // to be filled before calling run(), provideStaticRow(), provideContextualAction(), etc. if desired
+
+	// MARK: - Action implementation
+	func run() {
+		if beforeRunHandler != nil {
```
felix-schwarz: Might be a good idea to move the code running `beforeRunHandler` to a separate `willRun` method, so that
- subclasses can just call `self.willRun()`
- more code can be added before without having to change the subclasses

ownCloud/Client/Actions/Action.swift
felix-schwarz: Awesome, thanks!

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -307,11 +312,45 @@ class DisplayViewController: UIViewController {
...
+		let moreViewController = MoreViewController(item: item, core: core, header: header, viewController: tableViewController)
+
+		let title = NSAttributedString(string: "Actions", attributes: [NSAttributedStringKey.font: UIFont.systemFont(ofSize: 20, weight: .heavy)])
+
+		let actionsLocation = OCExtensionLocation(ofType: OCExtensionType.action, identifier: nil)
+		let actionContext = ActionContext(viewController: self, core: core, items: [item, rootItem], location: actionsLocation)
+		let actions = Action.sortedApplicableActions(for: actionContext)
+
+		actions.forEach({
+			$0.beforeRunHandler = {
```
felix-schwarz: Nice!

ownCloud/Client/Actions/Actions+Extensions/DeleteAction.swift
```diff
@@ -0,0 +1,54 @@
...
+	// MARK: - Action implementation
+	override func run() {
+		guard context.items.count > 0, let viewController = context.viewController else {
+			completionHandler?(NSError(ocError: .errorInsufficientParameters))
+			return
+		}
+
+		beforeRunHandler?()
+
+		let item = context.items[0]
```
felix-schwarz: To futureproof the actions, it should operate on all passed items (same for all other actions).

ownCloud/Client/Actions/Actions+Extensions/DeleteAction.swift
```diff
@@ -0,0 +1,54 @@
...
+
+		let alertController = UIAlertController(
+			with: item.name!,
+			message: "Are you sure you want to delete this item from the server?".localized,
```
felix-schwarz: These two messages, too, would need to be adapted for multiple files.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -0,0 +1,60 @@
...
+	override func run() {
+		guard context.items.count > 0 else {
+			completionHandler?(NSError(ocError: .errorInsufficientParameters))
+			return
+		}
+
+		beforeRunHandler?()
+
+		let item = context.items[0]
+		let rootItem = context.items[1]
```
felix-schwarz: `items` should only contain items to perform the action on, and might end up having different parent items. Therefore, the `parentItem` should be retrieved from `OCCore` dynamically.

I wrote an OCItem category that does this for you:
```swift
extension OCItem {
	func parentItem(from core: OCCore, completionHandler: ((_ error: Error?, _ parentItem: OCItem?) -> Void)? = nil) -> OCItem? {
		var parentItem : OCItem?

		if let parentItemIdentifier = self.parentFileID {
			var waitGroup : DispatchGroup?

			if completionHandler == nil {
				waitGroup = DispatchGroup()
				waitGroup?.enter()
			}

			core.retrieveItemFromDatabase(forFileID: parentItemIdentifier) { (error, _, item) in
				if completionHandler == nil {
					parentItem = item
					waitGroup?.leave()
				} else {
					completionHandler?(error, item)
				}
			}

			waitGroup?.wait()
		}

		return parentItem
	}
}
```

It can be used both synchronous and asynchronously:
```swift
 // Synchronous
let parentItem = item.parentItem(from: core)

 // Asynchronous (and preferred)
item.parentItem(from: core) { (_ error, _ parenItem) -> Void in 
    OnMainThread {
        // work with parentItem
    }
}
```

ownCloud/Client/Actions/Actions+Extensions/OpenInAction.swift
```diff
@@ -0,0 +1,69 @@
...
+
+		OnMainThread {
+			controller.present(on: viewController) {
+				if let progress = self.core.downloadItem(item, options: nil, resultHandler: { (error, _, _, file) in
+					if error != nil {
+						Log.log("Error \(String(describing: error)) downloading \(String(describing: item.path)) in openIn function")
+						self.completionHandler?(error!)
+					} else {
+
+						controller.dismiss(animated: true, completion: {
```
felix-schwarz: This resultHandler is called from a background thread, but could call `controller.dismiss` here. At least the `else` block here should be wrapped into `OnMainThread`.

ownCloud/Client/Actions/Actions+Extensions/OpenInAction.swift
```diff
@@ -0,0 +1,69 @@
...
+							self.completionHandler?(nil)
+							self.interactionController = UIDocumentInteractionController(url: file!.url)
+							self.interactionController?.delegate = self
+							OnMainThread {
+								self.interactionController?.presentOptionsMenu(from: .zero, in: viewController.view, animated: true)
+							}
+						})
+					}
+				}) {
+					OnMainThread {
```
felix-schwarz: This OnMainThread should not be needed as UI completion handler are generally executed on the main thread.

ownCloud/Client/Actions/Actions+Extensions/OpenInAction.swift
```diff
@@ -0,0 +1,69 @@
...
+			return
+		}
+
+		beforeRunHandler?()
+
+		let item = context.items[0]
+
+		let controller = DownloadFileProgressHUDViewController()
+
+		OnMainThread {
```
felix-schwarz: Maybe we should make it a convention that `run()` on actions may only be called from the main thread (and log a warning in `preRun()` if it's called on anything else, so we catch that early). You can then get rid of `OnMainThread` here.

ownCloud/Client/Actions/Actions+Extensions/OpenInAction.swift
```diff
@@ -0,0 +1,69 @@
...
+
+import ownCloudSDK
+
+class OpenInAction: Action {
+	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.openin") }
+	override class var category : ActionCategory? { return .normal }
+	override class var name : String { return "Open in".localized }
+
+	override class func applicablePosition(forContext: ActionContext) -> ActionPosition {
+		if forContext.items[0].type == .collection {
```
felix-schwarz: As mentioned in a previous comment, to make `Action`s truly universally usable throughout the app, it should look at all `items`. In this case, if any of the items is a `.collection`, that'd mean this action isn't available.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -0,0 +1,60 @@
...
+
+		if item.type != .collection {
+			let itemName = item.nameWithoutExtension()
+			var fileExtension = item.fileExtension()
+
+			if fileExtension != "" {
+				fileExtension = ".\(fileExtension)"
+			}
+
+			name = "\(itemName) copy\(fileExtension)"
```
felix-schwarz: This will fail when trying to make a second copy without renaming the first one before.

Makes me wonder… should the `OCCore` provide a method to suggest a likely unused, numbered variant of an item's name? Something like `core.suggestUnusedName(for: item, withAppendedLabel: "copy")`?

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -307,11 +312,45 @@ class DisplayViewController: UIViewController {
 		self.noNetworkLabel?.isHidden = true
 		self.showPreviewButton?.isHidden = false
 	}
+
+	@objc func optionsBarButtonPressed() {
```
felix-schwarz: Would be great to see this and `moreButtonTapped` merged into a single method, f.ex. into a class-level method of the `Action` class?

ownCloud/Client/Actions/Actions+Extensions/OpenInAction.swift
pablocarmu: I think it depends on the action itself. Of course, a `Delete` action should delete as many items as you want, but for open in, this action should not be available in multiselection environments because of `UIDocumentInteractionController` can be only created with one `URL` and I think we should not open more than one `UIDocumentInteractionController`. Same for the `Rename` action.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
pablocarmu: It could be one approach, another option is that the SDK takes care itself about the naming, so call `duplicate` function and the SDK provides the name, create the duplicate file and call the query delegate with the new item.

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -20,6 +20,7 @@ import UIKit
 import ownCloudSDK
 
 struct DisplayViewConfiguration {
+	weak var rootItem: OCItem!
```
felix-schwarz: I think this, too, is a left-over addition and can be removed. FWIW it doesn't seem to be needed/used anywhere.

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -40,9 +41,10 @@ protocol DisplayViewEditingDelegate: class {
 
 class DisplayViewController: UIViewController {
 
-	private let ICON_IMAGE_SIZE: CGSize = CGSize(width: 200.0, height: 200.0)
+	private let iconImageSize: CGSize = CGSize(width: 200.0, height: 200.0)
 
 	// MARK: - Configuration
+	weak var rootItem: OCItem!
```
felix-schwarz: Same here… seems to be a left-over addition and can be removed. Also doesn't seem to be needed/used anywhere.

ownCloud/Client/Actions/Action.swift
```diff
@@ -0,0 +1,260 @@
...
+			actionWillRunHandler!()
+		}
+	}
+
+	func run() {
+		if completionHandler != nil {
+			completionHandler!(nil)
+		}
+	}
+
```
felix-schwarz: **Completely optional**, but as I reviewed the code I thought this would make for an even nicer implementation overall (and may possibly also prove useful further down the road):
- add new `func completed(with: Error? = nil)` that calls the completionHandler (if one exists)
- throughout the actions, call `self.completed(error)` (or `self.completed()` where applicable) instead of the `completionHandler` directly.
- same for progressHandler with f.ex. a `func publish(progress: Progress)` that calls the `progressHandler` and replaces direct calls of it

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -746,40 +603,18 @@ extension ClientQueryViewController: UISearchResultsUpdating {
...
+		guard let item = cell.item else {
+			return
+		}
 
-			let moveRow: StaticTableViewRow = StaticTableViewRow(buttonWithAction: { [weak self] (_, _) in
-				moreViewController.dismiss(animated: true, completion: {
-					self?.move(item)
-				})
-				}, title: "Move".localized, style: .plainNonOpaque)
+		let actionsLocation = OCExtensionLocation(ofType: OCExtensionType.action, identifier: nil)
```
felix-schwarz: Please add the location identifier here:
```suggestion
		let actionsLocation = OCExtensionLocation(ofType: .action, identifier: .moreItem)
```

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -307,11 +310,24 @@ class DisplayViewController: UIViewController {
 		self.noNetworkLabel?.isHidden = true
 		self.showPreviewButton?.isHidden = false
 	}
+
+	@objc func optionsBarButtonPressed() {
+
+		let actionsLocation = OCExtensionLocation(ofType: OCExtensionType.action, identifier: nil)
```
felix-schwarz: Please add the location identifier here, too:
```suggestion
		let actionsLocation = OCExtensionLocation(ofType: .action, identifier: .moreItem)
```

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -275,52 +271,13 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
 			case .file:
-				let itemViewController = DisplayHostViewController(for: rowItem, with: core)
+				let itemViewController = DisplayHostViewController(for: rowItem, with: core, root: query.rootItem!)
 				self.navigationController?.pushViewController(itemViewController, animated: true)
 		}
 
 		tableView.deselectRow(at: indexPath, animated: true)
 	}
 
-	override func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
```
felix-schwarz: For the re-implementation of row actions, you can use `OCExtensionLocation(ofType: .action, identifier: .tableRow)` as location for building the `ActionContext`.

For actions that shouldn't turn up as row actions, `.tableRow` should not be included in the `ActionSubclass.locations` array, so it doesn't match.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -507,84 +464,6 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
-			}) {
-				self.progressSummarizer?.startTracking(progress: progress)
-			}
-		})
-
-		let pickerNavigationController = ThemeNavigationController(rootViewController: directoryPickerVC)
-		self.navigationController?.present(pickerNavigationController, animated: true)
-	}
-
 	func createFolder(viewDidAppearHandler: ClientActionVieDidAppearHandler? = nil, completionHandler: ClientActionCompletionHandler? = nil) {
```
felix-schwarz: "Create folder" should ultimately also move to an `Action` subclass. To make it possible to be found, we'd need to add a new location identifier:

```
extension OCExtensionLocationIdentifier {
	static let tableRow: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("tableRow") //!< Present as table row action
	static let moreItem: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("moreItem") //!< Present in "more" card view for a single item
	static let moreFolder: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("moreFolder") //!< Present in "more" options for a whole folder
	static let toolbar: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("toolbar") //!< Present in a toolbar
	// NEW:
	static let sortBar: OCExtensionLocationIdentifier = OCExtensionLocationIdentifier("sortBar") //!< Present in the sort bar on top of file lists
}
```

The `CreateFolderAction.locations` could then return `[.moreFolder, .sortBar]`, so it's only available when tapping the three dots in the navigation bar (in the future) or the "New folder" icon in the sort bar.

ownCloud/Client/Actions/Actions+Extensions/DeleteAction.swift
```diff
@@ -0,0 +1,78 @@
...
+*
+*/
+
+import ownCloudSDK
+
+class DeleteAction : Action {
+	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.delete") }
+	override class var category : ActionCategory? { return .destructive }
+	override class var name : String? { return "Delete".localized }
+
```
felix-schwarz: In order to make use of the location identifiers and control where an action should eventually turn up, please add 
```swift
override class var locations : [OCExtensionLocationIdentifier]? { return [.moreItem, .tableRow] }
```
to all `Action` subclasses (of course adjusting the returned locations array as needed).

pablocarmu:
I made all the changes but there are some things I want to comment:
1. I replaced the old implementation of _create folder_ with a new one, this time as an action, **BUT** it didn't come to my mind how to integrate it with the sort bar without rewriting the whole view. So I ask for the action in the sort bar delegate and call `run()`. If you, @felix-schwarz , know how to integrate it without too much effort into the sort bar please tell me. If not, I would like to address this in another PR.

2. I have to add more code to the custom matcher in the `Action` because previously was ignoring the location identifiers, so if you ask for example for the `.sortbar` extensions, it returns all the possible extensions.

felix-schwarz:
> I made all the changes but there are some things I want to comment:
> 
> 1. I replaced the old implementation of _create folder_ with a new one, this time as an action, **BUT** it didn't come to my mind how to integrate it with the sort bar without rewriting the whole view. So I ask for the action in the sort bar delegate and call `run()`. If you, @felix-schwarz , know how to integrate it without too much effort into the sort bar please tell me. If not, I would like to address this in another PR.

I'd be fine with a "hack" for that "Create folder" button for now, which just gets the actions for the `.sortBar` location and runs the first one. Never too late to rewrite the SortBar in the future… IF we ever allow/want other/more actions there.

> 2. I have to add more code to the custom matcher in the `Action` because previously was ignoring the location identifiers, so if you ask for example for the `.sortbar` extensions, it returns all the possible extensions.

That should not be necessary.

The `OCExtension.matchesContext()` should pass the `customMatcher` a `priority` value of `.noMatch` if the `ActionContext` has a location that's not included in the extension. 

I'll take a look.

felix-schwarz:
> I'd be fine with a "hack" for that "Create folder" button for now, which just gets the actions for the `.sortBar` location and runs the first one. Never too late to rewrite the SortBar in the future… IF we ever allow/want other/more actions there.

I see you already made that change. Thanks!

> > 1. I have to add more code to the custom matcher in the `Action` because previously was ignoring the location identifiers, so if you ask for example for the `.sortbar` extensions, it returns all the possible extensions.
> 
> That should not be necessary.
> 
> The `OCExtension.matchesContext()` should pass the `customMatcher` a `priority` value of `.noMatch` if the `ActionContext` has a location that's not included in the extension.
> 
> I'll take a look.

I tracked it down to a missing line of code in the OCExtension init method that would simply drop the locations passed to it: https://github.com/owncloud/ios-sdk/commit/8173316d4254c8a398a337c9678e57682be29f76

Moving to the latest version of the SDK addresses that issue, so that code in Action.swift is no longer needed:
```swift
			// Filter by location identifier i.e. .sortbar
 			let matches = self.locations?.filter({ (identifier) in
 				identifier == context.location?.identifier
 			})

 			guard matches != nil else {
 				return .noMatch
 			}

 			if matches!.count == 0 {
 				return .noMatch
 			}
```

pablocarmu:
I removed the extra code and updated the SDK.

jesmrec:
Let's start with QA here

jesmrec:
### (1)

1. Download a file in the app -> it is previewed
2. Tap on three-button on top right corner of the preview and select `open in`

Current: file is downloaded again

Not sure if this is the expected behaviour, so the file was downloaded before in order to be previewed. Does it need a new download to send to other app?

jesmrec:
### (2)

Download a PDF file. With the PDF viewer, no place for three-dot and open-in in preview (available in file list). It this desirable @michaelstingl ?

CC @mneuwert

jesmrec:
### (3)

With no connection in the device, select "open In" 

Current: file is tried to be downloaded waiting for connection
Expected: lack of connection notified to the user. no further actions. Same behaviour in regular download

iPadAir v12

jesmrec:
### (4)

With the same steps as before (no connection in device and open-in), by cancelling the operation, the bottom status bar does no vanish.

Expected: It is vanished

nov-27-2018 10-55-35

jesmrec:
(5) 

Look at this corner case:

1. In the web UI, an item is removed
2. The same item in the app, select "Open In" (in the time slot till file list is refreshed)

Current: item tries to be downloaded. Download screen disappears, but server-removed item is not removed in the app file list, even refreshing multiple times. From this point, all downloads are enqueued maybe waiting for the first one over a removed item. By killing the app and reopen, it starts to work again.

worthy to fix in the current branch?

pablocarmu:
@jesmrec 

>    Download a file in the app -> it is previewed
    Tap on three-button on the top right corner of the preview and select open in

>Current: file is downloaded again
Not sure if this is the expected behavior, so the file was downloaded before in order to be previewed. Does it need a new download to send to other apps?

>NOTE: Same behaviour in file list, by opening-in twice

I see your point but what happens if the file changed in the server? This way we're always downloading the latest version.

In any case, I think the app should call the download method and if we want or not to download the file, the SDK should be responsible for this choice, retrieving the file to the app through the completion handler.

> With no connection in the device, select "open In"

> Current: file is tried to be downloaded waiting for connection
Expected: lack of connection notified to the user. no further actions. Same behaviour in regular download

Should we then don't allow the user to trigger the action if there is no connection?

> With the same steps as before (no connection in device and open-in), by canceling the operation, the bottom status bar does no vanish.

> Expected: It is vanished

This is related to the previous one. If we don't allow to download files when there is no connection, this problem is not going to rise up anymore. I'll take a look at the reason of why is it empty and why it does not disappear

jesmrec:
> Should we then don't allow the user to trigger the action if there is no connection?

This is in order to keep consistence with the already offline behaviour in the app. If you tap on a file to download it with no connection, you are notified of such issue, f. ex.

jesmrec:
> I see your point but what happens if the file changed in the server? This way we're always downloading the latest version.

it is posible to check if the file has a new version, isn't it? (via etag). Maybe the SDK could take care of  this.

jesmrec:
(1) -> https://github.com/owncloud/ios-app/issues/178
(3) + (4) -> https://github.com/owncloud/ios-app/issues/179
(2) -> wont fix

PR #138, opened 2018-10-29
First implementation of PDF viewer feature

mneuwert:
## Description
PDF viewer shall be implemented allowing comfortable viewing of PDF files.

## Related Issue
#87 

## Motivation and Context
We would like to utilise PDFKit features to provide user experience similar to iBooks app while browsing PDF files. 

## Types of changes

- [x] PDF viewer with thumbnail view which can be hidden
- [x] Go to page X feature
- [x] Search feature showing the hits with page numbers and allowing to jump to selected search hit
- [x] Separate PDF navigation screen with two tabs: table of contents / larger thumbnails for all pages

## Checklist:


ownCloud/Viewer/PDFThumbnailsCollectionViewController.swift
```diff
@@ -0,0 +1,194 @@
...
+        weak var weakBlockOperation = blockOperation
+
+        blockOperation.addExecutionBlock { [unowned self] in
+            let thumbnailImage = page.thumbnail(of: size, for: .cropBox)
+            self.thumbnailCache.setObject(thumbnailImage, forKey: page.label! as NSString)
+            if (weakBlockOperation?.isCancelled)! {
+                return
+            }
+            DispatchQueue.main.async {
+                if let cell : PDFThumbnailCollectionViewCell = self.collectionView?.cellForItem(at: indexPath) as? PDFThumbnailCollectionViewCell {
```
pablocarmu: I've seen this in multiple places and it's only my preference of styling things, but you can chain `if let` and avoid a los of `{ }`

Something like:
```Swift
if let cell : PDFThumbnailCollectionViewCell = self.collectionView?.cellForItem(at: indexPath) as? PDFThumbnailCollectionViewCell, 
   let visibleCells = self.collectionView?.visibleCells,
   visibleCells.contains(cell) {
                            cell.imageView?.image = thumbnailImage
                            cell.pageLabel?.text = page.label
}
```

ownCloud/Viewer/PDFViewerViewController.swift
```diff
@@ -20,44 +20,368 @@ import UIKit
...
+
+    static let PDFGoToPageNotification = Notification(name: Notification.Name(rawValue: "PDFGoToPageNotification"))
+
+	fileprivate var gotoPageNotificationObserver : Any?
+
+    fileprivate let SEARCH_ANNOTATION_DELAY  = 3.0
+    fileprivate let ANIMATION_DUR = 0.25
+    fileprivate let THUMBNAIL_VIEW_WIDTH_MULTIPLIER: CGFloat = 0.15
+    fileprivate let THUMBNAIL_VIEW_HEIGHT_MULTIPLIER: CGFloat = 0.1
+    fileprivate let FILENAME_CONTAINER_TOP_MARGIN: CGFloat = 10.0
```
pablocarmu: I've seen that in some places you code constants in different styles, is there some reason to do it?

ownCloud/Viewer/PDFSearchTableViewCell.swift
```diff
@@ -0,0 +1,87 @@
...
+        setupSubviewsAndConstraints()
+    }
+
+    fileprivate func setupSubviewsAndConstraints() {
+
+        self.contentView.addSubview(titleLabel)
+        self.contentView.addSubview(pageLabel)
+
+        titleLabel.font = UIFont.systemFont(ofSize: titleFontSize, weight: UIFont.Weight.regular)
+        titleLabel.adjustsFontSizeToFitWidth = false
```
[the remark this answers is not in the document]
mneuwert: @pablocarmu yes, this line can be removed. The intention was to make the code more explicit

ownCloud/Viewer/PDFViewerViewController.swift
mneuwert: @pablocarmu for me it was not clear if we have any official coding conventions and I didn't find an example of constants in existing code. What is the preferred style?

ownCloud/Viewer/PDFThumbnailsCollectionViewController.swift
mneuwert: @pablocarmu yes, true. I can change it.

ownCloud/Viewer/PDFViewerViewController.swift
pablocarmu: Currently, we don't have any rule defined for this neither any complete style guide apart of what we have in the SwiftLint file.

I think we can discuss the style guide in the new opened issue here https://github.com/owncloud/ios-app/issues/139

felix-schwarz:
I found a couple of glitches worth fixing before diving into code review:

1) the name of the file is currently shown in the navigation bar and in an overlay. Having it in the navigation bar should be sufficient:

bildschirmfoto 2018-10-30 um 12 31 53

2) Tapping on the table of contents, a white background shows when dragging the table view down or up beyond its upper and lower end (rubberband-style):

bildschirmfoto 2018-10-30 um 12 35 26 2

The background has the correct color after switching to the the icon view and back.

3) When I search for something and tap on a hit, I see the search interface dismissing twice.

4) While entering something in search, the separator lines are moving around unexpectedly after every keystroke.

5) tapping on the page dismisses the navigation bar, but leads to a jumpy animation for both the page view and the thumbnails.

6) The page is cut off at the top and bottom when zooming in:

bildschirmfoto 2018-10-30 um 12 40 28
bildschirmfoto 2018-10-30 um 12 40 22

7) If I zoom into a page and then tap it, the zoom resets while going in/out of full screen mode. IIRC zoom mostly only resets after a double tap in other apps.

ownCloud/Viewer/PDFThumbnailsCollectionViewController.swift
mneuwert: @pablocarmu I checked once again, here chaining would sacrifice readability and in fact I only found one single other piece of code where chaining could have been applied.

ownCloud/Viewer/PDFViewerViewController.swift
mneuwert: @pablocarmu fixed

mneuwert:
@felix-schwarz Regarding your comment:
> 1. the name of the file is currently shown in the navigation bar and in an overlay. Having it in the navigation bar should be sufficient:

I was thinking that we might add more icons in the navigation bar and therefore introduced an overlay but forgot to remove the file name from navigation bar.

ownCloud/Viewer/PDFOutlineViewController.swift
```diff
@@ -0,0 +1,141 @@
...
+            self.navigationItem.rightBarButtonItem = resumeItem
+        }
+
+        Theme.shared.register(client: self, applyImmediately: true)
+
+    }
+
+    deinit {
+        Theme.shared.unregister(client: self)
+        NotificationCenter.default.removeObserver(self)
```
felix-schwarz: This might be left-over code: I can't find any prior `addObserver` calls.

ownCloud/Viewer/PDFTocTableViewCell.swift
```diff
@@ -0,0 +1,100 @@
...
+        pageLabel.rightAnchor.constraint(equalTo: self.contentView.rightAnchor, constant: -layoutMargin).isActive = true
+
+        pageLabel.leadingAnchor.constraint(equalTo: titleLabel.trailingAnchor).isActive = true
+
+        titleLabel.setContentCompressionResistancePriority(UILayoutPriority.defaultLow, for: UILayoutConstraintAxis.horizontal)
+        pageLabel.setContentCompressionResistancePriority(UILayoutPriority.defaultHigh, for: UILayoutConstraintAxis.horizontal)
+    }
+
+    // MARK: - Theme support
+
```
felix-schwarz: SwiftLint complains about a trailing white space in this line. 🤷‍♂️

ownCloud/Viewer/PDFThumbnailsCollectionViewController.swift
```diff
@@ -0,0 +1,203 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.collectionView?.prefetchDataSource = self
+
+        // Register cell classes
+        self.collectionView!.register(PDFThumbnailCollectionViewCell.self,
+                                      forCellWithReuseIdentifier: PDFThumbnailCollectionViewCell.identifier)
+
+        if self.themeCollection != nil {
+            self.collectionView!.applyThemeCollection(self.themeCollection!)
```
felix-schwarz: Ideally, this view controller should register as a client with `Theme` itself so it gets updated if the theme changes dynamically (f.ex. following sunrise/sunset in a future version) and a different `ThemeCollection` becomes the active one.

ownCloud/Viewer/PDFThumbnailsCollectionViewController.swift
```diff
@@ -0,0 +1,203 @@
...
+        }
+    }
+
+    // MARK: - Private methods
+
+    fileprivate func fetchThumbnail(for page:PDFPage, indexPath:IndexPath, size:CGSize) {
+        let blockOperation = BlockOperation()
+        weak var weakBlockOperation = blockOperation
+
+        blockOperation.addExecutionBlock { [unowned self] in
```
felix-schwarz: If the `PDFThumbnailsCollectionViewController` is dismissed and deallocated while a thumbnail is being fetched, the app will crash because the reference to self is `unowned`. Using `[weak self]` instead would account for that case.

ownCloud/Viewer/PDFThumbnailCollectionViewCell.swift
```diff
@@ -0,0 +1,73 @@
...
+    fileprivate func setupSubviews() {
+        self.backgroundColor = UIColor.lightGray
+
+        imageView = UIImageView()
+        imageView?.translatesAutoresizingMaskIntoConstraints = false
+        imageView?.contentMode = .scaleAspectFit
+        self.contentView.addSubview(imageView!)
+
+        pageLabel = UILabel()
+        pageLabel?.backgroundColor = UIColor.clear
```
felix-schwarz: Page labels are difficult to read if the page is darker. As far as I can see, Apple Books uses a white (or bright gray?) color here so the page number remains legible even on dark backdrops. Getting that "label" look of Apple Books, though, may require another container view around the label.

mneuwert:
@felix-schwarz Regarding icons, I used Google Material icons in PDF which use Apache v2 style license if I remember correctly.

ownCloud/Viewer/PDFViewerViewController.swift
```diff
@@ -20,44 +20,340 @@ import UIKit
...
+        alertController.addAction(UIAlertAction(title: "Cancel".localized, style: .cancel, handler: nil))
+
+        alertController.addTextField(configurationHandler: { textField in
+            textField.placeholder = "Page".localized
+            textField.keyboardType = .decimalPad
+        })
+
+        alertController.addAction(UIAlertAction(title: "OK".localized, style: .default, handler: { [unowned self] _ in
+            if let pageLabel = alertController.textFields?.first?.text {
+                self.selectPage(with: pageLabel)
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz right, I was even thinking about some logic: e.g. if page > max_page_count, then jump to the last page, or would you prefer to show an error message if page doesn't exist?

ownCloud/Viewer/PDFViewerViewController.swift
```diff
@@ -20,44 +20,340 @@ import UIKit
...
+            }
+        }
+    }
+
+    fileprivate func updatePageLabel() {
+        guard let pdf = pdfView.document else { return }
+
+        guard let currentPageLabel = pdfView.currentPage?.label else { return }
+
+        pageCountLabel.text = "\(currentPageLabel) of \(pdf.pageCount)"
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz good point!

ownCloud/Resources/Assets.xcassets/ic_pdf_view_continuous.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
```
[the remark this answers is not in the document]
mneuwert: Fixed and removed other unused icon as well.

ownCloud/Viewer/PDFViewerViewController.swift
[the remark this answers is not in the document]
mneuwert: @felix-schwarz however.. I think there is a problem with your suggestion to use `PDFDocument.index(for: page)` since it returns 0-based index.. so if page numbers start with something like i, ii, iii, 1, this method would return value 3 for page labeled "1".

ownCloud/Viewer/PDFSearchViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
+ * You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+ *
+ */
+
+import UIKit
+import PDFKit
+
+class PDFSearchViewController: UITableViewController, PDFDocumentDelegate, Themeable {
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz  Changed as suggested and removed `Done` alltogether

ownCloud/Viewer/PDFThumbnailsCollectionViewController.swift
mneuwert: @felix-schwarz fixed

ownCloud/Viewer/PDFTocTableViewCell.swift
mneuwert: @felix-schwarz fixed

ownCloud/Viewer/PDFOutlineViewController.swift
mneuwert: @felix-schwarz fixed

ownCloud/Viewer/PDFViewerViewController.swift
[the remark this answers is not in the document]
mneuwert: @felix-schwarz I think I tried to mage it better by differentiating between numeric page labels and once containing any other characters, but again ownCloud manual for instance does contain page with label "1" twice.. and still will confuse I guess.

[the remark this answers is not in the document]
felix-schwarz: > @felix-schwarz right, I was even thinking about some logic: e.g. if page > max_page_count, then jump to the last page, or would you prefer to show an error message if page doesn't exist?

I'd prefer an error message. macOS Preview, f.ex. just "dings" when you try to jump to a page that doesn't exist.

ownCloud/Viewer/PDFViewerViewController.swift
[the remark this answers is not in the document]
felix-schwarz: @mneuwert Good point! The first page should be page 1 of course :) … since the `y` element of the "x of y" text is using the page count, the `x` should be the page offset (+1) instead of the label to stay consistent.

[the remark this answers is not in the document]
felix-schwarz: @mneuwert I think for this "[page] of [total]" display at the bottom, labels are meaningless and prone to be inconsistent and confusing. Let's go for `PDFDocument.index(for: page) + 1`.

jesmrec:
Approved

Issue #133, opened 2018-10-25
Use displayname instead of username


jesmrec:
Latest implementation done here: https://github.com/owncloud/ios-app/issues/46 includes username in list of bookmarks. ownCloud core offers the option to set a display name, so the username is only valid in time of authentication, and the user sees his displayname.

This is appropiated in case of servers with an LDAP/AD backend, which id for the users is not talways he username, so the email or the uuid.

PR #143, opened 2018-11-05
[feature/misc] Miscellaneous additions from feature/fileprovider

felix-schwarz:
## Description
This branch contains all changes from `feature/fileprovider` – except the File Provider – including:
- Certificate management UI
- Logging settings
- Theme settings
- Beta warning
- Use of themed certificate view
- Pressing return in the URL field when adding a bookmark advances to the next step
- SDK nullability adaptions
- Auto-generated bookmark name shows username in front of server names
- Help button becomes Feedback button
- VendorServices class groups access to app version metadata and feedback options
- Unifying indentation in many places
- Extended Theme system with ThemeStyles and built-in method to switch themes with animation
- TextViewController to present license texts of components used in the app and SDK
- Fix crash for password manager integration on iPad
- Fix crash for sharing logs on iPad
- Various minor bug fixes

### Types of changes


ownCloud/Settings/SecuritySettingsSection.swift
```diff
@@ -22,244 +22,283 @@ import ownCloudSDK
...
+
+					passcodeViewController?.message = defaultMessage
+					viewController.present(passcodeViewController!, animated: true, completion: nil)
+				}
+			}
+		}, title: "Passcode lock".localized, value: isPasscodeSecurityEnabled)
+
+		// Creation of the biometrical row.
+		if let biometricalSecurityName = LAContext().supportedBiometricsAuthenticationName() {
+			// Creation of the biometrical row.
```
pablocarmu: This is probably a leftover from another pull request but we could delete the comment at this point.

ownCloud/Settings/SecuritySettingsSection.swift
```diff
@@ -22,244 +22,283 @@ import ownCloudSDK
...
-        case .thirtyMinutes:
-            return "After 30 minutes".localized
-        }
-    }
+	case always = 0
+	case oneMinute = 60
+	case fiveMinutes = 300
+	case thirtyMinutes = 1800
+
+	static let all = [always, oneMinute, fiveMinutes, thirtyMinutes]
```
pablocarmu: Maybe not on the scope of this PR but with Swift 4.2 an enum could be `CaseIterable` and this all workaround could be deleted. If we don't address this here, we should open an issue or left a `//TODO` so the linter can warn us next time.

ownCloud/Settings/Certificate Management/CertificateManagementViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+				certificateRow.certificate = certificate
+
+				uacSection.add(row: certificateRow)
+			}
+
+			self.addSection(uacSection)
+		}
+	}
+
+	override func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
```
pablocarmu: Didn't know about this feature until I've read this code! Nice thing!

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -269,26 +270,22 @@ class ClientQueryViewController: UITableViewController, Themeable {
 	}
 
 	override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
-
-		guard let rowItem : OCItem = self.items?[indexPath.row] else {
-			return
-		}
+		let rowItem : OCItem = self.items![indexPath.row]
```
pablocarmu: Maybe out of the scope of this PR but we've removed the 'optionality' of the `core` and `query` in this class, we could do the same with the `items` property. This way we could avoid all the nil checking and the `!` `?` symbols.

ownCloud/Server List/ServerListTableViewController.swift
```diff
@@ -388,3 +360,24 @@ class ServerListTableViewController: UITableViewController, Themeable {
...
 	}
 }
+
+// MARK: - OCClassSettings support
+extension OCClassSettingsIdentifier {
+	static let app = OCClassSettingsIdentifier("app")
+}
+
+extension OCClassSettingsKey {
+	static let showBetaWarning = OCClassSettingsKey("show-beta-warning")
```
pablocarmu: We can document this in the `CONFIGURATION.md` file so if any contributor/developer/QA wants to test something and this beta pop-up appears they know how to hide it. Or at least document it in `README.md` or some other place.

ownCloud/Settings/SecuritySettingsSection.swift
felix-schwarz: The next commit will remove that double comment, thanks!

ownCloud/Settings/SecuritySettingsSection.swift
felix-schwarz: Good point. The next commit will take advantage of `CaseIterable`.

ownCloud/Server List/ServerListTableViewController.swift
felix-schwarz: The next commit will amend `CONFIGURATION.md` accordingly.

ownCloud/Client/ClientQueryViewController.swift
felix-schwarz: Made that change in the latest commit.

felix-schwarz:
> I've left some comments, some of them maybe are out of scope for this but we should keep them in mind.

Thanks! Took the opportunity to address them all.

> One more thing is that we need to start thinking about how are we going to import Themes into the app, right now the light and dark color are hardcoded but in the future, we need to be able to set up those color from an MDM or a branding system.

When adding the Theme selection UI, I noticed that there was no real system for managing the Themes internally, not even the built-in ones.

To address this, I added the [`ThemeStyle`](https://github.com/owncloud/ios-app/blob/feature/misc/ownCloud/Theming/ThemeStyle.swift) class, which can encapsulate everything needed to make custom `ThemeCollection`s. ThemeStyles also include (so far untested, due to time constraints) support for [customizing single color values using dot notation](https://github.com/owncloud/ios-app/blob/69d2fc8f454d155962413d54b03385b39bd3473e/ownCloud/Theming/ThemeStyle.swift#L84).

What's still missing is an on-disk format for ThemeStyles and a way to provide custom images (f.ex. a custom logo) alongside it.

> One possible solution is through `asset catalogs`, here it's an interesting article describing how to do it [Unleashing the power of asset catalogs and bundles on iOS](https://rambo.codes/ios/2018/10/03/unleashing-the-power-of-asset-catalogs-and-bundles-on-ios.html)

I like bundles and asset catalogs a lot. Both could become components of an on-disk format.

Another option I've been thinking about is to bundle a `ThemeStyle` definition file alongside its assets in a zip file. That obviously doesn't offer asset catalog-like features, but since its just a binary blob, it could be distributed in all shapes and forms (including, possibly, via MDM).

jesmrec:
**Certificate Management UI**

### (1) 

1. Go to settings and revoke an user-accepted certificate
2. In files list, the user is warn to accept the certificate again
3. User cancel (by mistake)

The only option to get the certificate accepted again is entering again the credentials (by editing or creating again the bookmark). I will open a new issue to handle this non-blocking but improvable scenario

### (2)

with iPhone X in landscape, the section titles (`Validation Status`, `Subject`...) are overlapped with the notch:

screen shot 2018-11-08 at 09 57 05

### (3)

I would change (if feasible) the default color of `Validation Status` message. The dark blue does not fit with the dark theme:

screen shot 2018-11-08 at 10 00 01

jesmrec:
** Log feature**

some issues and stuff to deal from my side:

- What does the log level = `Off` mean? which is the difference with disabling `LogFile` ?

- Same question for `Default`? is it like `Info` level? i think `Default` is not a descriptive wording in this context.

- i'd simplify the view in this way:  one switch to enable or disable `Logging`. If it is disabled, no logs are produced and the view is empty. If the switch is enabled, view is filled with the options (log level, log output and privacy) to be set by the user.

- i guess that each logging level contains itself and the ones below. Am i right?

i will open another issue for refinements

felix-schwarz:
> **Certificate Management UI**
> 
> ### (1)
> 1. Go to settings and revoke an user-accepted certificate
> 2. In files list, the user is warn to accept the certificate again
> 3. User cancel (by mistake)
> 
> The only option to get the certificate accepted again is entering again the credentials (by editing or creating again the bookmark). I will open a new issue to handle this non-blocking but improvable scenario

What solution would you propose to better handle this scenario?

> ### (2)
> with iPhone X in landscape, the section titles (`Validation Status`, `Subject`...) are overlapped with the notch:

Fixed with latest commit.

> ### (3)
> I would change (if feasible) the default color of `Validation Status` message. The dark blue does not fit with the dark theme:

Fixed with latest commit.

jesmrec:
Other stuff in the current PR:

**Theme Settings** -> OK
**Beta warning** -> OK
**Auto-generated bookmark name shows username in front of server names** -> OK, but improvements: https://github.com/owncloud/ios-app/issues/133
**Fix crash for sharing logs on iPad**  -> OK
**Help button becomes Feedback button** -> OK
**TextViewController to present license texts of components used in the app and SDK** -> OK
**Pressing return in the URL field when adding a bookmark advances to the next step** -> OK

the remaining issues are code-related and takes part in the code review.

felix-schwarz:
> ** Log feature**
> 
> some issues and stuff to deal from my side:
> 
> * What does the log level = `Off` mean? which is the difference with disabling `LogFile` ?

That log level turns off logging completely. Long messages aren't even created in this level.

The switches control to which destinations logging does occur. By default, the app only logs to the console / `stdout`. Users wanting to capture a log file can turn that on separately.

I'm thinking about an overlay live log view (much like PiP) for the iPad version in the future. That, too, would turn up as a switch there.

> * Same question for `Default`? is it like `Info` level? i think `Default` is not a descriptive wording in this context.

Good point. I've changed that to Info with the latest commits.

> * i'd simplify the view in this way:  one switch to enable or disable `Logging`. If it is disabled, no logs are produced and the view is empty. If the switch is enabled, view is filled with the options (log level, log output and privacy) to be set by the user.

I agree this would be a better UI. Made changes accordingly in the latest commits. Also rebalanced privacy / ease-of-use:

- default log level is debug
- default destinations are both stdout and log file (so the log file output doesn't need to be enabled manually)
- logging is now off by default (so by default, nothing gets logged to the file)

> * i guess that each logging level contains itself and the ones below. Am i right?

Yes, that's right.

> i will open another issue for refinements

Ok! 👍

jesmrec:
> What solution would you propose to better handle this scenario?

I noticed that after disconnecting and connecting again, the issue view controller is again shown to the user. This is a good approach, and we can keep it.

Only as additional tips:

- Current iOS app & Android app: if certificate is not user-approved, each server request raises the pop-up with the approval dialog. A bit annoying (it is forcing the user to approve) but warns in all requests any posible mitm attack.

- In the `Certificates` section of Settings we can check only those certificates already approved. Here we can iterate and improve, including those certificates that were already shown and cancelled as `Pending for Approval` or similar, if feasible.  Users, sometimes, are frightened when some issue appears in the screen and can delay the approval till someone (system admin or whoever) checks and suggests to approve it. This is only a improvement to do in refinements issue. 

for me it is ok this first implementation of certificate security section.

jesmrec:
New UI for logging: much better!

jesmrec:
for me it is OK

PR #146, opened 2018-11-08
[feature/uploads] Upload files and pictures from the owncloud app

pablocarmu:
## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -662,42 +662,50 @@ class ClientQueryViewController: UITableViewController, Themeable {
 			style = .actionSheet
 		}
 
-		// Create the alert controller to choose from wich provider you want to upload
 		let controller = UIAlertController(title: nil, message: nil, preferredStyle: style)
 
 		let photoLibrary = UIAlertAction(title: "Upload from your photo library".localized, style: .default, handler: { (_) in
 
-			// Check if have permissons to access the library
-			PHPhotoLibrary.requestAuthorization({ newStatus in
-
-				var viewControllerToPresent: UIViewController
-
-				if newStatus ==  PHAuthorizationStatus.authorized {
-
-					// Has permissons
-					let picker = UIImagePickerController.regularImagePicker(with: .photoLibrary)
-					picker.delegate = self
-					viewControllerToPresent = picker
-				} else {
-
-					// Does not have permissons
-					let alert = UIAlertController(title: "ownCloud does not have permissons".localized, message: "We need this so that you can upload photos and videos from your photo library".localized, preferredStyle: .alert)
-
-					let settingAction = UIAlertAction(title: "Settings".localized, style: .default, handler: { _ in
-						UIApplication.shared.open(URL(string: UIApplicationOpenSettingsURLString)!, options: [:], completionHandler: nil)
-					})
-					let notNowAction = UIAlertAction(title: "Not now".localized, style: .cancel)
-
-					alert.addAction(settingAction)
-					alert.addAction(notNowAction)
-
-					viewControllerToPresent = alert
-				}
-
+			func presentImageGalleryPicker() {
+				let picker = UIImagePickerController.regularImagePicker(with: .photoLibrary)
+				picker.delegate = self
 				OnMainThread {
-					self.present(viewControllerToPresent, animated: true)
+					self.present(picker, animated: true)
 				}
-			})
+			}
+
+			let permisson = PHPhotoLibrary.authorizationStatus()
+			switch permisson {
+
+			case .authorized:
+				presentImageGalleryPicker()
+			case .notDetermined:
+				PHPhotoLibrary.requestAuthorization({ newStatus in
+					if newStatus == .authorized {
+						presentImageGalleryPicker()
+					}
+				})
+
+			default:
+				PHPhotoLibrary.requestAuthorization({ newStatus in
+
+					if newStatus == .denied {
+						let alert = UIAlertController(title: "ownCloud does not have permissons".localized, message: "We need this so that you can upload photos and videos from your photo library".localized, preferredStyle: .alert)
+
+						let settingAction = UIAlertAction(title: "Settings".localized, style: .default, handler: { _ in
+							UIApplication.shared.open(URL(string: UIApplicationOpenSettingsURLString)!, options: [:], completionHandler: nil)
+						})
+						let notNowAction = UIAlertAction(title: "Not now".localized, style: .cancel)
+
+						alert.addAction(settingAction)
+						alert.addAction(notNowAction)
+
+						OnMainThread {
+							self.present(alert, animated: true)
+						}
+					}
+				})
+			}
 		})
 
 		let locationsDirectory = UIAlertAction(title: "Upload file", style: .default) { _ in
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -291,6 +291,10 @@ class ClientQueryViewController: UITableViewController, Themeable {
 	override func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
 		let item: OCItem = itemAtIndexPath(indexPath)
 
+		guard item.isPlaceholder == false else {
+			return UISwipeActionsConfiguration(actions: [])
+		}
+
 		let deleteContextualAction: UIContextualAction = UIContextualAction(style: .destructive, title: "Delete".localized) { (_, _, actionPerformed) in
 			self.delete(item, viewDidAppearHandler: {
 				actionPerformed(false)
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -943,6 +943,11 @@ extension ClientQueryViewController: UIImagePickerControllerDelegate {
 			url = movieURL
 		}
 
+		if let imageAsset = info[UIImagePickerControllerPHAsset] as? PHAsset {
+			let resources = PHAssetResource.assetResources(for: imageAsset)
+			name = resources[0].originalFilename
+		}
+
 		if let progress = core.importFileNamed(name, at: query.rootItem, from: url, isSecurityScoped: false, options: nil, placeholderCompletionHandler: nil, resultHandler: { (error, _ core, _ item, _) in
 			if error != nil {
 				print("LOG ---> error uploading")
@@ -956,6 +961,7 @@ extension ClientQueryViewController: UIImagePickerControllerDelegate {
 		OnMainThread {
 			picker.dismiss(animated: true)
 		}
+
 	}
 }
 
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -651,6 +651,20 @@ class ClientQueryViewController: UITableViewController, Themeable {
 
 	}
 
+	func upload(itemURL: URL, name: String, viewDidAppearHandler: ClientActionVieDidAppearHandler? = nil, completionHandler: ClientActionCompletionHandler? = nil) {
+		if let progress = core.importFileNamed(name, at: query.rootItem, from: itemURL, isSecurityScoped: false, options: nil, placeholderCompletionHandler: nil, resultHandler: { [weak self](error, _ core, _ item, _) in
+			if error != nil {
+				Log.debug("Error uploading \(Log.mask(name)) file to \(Log.mask(self?.query.rootItem.path))")
+				completionHandler?(false)
+			} else {
+				Log.debug("Success uploading \(Log.mask(name)) file to \(Log.mask(self?.query.rootItem.path))")
+				completionHandler?(true)
+			}
+		}) {
+			self.progressSummarizer?.startTracking(progress: progress)
+		}
+	}
+
 	// MARK: - Navigation Bar Actions
 	@objc func uploadsBarButtonPressed(_ sender: UIBarButtonItem) {
 
@@ -704,10 +718,7 @@ class ClientQueryViewController: UITableViewController, Themeable {
 			let documentDirectory = UIDocumentPickerViewController(documentTypes: [kUTTypeData as String], in: .import)
 			documentDirectory.delegate = self
 			documentDirectory.allowsMultipleSelection = true
-
-			OnMainThread {
-				self.present(documentDirectory, animated: true)
-			}
+			self.present(documentDirectory, animated: true)
 		}
 
 		let cancelAction = UIAlertAction(title: "Cancel".localized, style: .cancel, handler: nil)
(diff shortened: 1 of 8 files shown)


ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -849,3 +935,48 @@ extension ClientQueryViewController: UITableViewDragDelegate {
...
 
 }
+
+// MARK: - UIImagePickerControllerDelegate
+extension ClientQueryViewController: UIImagePickerControllerDelegate {
+
+	func imagePickerController(_ picker: UIImagePickerController, didFinishPickingMediaWithInfo info: [String : Any]) {
+
+		var name: String = "test.jpg"
+		var url: URL = URL(string: "https://demo.owncloud.org")!
```
felix-schwarz: Both`name` and `url` should probably be optionals…

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -849,3 +935,48 @@ extension ClientQueryViewController: UITableViewDragDelegate {
...
+			name = movieURL.lastPathComponent
+			url = movieURL
+		}
+
+		if let imageAsset = info[UIImagePickerControllerPHAsset] as? PHAsset {
+			let resources = PHAssetResource.assetResources(for: imageAsset)
+			name = resources[0].originalFilename
+		}
+
+		upload(itemURL: url, name: name)
```
felix-schwarz: … and `upload(itemURL:name:)` only be called if `url` is non-nil.

ownCloud/Resources/Info.plist
```diff
@@ -29,8 +33,14 @@
 		<key>NSAllowsArbitraryLoads</key>
 		<true/>
 	</dict>
+	<key>NSAppleMusicUsageDescription</key>
+	<string>We need this so that you can upload photos and videos from your photo library</string>
```
felix-schwarz: Alternative suggestion:
This permission is needed by the ownCloud app to upload media files to your server.

ownCloud/Resources/Info.plist
```diff
@@ -29,8 +33,14 @@
...
 	<key>NSFaceIDUsageDescription</key>
 	<string>Unlock the app without entering the passcode.</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>We need this so that you can upload photos and videos from your photo library</string>
```
felix-schwarz: Alternative suggestion:
This permission is needed by the ownCloud app to upload photos and videos from your photo library.

ownCloud/Resources/Info.plist
```diff
@@ -29,8 +33,14 @@
...
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>We need this so that you can upload photos and videos from your photo library</string>
```
felix-schwarz: Alternative suggestion:
This permission is needed by the ownCloud app to upload photos and videos from your photo library.

ownCloud/Resources/Info.plist
```diff
@@ -60,7 +70,5 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
-	<key>ITSAppUsesNonExemptEncryption</key>
-	<false/>
```
felix-schwarz: Why did you remove this?

felix-schwarz:
For reasons unbeknownst to me, the dots grow large while being tapped:
screen shot 2018-11-15 at 17 09 10
screen shot 2018-11-15 at 17 08 41

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: Changed.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: Done

ownCloud/Resources/Info.plist
pablocarmu: Done.

ownCloud/Resources/Info.plist
pablocarmu: Done.

ownCloud/Resources/Info.plist
pablocarmu: Done.

ownCloud/Resources/Info.plist
pablocarmu: I think this was removed without notice in a merge/rebase, now it's in place again.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -849,3 +936,52 @@ extension ClientQueryViewController: UITableViewDragDelegate {
...
+			url = movieURL
+		}
+
+		if let imageAsset = info[UIImagePickerControllerPHAsset] as? PHAsset {
+			let resources = PHAssetResource.assetResources(for: imageAsset)
+			name = resources[0].originalFilename
+		}
+
+		guard name != nil, url != nil else {
+			return
```
felix-schwarz: This way, `picker.dismiss()` will not be called later on. Might therefore be better to just check for nil with an `if` ahead of the `upload()` call instead, but let the `picker.dismiss()` be called regardless.

ownCloud/Resources/Info.plist
```diff
@@ -29,8 +33,14 @@
 		<key>NSAllowsArbitraryLoads</key>
 		<true/>
 	</dict>
+	<key>NSAppleMusicUsageDescription</key>
+	<string>This permission is needed by the ownCloud app to upload media files to your server.</string>
```
felix-schwarz: Sorry for not thinking of this again, but mentioning "ownCloud app" might be a problem for branding.

"This permission is needed for uploading media files to your server." ?

ownCloud/Resources/Info.plist
```diff
@@ -29,8 +33,14 @@
...
 	<key>NSFaceIDUsageDescription</key>
 	<string>Unlock the app without entering the passcode.</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>This permission is needed by the ownCloud app to upload photos and videos from your photo library.</string>
```
felix-schwarz: Same here, sorry.

"This permission is needed to upload photos and videos from your photo library."?

ownCloud/Resources/Info.plist
```diff
@@ -29,8 +33,14 @@
...
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>This permission is needed by the ownCloud app to upload photos and videos from your photo library.</string>
```
felix-schwarz: And once more… sorry, again.

"This permission is needed to upload photos and videos from your photo library."?

ownCloud/Resources/en.lproj/Localizable.strings
```diff
@@ -80,6 +80,11 @@
 "Are you sure you want to delete this item from the server?" = "Are you sure you want to delete this item from the server?";
 "Deleting '%@'" = "Deleting '%@'";
 "Renaming to %@" = "Renaming to %@";
+"Upload from your photo library" = "Upload from your photo library";
+"ownCloud does not have permissons" = "ownCloud does not have permissons";
```
felix-schwarz: Little typo here. "permissons" vs. "permissions".

With branding in mind… maybe turn this into "Missing permissions"?

(also at line 700 of ClientQueryViewController.swift)

ownCloud/Resources/en.lproj/Localizable.strings
```diff
@@ -80,6 +80,11 @@
...
+"This permission is needed by the ownCloud app to upload photos and videos from your photo library." = "This permission is needed by the ownCloud app to upload photos and videos from your photo library.";
```
felix-schwarz: With branding in mind… maybe turn this into "This permission is needed to upload photos and videos from your photo library."?

(also at line 700 of ClientQueryViewController.swift)

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -645,6 +651,87 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+						OnMainThread {
+							self.present(alert, animated: true)
+						}
+					}
+				})
+			}
+		})
+
+		let locationsDirectory = UIAlertAction(title: "Upload file".localized, style: .default) { _ in
+			let documentDirectory = UIDocumentPickerViewController(documentTypes: [kUTTypeData as String], in: .import)
```
mneuwert: I think `documentDirectory` is confusing name here, since it is actually referring to a view controller instance

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -645,6 +651,87 @@ class ClientQueryViewController: UITableViewController, Themeable {
...
+
+						OnMainThread {
+							self.present(alert, animated: true)
+						}
+					}
+				})
+			}
+		})
+
+		let locationsDirectory = UIAlertAction(title: "Upload file".localized, style: .default) { _ in
```
mneuwert: `locationsDirectory` is confusing. I would rather call it `uploadFileAction`

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -145,6 +146,11 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		self.tableView.dragDelegate = self
 		self.tableView.dropDelegate = self
 		self.tableView.dragInteractionEnabled = true
+
+		let actionsBarButton: UIBarButtonItem = UIBarButtonItem(title: "● ● ●", style: .done, target: self, action: #selector(uploadsBarButtonPressed))
+		actionsBarButton.setTitleTextAttributes([.font :UIFont.systemFont(ofSize: 10)], for: .normal)
```
mneuwert: I think we should start thinking about moving fonts into Theming support.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -645,6 +651,87 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		}
 
 	}
+
+	func upload(itemURL: URL, name: String, viewDidAppearHandler: ClientActionVieDidAppearHandler? = nil, completionHandler: ClientActionCompletionHandler? = nil) {
```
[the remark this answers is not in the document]
pablocarmu: Good catch, just copy-paste from another action, I think even that we don't need this for the uploads because it's not presenting any screen in this step.

I'll remove it from the method definition.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: I'll change it to `documentPickerVC`

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: I'll change it to `uploadFileAction` then.

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: I think there is no current mechanism implemented to allow fonts with Themes but I'm with you, we should code it and move all the fonts there.

ownCloud/Client/ClientQueryViewController.swift
mneuwert: @pablocarmu I think we have put a rule into our coding guide, that abbreviations like `VC` shall not be used. Probably better to just name it `documentPickerViewController`. cc: @felix-schwarz

ownCloud/Client/ClientQueryViewController.swift
mneuwert: @pablocarmu yeah, what do you think of creating a separate issue for this topic?

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: changed

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: Sure, I'll open it so we can discuss there.

jesmrec:
Starting QA here...

jesmrec:
### (1) 

I noticed that .HEIC files uploaded from camera roll, does not show the thumbnail. Probably this is not something to fix here.

jesmrec:
### (2)

internal issue? or iOS one?

1. Select upload form Photo library
2. Select a video
3. `Choose`, in bottom bar

Current: Video starts to compress. By tapping in `Cancel` (bottom left) nothing happens
Expected: Cancel should stop the compression and return to picker. Or, `Cancel` disabled.

iOS12.0
iOS12

jesmrec:
### (3)

1. Switch device connection off
2. Upload a pic from Photo library

Current: Bottom message status does not appear
Expected: Issue shown

nov-20-2018 13-29-30

jesmrec:
### (4)

Also regarding offline:

1. Select a file to upload
2. Before the upload finishes, remove the device connection (upload stops)
3. Restablish the connection -> upload starts again

Current: file finishes to uploads but at the end an error happens:

nov-22-2018 13-15-11

Expected: Retry finishes correctly

felix-schwarz:
@jesmrec 

(2) Lacking a modern device that does re-encoding in this place, I unfortunately can't test this. But from the description, it sounds like an OS bug.

(3) + (4) need to be dealt with in the scope of offline support in the SDK. Please reference these issues in the respective issue, so I can work through them.

michaelstingl:
> (2) Lacking a modern device that does re-encoding in this place, I unfortunately can't test this. But from the description, it sounds like an OS bug.

@felix-schwarz I think re-encoding at upload is happening also with slow-mo videos (introduced with iPhone 5S). This might be related: https://github.com/owncloud/ios/issues/1040

jesmrec:
### (5) [ERROR HANDLING]

Related with the error handling improvements, probably to be added here: https://github.com/owncloud/ios-app/issues/152

upload twice the same pic of the Photo album to the same target folder

Current: Non friendly error
Expected: friendly error

jesmrec:
### (6)

also, SDK related

1. Select "Upload a file". 
2. Select a huge amount if files (>500)

Current: files not uploaded and finally, a crash. Any progress indicator
Expected: files are uploaded

jesmrec:
I noticed this behaviour:

1. Upload a big file (you will have time to do other actions meanwhile)
2. While it is being uploaded, tap on the three-dot-button on the file cell and delete it

Current: file is deleted from the list but keep being uploaded
Expected: By deleting, we can guess the user does not want the file there, so the upload could be cancelled

For a safe behaviour, and avoid weird user actions, i would disable the three-dot-button till the upload is finished. What do you think about? @pablocarmu @michaelstingl @felix-schwarz

michaelstingl:
> i would disable the three-dot-button

replace it with a spinner/progress indicator?

jesmrec:
the progress is shown in the bottom bar. I would set a cancel option in case the user selects wrongl a huge wrong file. But this would go for a different issue as improvement.

Fix of @pablocarmu works for me. I will the issue

jesmrec:
For me it is ready to go! great job!

PR #148, opened 2018-11-09
[icon_set] Added beta images in AppIcon.appiconset

ChrisEdS:
## Description
Added "beta" logo set to Assets.xcassets

## Motivation and Context
We have launched an official beta test. I think the icon should also show that this is a beta version. At the moment it causes confusion.
__

owncloud_ios_beta_icon-83 5 2x


michaelstingl:
@ChrisEdS please check definitions for Alpha and Beta:
https://en.wikipedia.org/wiki/Software_release_life_cycle#Stages_of_development

pablocarmu:
What is the status on this PR, should I review it?

ChrisEdS:
It seems we are waiting with the icons until the app is by definition a beta app.

hosy:
We can add another target to the Xcode project, which builds with different icons.

> Can we have different icons for TestFlight builds and the App Store builds? Or should we build future TestFlight builds with another App ID, so testers could install them side-by-side with the production App Store version?

michaelstingl:
> We can add another target to the Xcode project, which builds with different icons.

Yes, different targets seem to be the right way.

hosy:
> > We can add another target to the Xcode project, which builds with different icons.
> 
> Yes, different targets seem to be the right way.

Ok, I can do this.

hosy:
What we can do, if we add a new target, to set a different application identifier for the beta app. So you can install both apps side by side? We discussed this for Android. Does this make sense?

Disadvantage: Testflight build needs to be a different App as the App Store build.

michaelstingl:
> Disadvantage: Testflight build needs to be a different App as the App Store build.

After App Store release, I'd see it as an advantage, that the TestFlight build doesn't mess up your production data…

ChrisEdS:
We have "NEW" icons in this PR https://github.com/owncloud/ios-app/pull/313
Therefor I close this PR.

PR #167, opened 2018-11-26
[feature/fp-series-2] File Provider improvements

felix-schwarz:
## Description
Further improvements to the file provider.

## Types of changes


jesmrec:
After first Files App iteration, some pending stuff (improvements) and new features to be posted here:

- [ ] (1) Save new doc https://github.com/owncloud/ios-app/pull/67#issuecomment-432720003
- [ ] (2) https://github.com/owncloud/ios-app/pull/67#issuecomment-441060459
- [ ] (3) Refresh after adding second account of the same server (not easy to reproduce) https://github.com/owncloud/ios-app/pull/67#issuecomment-436975945

kopischke:
Is this the right place to report issues with the file provider / Files app integration in the current TestFlight build, or is there a preferred channel?

jesmrec:
@kopischke here is the right place

kopischke:
I have the following issues to report with the current TestFlight version of ownCloud (build 85):

1. **The system directory picker is unusable due to something I would call collapsing tremors:** whenever you expand a directory hierarchy, the expanded hierarchy almost immediately  collapses again. Difficult to describe properly, but [this video](https://cl.ly/d22037) shows it quite well. Try saving a file from an app supporting the iOS document management model to an ownCloud server, you’ll see what I mean.
2. **Files app occasionally refuses to display the contents at the root of the ownCloud server.** Bit of a Heisenbug, I am afraid to say, but very much real:
ownCloud root issues.
3. **Downloads triggered in Files app stall:** the download progress circle appears, and the ownCloud app shows a download has been requested in the UI, but the file is not downloaded in any reasonable amount of time (speaking of 15 minutes and more for 70 Kb PDFs over a high bandwidth non-mobile connection).
4. **ownCloud folders marked as Favourites disappear from the Files app sidebar after Files is purged from memory.** They are still marked as Favourites (the long press context menu offers a non-functional Unfavourite option), but are not shown in the sidebar. This is a common issue in several file providers I have used; the only ones I am aware of that do not suffer form this issue are _iCloud_, _Dropbox_ and _Working Copy_ (in the most recent release, according to release notes; Anders Borum may be open to sharing his secret sauce 😉).

felix-schwarz:
@kopischke Thanks for reporting these issues! To get a better idea on where to look, if you have some time, answers to these questions would help me a great deal:

Regarding issue 1: are the folders shown in the video frequently updated by multiple users - f.ex. while someone else modifying a file while you're trying to select a folder?

Regarding issue 2: is this to a local or a remote server?

Regarding issue 3: does this happen for all files, right away? Or only after some other operation seemingly didn't succeed? Does re-creating the bookmark solve the issue?

Regarding issue 4: Thanks for reporting! This is likely a side effect of the File Provider not yet providing a Working Set to the Files app (which expects to find all favorited items in that).

Thanks!

kopischke:
> @kopischke Thanks for reporting these issues! To get a better idea on where to look, if you have some time, answers to these questions would help me a great deal:

Yeah, sure, and sorry for the delay.

> **Regarding issue 1:** are the folders shown in the video frequently updated by multiple users - f.ex. while someone else modifying a file while you're trying to select a folder?

Shared and occasionally updated by other users, yes. Not at the time I encountered the issue, no (to my best knowledge). However, we also access the same instance via _Webdrive_ on our desktop machines; these may be polling (but should not be updating anything) rather aggressively…

> **Regarding issue 2:** is this to a local or a remote server?

Remote (large ownCloud instance managed on behalf of several German universities).

> **Regarding issue 3:** does this happen for all files, right away? Or only after some other operation seemingly didn't succeed? Does re-creating the bookmark solve the issue?

All files, right away, independently of if an start the download in Files or in the ownCloud app (the status feedback is consistent across both apps, but the net effect is stalling). Not sure what you mean by “re-creating the bookmark”.

> **Regarding issue 4:** Thanks for reporting! This is likely a side effect of the File Provider not yet providing a Working Set to the Files app (which expects to find all favorited items in that).

Can’t comment on that, but I would be glad to see the issue solved: the ownCloud instance I use is my main work cloud, so Favourites are a much needed feature.

felix-schwarz:
@kopischke Thanks for the additional information! 

> > **Regarding issue 3:** does this happen for all files, right away? Or only after some other operation seemingly didn't succeed? Does re-creating the bookmark solve the issue?
> 
> All files, right away, independently of if I start the download in Files or in the ownCloud app (the status feedback is consistent across both apps, but the net effect is stalling). Not sure what you mean by “re-creating the bookmark”.

Sorry, I was trapped in the project's internal terminology there. ;)

By "re-creating the bookmark", I meant deleting the server from the server list in the new iOS app and then creating it anew.

> Can’t comment on that, but I would be glad to see the issue solved: the ownCloud instance I use is my main work cloud, so Favourites are a much needed feature.

I agree wholeheartedly. There absolutely is the will and intention here to make this work before releasing the new app to the general public.

kopischke:
> By "re-creating the bookmark", I meant deleting the server from the server list in the new iOS app and then creating it anew.

@felix-schwarz I have “re-created the bookmark”; unluckily, I cannot test if that affects my issue 3 (no downloads can be triggered via Files.app) because issue 2 (Files.app shows an empty workspace) persists …

felix-schwarz:
Regarding `sdk-pending`: work on the File Provider is pending improvements in the SDK to make it more resilient.

jesmrec:
About the sanity check performed to Files App (208733cfce79e93e84f7bb3d75e6d663431a9e1a):

- Issues from the first PR of Files App support, mentioned above (https://github.com/owncloud/ios-app/pull/167#issuecomment-441618119): 
    - (1) : still reproducible
    - (2): file is uploaded, but no progress is shown to the user, so it is not easy to know what is going on till the upload finished. Not sure is something could be donde here because the action is done completely inside Files App. 
    - (3) is fixed 

- https://github.com/owncloud/ios-app/issues/221 is still reproducible

- Batched uploads are not too stable when the number of files is high (https://github.com/owncloud/ios-app/issues/224). Also i tried to upload a huge file together with the bunch of small files, and the app becomes stucked.

felix-schwarz:
> About the sanity check performed to Files App ([208733c](https://github.com/owncloud/ios-app/commit/208733cfce79e93e84f7bb3d75e6d663431a9e1a)):
> 
> * Issues from the first PR of Files App support, mentioned above ([#167 (comment)](https://github.com/owncloud/ios-app/pull/167#issuecomment-441618119)):
>   
>   * (1) : still reproducible

I tried with Pages and could (partially) reproduce it: the template I chose was uploaded to the ownCloud server, but when I made changes and tried to save them, I got an error "file does not exist".

The underlying issue seems to be that iOS / Pages expects that it can keep writing to the location that was returned from the File Provider's `importDocumentAtURL:` method.

Right now, the OC FileProvider returns a placeholder item here, so the location of the file is `../files/_placeholder_RANDOM_UUID/Example.pages`. Paths follow the `../files/FILE_ID/` pattern.

Once the upload is finished, the placeholder item is replaced by the actual item on the server, but this also changes the location of the file to `../files/FILE_ID_FROM_OWNLOUD/Example.pages` on the device.

If the upload finishes before the app writes changes, `../files/_placeholder_RANDOM_UUID/Example.pages` is gone and iOS / Pages brings up the "File does not exist" alert.

Possible solutions:
1) Only return from the `importDocumentAtURL:` once the upload has actually finished. Being just a two-liner, I tried this and it works, but it effectively prevents creating new documents when offline, because Pages / iOS waits for the FileProvider to return before actually open a new document for editing.

2) Preserve the placeholder location and track it along the finished item. This would add a lot of complexity to the Sync Engine going forward, though. A possibility, but not my favorite.

3) Allow the app to create its own FileID and have the server use it. This way, the placeholder location would also be the location of the uploaded item, because the FileID doesn't change. Unless this is already supported by the server, this would require adding support for such a feature to the server.

>   * (2): file is uploaded, but no progress is shown to the user, so it is not easy to know what is going on till the upload finished. Not sure is something could be done here because the action is done completely inside Files App.

I'm not sure what's causing this as the FileProvider provides the necessary information to iOS. From what I have been able to observe, progress information is most likely available when doing uploads inside the Files app itself.

>   * (3) is fixed
> * #221 is still reproducible

I have not been able to reproduce this with https://github.com/owncloud/ios-app/commit/208733cfce79e93e84f7bb3d75e6d663431a9e1a . Please send me a log so I can have a look.

> * Batched uploads are not too stable when the number of files is high (#224). Also i tried to upload a huge file together with the bunch of small files, and the app becomes stucked.

Likewise, please also send me a log for this.

> * If the server is down (not the network), app gets stucked

And another log for this, too :)

felix-schwarz:
@jesmrec I just pushed 6a5581d0f46c1bc2e6211d3386187cd4baf64aee that implements possible solution 1.

jesmrec:
Let's go for some tests with 6a5581d0f46c1bc2e6211d3386187cd4baf64aee

- Batched uploads: with a small amount of files works properly. I repeated the process with the 330 files and the log is the following (from XCode, if you need something else, tell me). Here, i noticed that te process do not start (my first report was about not finishing) 

[batchuploads.txt](https://github.com/owncloud/ios-app/files/2782249/batchuploads.txt)

- Server down: i opened a separate issue (https://github.com/owncloud/ios-app/issues/230). In any case, here it is the log. With the server down, i tried to delete a file:

[serverdown.txt](https://github.com/owncloud/ios-app/files/2782269/serverdown.txt)

felix-schwarz:
@jesmrec After discussing the issue with @DeepDiver1975, we came to the conclusion that option 3 is not possible. So we have option 1 implemented in 
https://github.com/owncloud/ios-app/commit/6a5581d0f46c1bc2e6211d3386187cd4baf64aee for now, but will eventually have to find an efficient way to do option 2.

jesmrec:
option 1 is working fine in 6a5581d.

ownCloud File Provider/FileProviderExtension.m
```diff
@@ -45,20 +45,57 @@ - (instancetype)init
 	OCAppIdentity.sharedAppIdentity.appGroupIdentifier = bundleInfoDict[@"OCAppGroupIdentifier"];
 
 	if (self = [super init]) {
-		_fileManager = [[NSFileManager alloc] init];
+		_fileManager = [NSFileManager new];
```
mneuwert: @felix-schwarz I find alloc / init more explicit but it's a matter of personal preference I guess

ownCloud/Client/Actions/MoreViewHeader.swift
```diff
@@ -100,19 +100,13 @@ class MoreViewHeader: UIView {
 			labelContainerView.bottomAnchor.constraint(lessThanOrEqualTo: self.safeAreaLayoutGuide.bottomAnchor, constant: -20)
 		])
 
-		titleLabel.attributedText = NSAttributedString(string: item.name, attributes: [NSAttributedString.Key.font: UIFont.systemFont(ofSize: 17, weight: .semibold)])
+		titleLabel.attributedText = NSAttributedString(string: item.name ?? "", attributes: [NSAttributedString.Key.font: UIFont.systemFont(ofSize: 17, weight: .semibold)])
```
mneuwert: @felix-schwarz minor complain.. magic number 17 could be replaced with constant. Would make sense if this font size is used in other places..

ownCloud File Provider/FileProviderExtension.m
felix-schwarz: Until recently, I used to be in the `alloc` / `init` camp, too. But since I started writing more and more Swift code, `alloc` / `init` reminded me of `allocWithZone:` and felt outdated. FWIW `new` has also been around since macOS 10.0. It really boils down to personal preference as you say. :)

ownCloud/Client/Actions/MoreViewHeader.swift
felix-schwarz: I totally agree. The Theme-system in the app should also provide styles like this, but this is outside the scope of the file provider. I see an opportunity to standardize these styles alongside making other `Theme` changes in `feature/branded-login` (#168).

Issue #120, opened 2018-09-27
Login feature branding


jesmrec:
As first issue for branding, we can start to think about implementing the first fields to be branded in the app... these are the ones in the login view.

First point to attend is the way in which we are going to implement this, and the favourite option is using appConfig configuration to generate an xml file, that can be also overwritten. We have an example of the file structure here: https://github.com/owncloud/ios-app/blob/master/MDM/AppConfig/specfile.xml

In the app, currently, we are storing branding values in pairs key-value defined in .plist files. The target is abstract this values using the appConfig file to be able to brand the app in this way.

Thinking in the login view, these parameters (already defined in Epic https://github.com/owncloud/ios-app/issues/2)  are a first approach to move forward branding:

- URL: set a url by default. User will not need to type it
- Edit URL: boolean: true: user can edit url, false: user can not edit the url
- Hide URL: boolean: true: URL not visible. false: URL visible
- Name: set a name by default. User will not need to type it
- Hide Name: boolean: true: Name not visible. false: Name visible
- Multiaccount: boolean: true: more than one account to be created. false: only one account
- OAuth2 fields: endpoints, client ID and secret, by default.

A couple of colours in `Add bookmark`:

- Background color of `Continue` button 
- Font color
- Cell background color 

Summarizing, three tasks:

- [ ] How to integrate AppConfig standard files in the app to brand different options
- [ ] Brand options regarding URL mentioned above
- [ ] Brand colors above

pablocarmu:
I've been taking a look at the possibilities here and I found some questions that need to be answered to cover all the possible cases:

1. We want to let the admins hide the URL to the user, but we have another text field for the bookmark name. We should hide this text field also?

    1(b). In case we don't want to hide the bookmark's name text field, what should be the default hint for this text field?. Right now we have as the hint the demo server URL.

2. In the server list or bookmark list, the cell has as title the name the user gives to the bookmark, and as detail label, we have the URL of the bookmark plus the user in some cases. I think that if the admin wants the URL hidden also wants it hidden in the server list. Is this assumption correct?  In case the answer is true, what should we code as detail label?

@michaelstingl @jesmrec @mneuwert @felix-schwarz

felix-schwarz:
@pablocarmu Regarding 1:
I think it'd make sense to add two more options:
- show/hide bookmark name field
- default bookmark name

Regarding 2:
Yes, I think that's how it should work. Regarding the detail label: I think it could then either be omitted - or the username be shown in that place.

[...]

PR #168, opened 2018-11-26
[feature/branded-login] Login branding / Static Login support

felix-schwarz:
## Description
This pull request seeks to implement an elegant login interface ready for branding and MDM.

## Related Issue
#2 #120 

## Types of changes

### Types
There are three object types of interest here:
- **Theme Style**: contains everything needed to generate a customized `ThemeCollection`, which defines the look of the UI
- **Static Login Profile**: contains everything needed to create a particular, new bookmark. Several profiles could f.ex. target different servers or cater to the needs of different departments.
- **Static Login Bundle**: contains global settings like name and logo of the organization and acts as a container for one of more *Static Login Profiles*.

### Theme Style
| Property   | Type | Meaning           |
| ---------- | ---- | ----------------- |
| `identifier` | `String` | Identifier uniquely identifying the style. |
| `lightColor` | `Color` | The light color to use in the generation of the theme colors. |
| `darkColor` | `Color` | The dark color to use in the generation of the theme colors. |
| `themeStyle` | `String` | The style to generate from the colors: dark (new app), light (new app), contrast (old iOS app) |
| `customizedColorsByPath ` | `String`:`String` dictionary | Key-value dictionary with colors to customize after automatic computation. F.ex. `tableRowColors.labelColor` = `#ff0000` to make labels in table rows red. |

### Static Login Profile
| Property   | Type | Meaning           |
| ---------- | ---- | ----------------- |
| `identifier` | `String` | Identifier uniquely identifying the static login profile. |
| `name` | `String` | Name of the login profile during setup. |
| `prompt` | `String` | Custom message to prompt the user to enter the account credentials. Presented above login/password fields for Basic Auth or above "Continue" button for OAuth2 |
| `customLogoName` | `String` | Name of the custom logo file to use for this bookmark. Placed in front of `name` in the profile picker and in front of bookmarks in the server list. |
| `bookmarkName` | `String` | The name that should be used for the bookmark that's generated from this profile. |
| `url` | `String` | The URL of the server targeted by this profile. |
| `allowedAuthenticationMethods` | `String` array | The identifiers of the authentication methods allowed for this profile. Allows to f.ex. force OAuth2, or to use Basic Auth even if OAuth2 is available. |
| `allowedCertificateFingerprints` | `String` array | Array of fingerprint hex strings of the certificate(s) that the app may connect to in the context of this bookmark. Connections that use certificates not contained in this array will be aborted. |
| `allowedPublicKeyFingerprints` | `String` array | Like `allowedCertificateFingerprints`, but targeting just the public key contained in the certificates rather than the certificate as a whole. |
| `maxBookmarkCount` | `Number` | The maximum number of bookmarks a user can generate from this profile. `0` if no limits apply. |
| `themeStyleIdentifier` | `String` | Identifier of the Theme Style to use for the UI when a bookmark generated from this profile is chosen by the user. |

### Static Login Bundle
| Property   | Type | Meaning           |
| ---------- | ---- | ----------------- |
| `organizationName` | `String` | Name of the organization to use throughout the app. Including on top of the whole static login flow. |
| `organizationLogoName` | `String` | Name of the custom logo file to use throughout the app. Including on top of the whole static login flow. |
| `organizationBackgroundName` | `String` | Name of the custom image file to use as background for the static login interface. |
| `organizationBackgroundName` | `String` | Name of the custom image file to use as background for the static login interface. |
| `loginThemeStyleIdentifier` | `String` | Identifier of the Theme Style to use for the static login UI. |
| `pickerTitle` | `String` | Custom title to be shown on top of the profile picker. |
| `pickerMessage` | `String` | Custom message to be shown on top of the profile picker.  |
| `profiles` | `Static Login Profile` array | An array of static login profiles contained in this bundle. If only one if provided, no profile picker is shown. |

## Screenshots:
If more than one Static Login Profile have been provided, the user can choose:
bildschirmfoto 2018-11-27 um 00 30 44

Basic Auth interface:
bildschirmfoto 2018-11-27 um 00 30 50

OAuth2 interface:
bildschirmfoto 2018-11-27 um 00 30 56

After setup completed:
bildschirmfoto 2018-11-27 um 16 27 12

Account list (start screen when at least one account has been set up):
 bildschirmfoto 2018-11-27 um 16 27 58

Themes can be chosen on a per-profile basis, to give the user a strong visual hint at the type of the account:
bildschirmfoto 2018-11-27 um 16 03 54
bildschirmfoto 2018-11-27 um 16 03 48


jesmrec:
I have some questions, taking in account this is WIP:

- I do not catch the necessity of different profiles. Which is the user case? different (hardcoded) URLs?
- Will fit here other user fields, like bookmark names, certificate handling... ? 
- The current implementation of the login feature will be replace totally or these mocks are only customer-oriented? so, we have to re-ellaborate the UI tests.

felix-schwarz:
> I have some questions, taking in account this is WIP:
> 
> * I do not catch the necessity of different profiles. Which is the user case? different (hardcoded) URLs?

I can't seem to find the original quote, but IIRC @michaelstingl wrote about some organizations currently using multiple branded apps to achieve something like this.

> * Will fit here other user fields, like bookmark names, certificate handling... ?

Anything we want. Already in there:
- `name` : The name appearing in the selector / title during setup
- `prompt` : The text appearing below the title during setup
- `customLogo`: a custom logo (not yet implemented, mimicked with Emojis) for the profile
- `bookmarkName`: The name the created bookmark should have
- `url`: The URL to use
- `allowedAuthenticationMethods`: allowed authentication methods
- `maxBookmarkCount`: The maximum number of bookmarks that can be created for that profile
- `themeStyleID`: Theme to use when connecting with the generated bookmark 

What I'd also like so there:
- `allowedCertificateFingerprints`: SHA-256 sums of allowed certificates
- `allowedCertificatePublicKeyFingerprints`: SHA-256 sums of allowed public keys in certificates

More in `StaticLoginProfile.swift` and `StaticLoginBundle.swift` (the latter contains the profiles).

> * The current implementation of the login feature will be replace totally or these mocks are only customer-oriented? so, we have to re-ellaborate the UI tests.

I started it solely with branding in mind, but the thought of using it as a pretty replacement of the current list certainly also has crossed my mind.

jesmrec:
sounds very interesting from a branding point of view. I had a look into it:

**Static Login Bundle**

Here they are the main options to be branded for customers (look of the app itself). Lot of branding options, but i miss an option to show/hide the URL in the welcome. Anyway, i think the most important thing here is the mechanism to add easily more stuff and not the stuff itself. We should match with current ownBrander.


**Static Login Profile**

My thoughts here:

It is needed to check the real needs of customers, but the development is very flexible, it is posible to do almost everything with all these properties. The point is that the other clients do not use these kind of profiles to distinguish branding options inside the same organization. Of course, technically is OK, from a product management... @pmaier1 maybe can share his opinion about.

Also, we can not forget that login profiles should not be mandatory. I mean: app in App Store, f .ex, should not include this kind of profiles (or including a default one?). This comparison https://github.com/owncloud/ios-app/pull/168/files#diff-9a0cbc3fe1de6f2354b8e77447fed816R209 is forcing to work with (at least) one default profile. But I also guess the current code target is checking easily how powerful the feature is. (am i wrong?). We should include in the code to be merged, the most basic option to fit everything's needs (from an environment with only one user till organizations with thousands of them).

More ideas welcome, i will also think more about it. 

CC @michaelstingl

jesmrec:
Recovering this. To sum up:

I think the basic needs are (in terms of login):

- Choice to set a default URL, choice to be editable, choice to hide it.
- Choice to set multiaccount. If multiaccount is enabled, bookmark name does not make sense, i think.
- Set as parameters all the OAuth2 stuff: client id, secret id, and endpoints for authorization and token.

Then, will come colour and icon branding.

miss i something @michaelstingl?

Apart of that, here we have all this properties: https://github.com/owncloud/ios-app/pull/168#issue-233730912, and here are my thoughts about https://github.com/owncloud/ios-app/pull/168#issuecomment-445830769. It is a great work, but needs to discuss if it covers the customer expectations or it is overenginering. Also, do not forget that it has to fit with ownBrander

Issue #142, opened 2018-10-30
File list: Pull-to-Refresh glitchy 


michaelstingl:
![screenflow](https://user-images.githubusercontent.com/214010/47748950-4a9aef00-dc8c-11e8-8e85-fd1a62e4cc4a.gif)

- White background in spinner region doesn't look good with the Classic theme
- File list "jumps back" a when pulled over trigger point
- White line stays there after Pull-to-Refresh

---

- https://github.com/owncloud/ios-app/commit/a03260ea7a92640f1df2dd2e008f07a3856e6003 on iPhone 7 Plus running iOS 12.1
- ownCloud 10.0.10

pablocarmu:
> White background in spinner region doesn't look good with the Classic theme

The problem is that the table view background color is white, and this is the same color that the color in the spinner region, here you have an example with a yellow background. 

![simulator screen shot - iphone x - 2018-11-28 at 16 55 21](https://user-images.githubusercontent.com/8929738/49164166-e0c74100-f32e-11e8-9ad5-364c9fa4d5b7.png)

To fix this we need to set the background color of the refresh controller to the navigation's bar one.

> File list "jumps back" a when pulled over trigger point

This can be fixed coding:

```Swift
self.tableView.estimatedRowHeight = 0

override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
	return 80
}
```

> White line stays there after Pull-to-Refresh

I've tried different approaches here and I end up following what it's explained in the following StackOverflow answer. It's a pretty hacky solution but it's the only one I've found that actually solves the problem.
https://stackoverflow.com/a/50670500

The final result is:

![refreshcontrollupdate](https://user-images.githubusercontent.com/8929738/49167033-de67e580-f334-11e8-8f80-cc3ca1ad5f4c.gif)

it's this what you're expecting @michaelstingl ?

hosy:
@pablocarmu 
I also found an (unsweet) solution, but maybe this will not break in a feature iOS release. You can set a UIView below of the RefreshControl, set the same height as the tableView and set the Y-Coordinate negative of the height.
But maybe as @felix-schwarz mentioned in the meeting, you can also try to debug the views with Xcode.

```
let gapCoverView = UIView(frame: CGRect(x: 0, y: -self.view.frame.size.height, width: self.view.frame.width, height: self.view.frame.size.height))
gapCoverView.backgroundColor = UIColor.yellow
[...]

PR #174, opened 2018-11-29
[fix/pull_to_refresh_glitchy] File list: Pull-to-Refresh glitchy Fix

pablocarmu:
## Description

Solves the problem listed [here](https://github.com/owncloud/ios-app/issues/142)

## Related Issue

https://github.com/owncloud/ios-app/issues/142

## Types of changes

## Checklist:


ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -321,6 +323,10 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		return actionsConfigurator
 	}
 
+	override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+		return 80
```
mneuwert: @pablocarmu please use constants and also why is it fixed size?

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -48,7 +48,7 @@ class ClientQueryViewController: UITableViewController, Themeable {
 	var initialAppearance : Bool = true
 	private var observerContextValue = 1
 	private var observerContext : UnsafeMutableRawPointer
```
mneuwert: @pablocarmu Is it necessary to configure key value observing using UnsafeMutableRawPointer? Can we avoid it? Check https://developer.apple.com/documentation/swift/cocoa_design_patterns/using_key-value_observing_in_swift

ownCloud/UI Elements/RefreshControl.swift
```diff
@@ -0,0 +1,56 @@
+//
```
mneuwert: @pablocarmu if possible we should not introduce hacks like this and subclass UI classes which are not supposed to be subclassed

ownCloud/Client/ClientQueryViewController.swift
mneuwert: @pablocarmu In commit 560917e21eafbf4511f77cb4ff0e8728835fcacd I only see that you have renamed reference to RefreshControl

ownCloud/UI Elements/RefreshControl.swift
mneuwert: @pablocarmu Did you try to fix it by setting UITableView background color eventually?

pablocarmu: Yes, that was my first approach, but it only fixes the color problem. The "blink" of the table view problem still persisted until I made the _hack_ on `UIRefreshControl`

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: Seems that I forgot to include these changes in the commit, will do in the following one. Thanks!

michaelstingl:
@pablocarmu what is the status here? shall @mneuwert take over?

hosy:
@pablocarmu please can you check, if this commit can solve your problem in this issue: https://github.com/owncloud/ios-app/pull/238/commits/bc8f0ecb26243f8bed7859a2fecd97609bc17243

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -46,7 +46,7 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		}
 	}
 	var progressSummarizer : ProgressSummarizer?
-	var queryRefreshControl: UIRefreshControl?
+	var refreshController: UIRefreshControl?
```
mneuwert: @hosy would prefer `refreshControl` as a name here

hosy: I don´t understand what you mean. The current naming is `refreshControl`.

mneuwert: @hosy it is called `refreshController`, not `control`

hosy: Ah! Thanks for your hint! Yes, you are right!

hosy: Just saw, that this it not possible, because UITableViewController has already an property called `refreshControl`: Cannot override with a stored property 'refreshControl'.

mneuwert: @hosy Ok, fine. That's an argument:) Will approve.

jesmrec:
With commit `b2c23781afce28fa1a0e8137465a4b6d5a169953` in classic theme, i see this:

mar-04-2019 08-49-31

this is the blank space reported in #142. 

In the gif above is not visible, but in a real device there is a line just above:

screen shot 2019-03-04 at 08 51 49

hosy:
It looks like, there is currently something broken. I will fix it.

hosy:
@jesmrec should be fixed now.

jesmrec:
Fixed.

PR #186, opened 2018-12-05
Exclude app ownCloudTests from lint rules

jesmrec:
### Description
SwiftLint rules are designed to correctness in app code. UI tests are not fitting such rules:

- Test names use to be longer than `func`names, so they describe the scope of the test
- Trailing whitespaces to distinguish different parts of each test.

App and tests also run on different targets.

For such reasons, excluding test folder will avoid unnecessary warnings.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,5 +1,5 @@
 excluded:
-    - ownCloudTests/EarlGrey.swift
+    - ownCloudTests
     - external
 
 function_body_length: 100


pablocarmu:
I don't think to disable the lint process for the test files is a good thing.
After a quick search inside the __SwiftLint__ documentation I found this: 
https://github.com/realm/SwiftLint#nested-configurations

Maybe we can try it and only remove the test folder from the _SwiftLint_ range in case there are no more options to go with.

jesmrec:
You mean, creating an specific `.swiftlint.yml` for `ownCloudTest` folder. I can check available rules for linting tests

Issue #182, opened 2018-12-02
Double-duplicate breaks file list


michaelstingl:
### Steps to reproduce
1. Duplicate `Image.jpg` ==> `Image copy.jpg`
2. Duplicate `Image.jpg` again ==> no additional file ??
3. Open `Image copy.jpg` ==> download doesn't start
4. Delete `Image copy.jpg` ==> `Image copy.jpg` gone, but progress bar stays forever in broken state
5. Pull-to-refresh ==> `Image copy.jpg` is back again
6. Open any other file ==> download doesn't start

**Workaround:**
Delete account and re-create again

### Expected behaviour
Double-duplicate should create another copy

### Actual behaviour
Double-duplicate breaks some features of the app

### Client
**iOS version:**
iOS 12.1 (16B92)

**ownCloud app version:**
ownCloud beta version 1.0 build 85 (https://github.com/owncloud/ios-app/commit/5df9764c4975f28c21eb7cb24eb133f67e6b7a59) 

**Device model:**
Model: iPhone XS Max

### Server configuration
**ownCloud version:**
```
$ curl https://demo.owncloud.com/status.php
{"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.0.10.4","versionstring":"10.0.10","edition":"Enterprise","productname":"ownCloud"
```

@felix-schwarz @pablocarmu

felix-schwarz:
Fixed as of e1263dae4cc2c7b458fcc281e3da5ce7be65269b.

jesmrec:
Fixed in #200

Issue #179, opened 2018-11-30
Set same behaviour downloading from "Open In" as in regular download when device is offline


jesmrec:
With no connection:

Clicking on a file to download

![nov-30-2018 12-04-19](https://user-images.githubusercontent.com/14894746/49285586-1dfb1280-f498-11e8-859e-30138dc45896.gif)

"Open In":

![nov-30-2018 12-10-56](https://user-images.githubusercontent.com/14894746/49285898-0ec89480-f499-11e8-93df-b703214e73b0.gif)

- Direct download warn the user he does not have connection
- Open In tries to download forever

Both should have the same behaviour.

michaelstingl:
After performing "Open in…" whilst being offline, progress bar stays forever in broken state. (same in https://github.com/owncloud/ios-app/issues/182)

pablocarmu:
If I understand correctly this would add a new argument (or another kind of parameter/option) to the OCCore's `download` method to download from server or retrieve the file from cache depending if there's a connection or not.

Something like: 

```Swift
// Please don't take attention to the argument name, it's a dummy one
//
// With connection
core.download(..., fromServer: true, ...)

// Without connection
core.download(..., fromServer: false, ...)
```

Is this correct?
if this is correct I totally agree

felix-schwarz:
@pablocarmu I thought of something like this

```swift
core.download(..., options: [ OCDownloadOption.returnImmediatelyIfOfflineOrUnavailable : true ], …)
```

This would behave like this:
- when online: download the item and then return it via the `completionHandler`
- when offline or in maintenance mode and …
    - … a local copy exists: return the local copy via the `completionHandler`
    - … no local copy exists: return an error to the `completionHandler`

pablocarmu:
works for me!

felix-schwarz:
The API is available as of ca11f9421d859724ea7609056fad1342fbc9b18e. Also in there I've changed DisplayViewController to take advantage of the new API to show how it can be used.

jesmrec:
An alert will show the user that he lacks connection, for this first iteration.

Issue #158, opened 2018-11-19
Maintenance mode


jesmrec:
ownCloud server has the choice to set a maintenance mode, so the admin team can deal with different actions by restricting the user access. 

App should deal with such mode correctly

To enable/disable maintenance mode:

```
occ maintenance:mode --on
occ maintenance:mode --off

```
A server in maintenance mode will return the following response against any client request:

```
<?xml version='1.0' encoding='utf-8'?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\ServiceUnavailable</s:exception>
  <s:message>System in maintenance mode.</s:message>
</d:error>
```

inside a  `503 Service Unavailable`

Currently, app is not displaying anything when the server is in maintenance mode, in login stage.
Inside the app, cached content is shown
Download content fails with a non-user-readable message
Files App should support this

felix-schwarz:
Fixed as of https://github.com/owncloud/ios-sdk/pull/35/commits/9b9ca60007dbc77ec11a8a2f8b500b3767aed4a8 and https://github.com/owncloud/ios-app/pull/167/commits/d4ce438f384a26596aad99661fd787cfe5281807:

- While adding bookmarks:
![bildschirmfoto 2018-12-06 um 16 47 12](https://user-images.githubusercontent.com/639669/49594653-aab13f00-f976-11e8-8313-3273a97a9b28.png)

- Logging in to an existing bookmark:
![bildschirmfoto 2018-12-06 um 16 47 19](https://user-images.githubusercontent.com/639669/49594654-aab13f00-f976-11e8-9a6f-63ae6391465d.png)

- While connected:
![bildschirmfoto 2018-12-06 um 16 47 22](https://user-images.githubusercontent.com/639669/49594655-ab49d580-f976-11e8-84d0-9ff44a169b85.png)

jesmrec:
Some comments regarding operations in maintenance mode:

- i noticed that several operations are enqueued (like in offline mode): rename, duplicate, move, to be pushed once maintenance mode is off. Other ones are directly forbidden-503 like delete, create folder and upload. We should unify criteria and make all run with the same behaviour. 

- `Move` operation under maintenance mode, shows the bottom status bar empty

![dec-10-2018 12-44-41](https://user-images.githubusercontent.com/14894746/49730681-6923db00-fc79-11e8-839e-f7e9eb1366c3.gif)

[...]

PR #200, opened 2019-01-09
[fix/sdkupdates] SDK updates, improved maintenance mode support and bugfixes

felix-schwarz:
## Description
This updates the app for the latest changes in the SDK (incl. the updated Sync Engine), improves support for maintenance mode and fixes various bugs.

## Related Issue
#158, #179 and #182

## Test Focus
A focus of the tests should be on the issues referenced above as well as on the FileProvider.

## Types of changes


felix-schwarz:
@jesmrec Regarding 1: Background support is not yet complete.

Regarding 2: chances are you killed the app while it was processing a response. The version of the SDK used in `fix/sdkupdates` is not yet immune to this.

The more recent, current version of the SDK in the `develop`-branch offers protection against this by storing responses in the database first before processing them.

The final piece in the crash-/kill-protection puzzle will be when the Sync Engine can track HTTP requests and thereby diagnose if a HTTP request is no longer running, but the response is also missing (in case the app was terminated while receiving the response) - and react accordingly.

Issue #202, opened 2019-01-11
[TX] Transifex job is pushing more languages that we supported


javiergonzper:
Transifex is pushing to our project several languages that are not supported yet like:

- cs_CZ.lproj
- de.lproj
- de_DE.lproj
- en_GB.lproj
- ko.lproj
- mk.lproj
- nb_NO.lproj
- nn_NO.lproj
- pt_BR.lproj
- pt_PT.lproj
- ru.lproj
- sq.lproj
- th_TH.lproj
- zh_CN.lproj

Right now we only support English and Spanish. And Spanish is not being updating.

AC:
- Support all the languages that TX Job is pushing to GitHub
- ~Add Spanish to the TX Job~

michaelstingl:
Tx pushes all languages that have a minimal percentage. This value is defined in the drone job

javiergonzper:
> Tx pushes all languages that have a minimal percentage. This value is defined in the drone job

Ok. Then the second AC is not necessary will be automatically pushed when it have enough percentage of translation.

PR #231, opened 2019-01-22
Added several languages

javiergonzper:
## Description
Added the following languages:

- cs_CZ
- de 
- de_DE 
- en_GB 
- ko 
- mk
- nb_NO
- nn_NO
- pt_BR
- pt_PT
- ru
- sq
- th_TH
- zh_CN

There are only changes on:
- Project (supported new languages)
- .tx/config to make the mapping from Transifex to Xcode

## Related Issue
https://github.com/owncloud/ios-app/issues/202

## How Has This Been Tested?
Once it has been merged to master we will check if the languages are updated correctly

## Types of changes


michaelstingl:
@javiergonzper check the conflicts, then merge…

javiergonzper:
> @javiergonzper you only prepared the files for future translation and translations are meant to be added later?

Yes. Translations are automatically pushed to the repository by Drone (on Master branch). But I discover a problem here:
- Drone push strings files when the files are over 70% of translation. The problem here is that Drone do not calculate the 70% taking into account all the files together and we use 2 files on the app and 2 files on the SDK.
The good point is that we use English strings to be translated instead of "keys". So even if Drone do not push some file because it is not over 70% the UI do not breaks with strange keys.

Issue #223, opened 2019-01-16
[UI] Refresh Serverlist View


hosy:
**Reorder/Remove Buttons**

The `+` Button should be removed if the view is not in Edit mode. It should be only visible, if the user sets the view in Edit mode. I think this button is not so often needed and so we should not show rarely needed elements.

The `Edit` button should be on the right side. This can also be seen in the Files App and this does better match the UI guidelines.

**Large Title, Header**

The title `ownCloud` should be displayed as large title. This is a UI concept which was introduced in iOS 11 and is a better welcome for the user.

Set a header title `Locations` above the server list. This is analog to the Files app. Wording can be changed, to match more the ownCloud wording.

**Select Server from List:**

If a Server is selected by the user, the File list should not be shown as modal transition. A better behavoir would be to use a Push transition on the `UINavigationViewController`.

Account list today | Account list in edit mode  | draft&nbsp;of&nbsp;the&nbsp;suggestions&nbsp;above&nbsp;&nbsp;&nbsp;
-- | -- | --
![2019-01-16 19 07 07](https://user-images.githubusercontent.com/214010/51269189-1c42cb00-19c2-11e9-9756-0023afeb79fc.png) | ![2019-01-16 19 07 20](https://user-images.githubusercontent.com/214010/51269190-1c42cb00-19c2-11e9-8d0f-bbb65abd0e83.png) | ![serverlist draft](https://user-images.githubusercontent.com/736109/51238178-9cdbda00-1976-11e9-8e95-55c7ba63362a.png)
(Build 96) | (Build 96) |

jesmrec:
>The + Button should be removed if the view is not in Edit mode. It should be only visible, if the user sets the view in Edit mode. I think this button is not so often needed and so we should not show rarely needed elements.

I think that the '+' button is needed to add new bookmarks to the list, so, it is not needed in `Edit` mode. But it you are not in `Edit`, we'd have to think in a different option to create bookmarks.

> Set a header title Locations above the server list. This is analog to the Files app. Wording can be changed, to match more the ownCloud wording.

i think ' Bookmarks` or `Servers` fits better than `Locations` inside the ownCloud words.

@wuenschedesign opened this related one: https://github.com/owncloud/ios-app/issues/210

michaelstingl:
What about `Accounts`? I don't like `Bookmarks` and `Servers` is also not perfect, because you can have multiple accounts on one ownCloud server.

michaelstingl:
[...]

PR #238, opened 2019-01-28
#223 [UI] Refresh Serverlist View

hosy:
## Description
Changed logic if an item is selected from server list: removed modal logic and implemented to push the ClientRootViewController. Because we using there a UITabBarController, we have to hide the previous UINavigationController navigationBar and to some tricks to show the Back button.

Implemented a Header View for the server list and now using large titles.

## Related Issue
#223 

## Motivation and Context
Better UX and UI

## Types of changes

## Checklist:

______

BUGS & IMPROVEMENTS

- [ ] Title `ownCloud` with the big font fits OK for short titles. With long titles (long app names), will look something like that: https://github.com/owncloud/ios-app/pull/238#issuecomment-461408474
- [x] Glitch  when pulling down https://github.com/owncloud/ios-app/pull/238#issuecomment-463833594
- [x] Change wording:  `Locations` -> `Accounts` https://github.com/owncloud/ios-app/pull/238#issuecomment-463833594
- [x] Remove from file list https://github.com/owncloud/ios-app/pull/238#issuecomment-463833594


ownCloud/Client/ClientRootViewController.swift
```diff
@@ -19,10 +19,15 @@
 import UIKit
 import ownCloudSDK
 
-class ClientRootViewController: UITabBarController {
+class EmptyViewController : UIViewController {
```
mneuwert: @hosy why not just use UIViewController directly?

ownCloud/Client/ClientRootViewController.swift
```diff
@@ -139,10 +144,16 @@ class ClientRootViewController: UITabBarController {
...
+        self.view.backgroundColor = Theme.shared.activeCollection.tableBackgroundColor
+        
+        self.navigationController?.setNavigationBarHidden(true, animated: true)
+        
+        filesNavigationController = ThemeNavigationController()
+        filesNavigationController?.delegate = self
+        filesNavigationController?.navigationBar.isTranslucent = false
+        filesNavigationController?.navigationBar.prefersLargeTitles = true
+        filesNavigationController?.tabBarItem.title = "Browse".localized
+        filesNavigationController?.tabBarItem.image = Theme.shared.image(for: "folder", size: CGSize(width: 25, height: 25))
```
mneuwert: @hosy can we change 25 to a constant?

ownCloud/Client/ClientRootViewController.swift
```diff
@@ -155,32 +166,28 @@ class ClientRootViewController: UITabBarController {
...
-
-			queryViewController.navigationItem.leftBarButtonItem = self.logoutBarButtonItem()
-
-			self.filesNavigationController?.pushViewController(queryViewController, animated: false)
+            self.filesNavigationController?.setViewControllers([self.emptyViewController, queryViewController], animated: false)
 		}
 	}
+    
+    func navigationController(_: UINavigationController, willShow: UIViewController, animated: Bool) {
+        if willShow.isEqual(emptyViewController) {
```
mneuwert: @hosy I would at least explain in a comment what emptyViewController is for to make it less confusing

ownCloud/Server List/ServerListTableHeaderView.swift
```diff
@@ -0,0 +1,81 @@
...
+        
+        NSLayoutConstraint.activate([
+            
+            shadowView.bottomAnchor.constraint(equalTo: self.bottomAnchor),
+            shadowView.widthAnchor.constraint(equalTo: self.widthAnchor),
+            shadowView.leadingAnchor.constraint(equalTo: self.leadingAnchor),
+            shadowView.trailingAnchor.constraint(equalTo: self.trailingAnchor),
+            shadowView.heightAnchor.constraint(equalToConstant: 1.0),
+            
+            textLabel.topAnchor.constraint(equalTo: self.topAnchor, constant: 10.0),
```
mneuwert: @hosy few magic numbers..

ownCloud/Server List/ServerListTableHeaderView.swift
```diff
@@ -0,0 +1,81 @@
...
+        Theme.shared.register(client: self, applyImmediately: true)
+        
+        textLabel.translatesAutoresizingMaskIntoConstraints = false
+        textLabel.text = "Locations".localized
+        textLabel.font = UIFont.boldSystemFont(ofSize: 24.0)
+        
+        self.addSubview(textLabel)
+        
+        let shadowView = UIView()
+        shadowView.backgroundColor = UIColor.init(white: 0.0, alpha: 0.1)
```
mneuwert: @hosy This is jut one pixel line at the bottom? Just was thinking if it can use a color which is the result of blending the background color of the parent view with 10% opaque black to avoid compositing all-together.

ownCloud/Server List/ServerListTableHeaderView.swift
```diff
@@ -0,0 +1,81 @@
...
+    
+    override init(frame: CGRect) {
+        
+        super.init(frame: frame)
+        
+        Theme.shared.register(client: self, applyImmediately: true)
+        
+        textLabel.translatesAutoresizingMaskIntoConstraints = false
+        textLabel.text = "Locations".localized
+        textLabel.font = UIFont.boldSystemFont(ofSize: 24.0)
```
mneuwert: @hosy can we use dynamic-type here? Seen in another PR that we started using it.

ownCloud/Client/ClientRootViewController.swift
```diff
@@ -139,10 +144,16 @@ class ClientRootViewController: UITabBarController {
 	override func viewDidLoad() {
 		super.viewDidLoad()
 
-		filesNavigationController = ThemeNavigationController()
-		filesNavigationController?.navigationBar.isTranslucent = false
-		filesNavigationController?.tabBarItem.title = "Browse".localized
-		filesNavigationController?.tabBarItem.image = Theme.shared.image(for: "folder", size: CGSize(width: 25, height: 25))
+        self.view.backgroundColor = Theme.shared.activeCollection.tableBackgroundColor
+        
+        self.navigationController?.setNavigationBarHidden(true, animated: true)
```
mneuwert: @hosy do you intend `animated: true` here in `viewDidLoad`?

ownCloud/Client/ClientRootViewController.swift
```diff
@@ -155,32 +166,28 @@ class ClientRootViewController: UITabBarController {
...
-
-	func logoutBarButtonItem() -> UIBarButtonItem {
-		let barButton = UIBarButtonItem(title: "Disconnect".localized, style: .plain, target: self, action: #selector(logout(_:)))
-		barButton.accessibilityIdentifier = "disconnect-button"
-		return barButton
-	}
-
-	@objc func logout(_: Any) {
-		self.closeClient()
+        self.viewControllers = [filesNavigationController] as? [UIViewController]
```
mneuwert: @hosy is it needed as `UINavigationController` inherits from `UIViewController`?

ownCloud/Client/ClientRootViewController.swift
```diff
@@ -155,32 +166,28 @@ class ClientRootViewController: UITabBarController {
...
 	}
 
 	func closeClient(completion: (() -> Void)? = nil) {
-		self.presentingViewController?.dismiss(animated: true, completion: completion)
+        self.navigationController?.setNavigationBarHidden(false, animated: false)
+        self.navigationController?.popViewController(animated: true, completion: {
```
mneuwert: @hosy is `completion` lambda not optional here?

ownCloud/Client/ClientRootViewController.swift
```diff
@@ -155,32 +166,28 @@ class ClientRootViewController: UITabBarController {
...
 	}
 
 	func coreReady() {
 		OnMainThread {
 			let queryViewController = ClientQueryViewController(core: self.core!, query: OCQuery(forPath: "/"))
-
-			queryViewController.navigationItem.leftBarButtonItem = self.logoutBarButtonItem()
-
-			self.filesNavigationController?.pushViewController(queryViewController, animated: false)
+            self.filesNavigationController?.setViewControllers([self.emptyViewController, queryViewController], animated: false)
```
mneuwert: @hosy would at least explain in a comment, why you put emptyViewController onto navigation stack.

ownCloud/Client/ClientRootViewController.swift
hosy: Changed my indent behavior in Xcode from Spaces to Tabs no.

ownCloud/Client/ClientRootViewController.swift
hosy: Yes, because there is a compiler error, if I remove `as? [UIViewController]`

ownCloud/Client/ClientRootViewController.swift
hosy: That's right!

ownCloud/Client/ClientRootViewController.swift
hosy: Because we have nested UINavigationControllers (first one from ServerListTableViewController and each item UITabBarController needs it own UINavigationController), we have to fake the UINavigationController logic. Here we insert the emptyViewController, because in the UI should appear a "Back" button if the root of the queryViewController is shown. Therefore we put at first the emptyViewController inside and at the same time the queryViewController. Now, the back button is shown and if the users push the "Back" button the ServerListTableViewController is shown. This logic can be found in navigationController(_: UINavigationController, willShow: UIViewController, animated: Bool) below.

ownCloud/Client/ClientRootViewController.swift
hosy: if the emptyViewController will show, because the push button in ClientQueryViewController was triggered, push immediately to the ServerListTableViewController, because emptyViewController is only a helper for showing the "Back" button in ClientQueryViewController

ownCloud/Server List/ServerListTableHeaderView.swift
hosy: Will be handled in a separate issue

ownCloud/Server List/ServerListTableHeaderView.swift
hosy: I don´t understand what you mean. This should looks like the shadow line of a bar from UINavigationBarController.

ownCloud/Server List/ServerListTableHeaderView.swift
hosy: Fixed

ownCloud/Client/ClientRootViewController.swift
mneuwert: @hosy I am fine with that approach, I just would like to have a comment in the code explaining why an empty view controller is used

ownCloud/Server List/ServerListTableHeaderView.swift
mneuwert: @hosy I was just thinking that using solid color in the correct shade would make rendering faster. But here the benefit is probably neglectable.

ownCloud/Client/ClientRootViewController.swift
mneuwert: @hosy ok, fine

ownCloud/Client/ClientRootViewController.swift
mneuwert: @hosy as already pointed out, I am fine with the approach, I would just like a short comment in the code

hosy: I added the comment above also to the code

ownCloud/Client/ClientRootViewController.swift
hosy: I added the comment above also to the code

jesmrec:
Code approved, let's move to QA

jesmrec:
Change of orientation changes font size.

1. Initially:

screen shot 2019-02-05 at 12 48 33

2. Change to landscape: font size is reduced

screen shot 2019-02-05 at 12 49 10

3. Change again to portrait:

screen shot 2019-02-05 at 12 50 14

font sizes (owncloud and Locations) are not the same as in 1.

jesmrec:
A couple of thoughts:

1. `ownCloud` word in the title is prepared to be branded. isn't it?

2. `Locations` word should be hidden in case multiaccount is not selected. This is kind of new branding option. Is this also prepared?

hosy:
> A couple of thoughts:
> 
> 1. `ownCloud` word in the title is prepared to be branded. isn't it?
Fixed in #259 

> 2. `Locations` word should be hidden in case multiaccount is not selected. This is kind of new branding option. Is this also prepared?
 I don’t think it should be hidden, because it is also a description if you only have one account.

hosy:
> Change of orientation changes font size.
> 
> 1. Initially:
> 
> screen shot 2019-02-05 at 12 48 33
> 
> 1. Change to landscape: font size is reduced
> 
> screen shot 2019-02-05 at 12 49 10
> 
> 1. Change again to portrait:
> 
> screen shot 2019-02-05 at 12 50 14
> 
> font sizes (owncloud and Locations) are not the same as in 1.

Looks like in landscape mode is the left padding not correct for Locations label. Reduced font size is correct, because this is the behaviour for a large title in this mode.

jesmrec:
ok.

i guess that different size of `ownCloud` label on the top while changing orientation is because the view needs a small scroll.

jesmrec:
Could you update with `master`? so we can check if the tests pass

jesmrec:
### (2)

1. Delete all bookmarks you have in your list

Current: ownCloud and Locations remain in the welcome view
Expected: welcome view clean:

feb-06-2019 08-47-13

jesmrec:
In the current branch, `testShowFileList` and `testShowFileListWithItems` are failing because the mocked list of flies is not being leaded.

jesmrec:
I was wondering if this is a good design view: when the bookmark does not have name (or a long name), you will see the file list view like:

screen shot 2019-02-07 at 08 36 10

for me it is a bit weird such big title, what do you think about? @michaelstingl, @wuenschedesign, @pablocarmu @javiergonzper @mneuwert @felix-schwarz

michaelstingl:
Usernames can become very, very long…

`1stname.lastname@companyname.tld@owncloudinstance.tld`

hosy:
I think we should include it in the master and everybody should have an own look. Large Titles is what Apple recommends to use.
If we do not use large titles, the UI looks a little bit _easy_ or like _Hello World_.

hosy:
Another solution could be to remove the large title in the file list and set it only on the server list.

michaelstingl:
I need to build it again and play with it a bit…

hosy:
@michaelstingl all checks passed. I could merge it into the master!?

michaelstingl:
> @michaelstingl all checks passed. I could merge it into the master!?

Didn't check it yet.

jesmrec:
Locations does not appear anymore in the welcome view, but the name of the app: here is the difference with `master`

untitled

This is not a bad behaviour, just for you to compare

@hosy will merge when passes from QA or all the team agrees with the suggested solution. We are close to that.

jesmrec:
> Another solution could be to remove the large title in the file list and set it only on the server list.

i'd go for this solution.

hosy:
> Locations does not appear anymore in the welcome view, but the title name of the app: here is the difference with `master`
> 
> untitled
> 
> This is not a bad behaviour, just for you to compare
> 
> @hosy will merge when passes from QA or all the team agrees with the suggested solution. We are close to that.

Large Title will stay, if no server is in the list. This does not depend on the table header.

jesmrec:
> Large Title will stay, if no server is in the list. This does not depend on the table header.

i know, the title is there like the screenshot at the right.

The only objection here is that the big title (screenshot on the left) is OK for `ownCloud`. When the title is `My nice company GmbH` or a longer app/company name, this can lead to the same effect as the list view

jesmrec:
i mean, something like that:

screen shot 2019-02-07 at 13 56 11

maybe this kind of titles are more suitable for static wording or similar.

michaelstingl:
Okay, had a closer look today. I think the server list is a fresh approach and nice for the first release. 👍 

I see some glitches on the file list changes and wouldn't spend time on this for the first release. I think the top-section needs more love in one of the next sprints.

hosy:
With file list changes you mean no push on navigation controller? Keep modal file list animation?

michaelstingl:
> With file list changes you mean no push on navigation controller? Keep modal file list animation?

@hosy I clarified in the 2 previous comments https://github.com/owncloud/ios-app/pull/238#issuecomment-463833594 and https://github.com/owncloud/ios-app/pull/238#pullrequestreview-204012536 . Let me know if you have any questions.

jesmrec:
- Glitch in pulling down does not happen anymore
- Wording change to `Accounts` is done
- Feature removed from file list

For me it is OK to go, without forgetting the branding issues.

hosy:
@michaelstingl did you test the new build and can we go on?

Issue #196, opened 2019-01-04
[UX] Improve accessibility


michaelstingl:
I played a bit with [VoiceOver](https://www.apple.com/accessibility/iphone/vision/):

![2019-01-04 09 15 22](https://user-images.githubusercontent.com/214010/50681358-84310300-100b-11e9-9133-ee176888837e.jpg)

I think we can improve in different places. Apple provides great resources for this topic:

- https://developer.apple.com/accessibility/ios/
- https://developer.apple.com/documentation/uikit/accessibility/
- https://developer.apple.com/videos/frameworks/accessibility/

@pablocarmu Would you like work on this a few days in the next milestone? (to achieve some quick wins)

### Client
**iOS version:**
iOS 12.1.2 (16C101)

**ownCloud app version:**
ownCloud beta version 1.0 build 85 (https://github.com/owncloud/ios-app/commit/5df9764c4975f28c21eb7cb24eb133f67e6b7a59) 

**Device model:**
Model: iPhone XS Max

felix-schwarz:
@pablocarmu @michaelstingl The relevant APIs here are `accessibilityLabel` and `accessibilityHint`, which can be used to add VoiceOver specific descriptions to any UI element:
- https://developer.apple.com/documentation/objectivec/nsobject/1615181-accessibilitylabel
- https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint

This f.ex. should be all it takes to fix the issue for the create folder button:

```swift
createFolderButton.accessibilityLabel = "Create folder"
```

The greater difficulty here is to follow the guidelines provided for those two APIs to provide the "right" string.

There also is Apple's Accessibility Inspector app, which can perform an automatic audit of the app's currently visible UI and identify and report common issues.

pablocarmu:
I guess that we could localize this accessibility hint the same way that other strings using our `.localize` API.

We can start adding this accessibility hint to the current features. Should we create new `.string`filet to write down the hints and localize them?

felix-schwarz:
@pablocarmu I think we can use the existing `.strings` files as there will be some overlap between UI and accessibility hint strings, which would then have to localize twice.

michaelstingl:
In the iOS Files app for example, the "New folder" button says "New folder, button" with VoiceOver.

PR #239, opened 2019-01-29
Accessibility improvements

pablocarmu:
## Description

Improves the Voice Over suggestions and adds support for dynamic font sizes.

## Related Issue

https://github.com/owncloud/ios-app/issues/196

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -147,6 +147,9 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		self.tableView.dragInteractionEnabled = true
 
 		let actionsBarButton: UIBarButtonItem = UIBarButtonItem(barButtonSystemItem: .add, target: self, action: #selector(uploadsBarButtonPressed))
+		actionsBarButton.accessibilityLabel = "Upload files".localized
+		actionsBarButton.setTitleTextAttributes([.font :UIFont.systemFont(ofSize: 10)], for: .normal)
+		actionsBarButton.setTitleTextAttributes([.font :UIFont.systemFont(ofSize: 10)], for: .highlighted)
 		self.navigationItem.rightBarButtonItem = actionsBarButton
 	}
 
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -148,8 +148,6 @@ class ClientQueryViewController: UITableViewController, Themeable {
 
 		let actionsBarButton: UIBarButtonItem = UIBarButtonItem(barButtonSystemItem: .add, target: self, action: #selector(uploadsBarButtonPressed))
 		actionsBarButton.accessibilityLabel = "Upload files".localized
-		actionsBarButton.setTitleTextAttributes([.font :UIFont.systemFont(ofSize: 10)], for: .normal)
-		actionsBarButton.setTitleTextAttributes([.font :UIFont.systemFont(ofSize: 10)], for: .highlighted)
 		self.navigationItem.rightBarButtonItem = actionsBarButton
 	}
 
--- a/ownCloud/Client/SortBar.swift
+++ b/ownCloud/Client/SortBar.swift
@@ -32,6 +32,9 @@ class SortBar: UIView, Themeable {
 
 	weak var delegate: SortBarDelegate?
 
+	// MARK: - Constants
+	let sideButtonsSize: CGSize = CGSize(width: 30.0, height: 30.0)
+
 	// MARK: - Instance variables.
 
 	var stackView: UIStackView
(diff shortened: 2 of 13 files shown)


ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -246,6 +254,15 @@ class DisplayViewController: UIViewController {
 		// This function is intended to be overwritten by the subclases to implement a custom view based on the source property.s
 	}
 
+	func hideUIElements() {
```
mneuwert: @pablocarmu Can we use a better method name here? It is not hiding arbitrary UI elements, I think all these elements are related to certain context, right?

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -159,7 +164,8 @@ class DisplayViewController: UIViewController {
 
 			metadataInfoLabel!.centerXAnchor.constraint(equalTo: iconImageView.centerXAnchor),
 			metadataInfoLabel!.topAnchor.constraint(equalTo: iconImageView!.bottomAnchor, constant: 10),
-			metadataInfoLabel!.widthAnchor.constraint(equalTo: iconImageView.widthAnchor),
+			metadataInfoLabel!.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor, constant: 10),
```
mneuwert: @pablocarmu may be use let constant instead of plain value 10?

ownCloud/UI Elements/StaticTableViewRow.swift
```diff
@@ -151,7 +151,7 @@ class StaticTableViewRow : NSObject, UITextFieldDelegate {
 	// MARK: - Text Field
 	public var textField : UITextField?
 
-	convenience init(textFieldWithAction action: StaticTableViewRowAction?, placeholder placeholderString: String = "", value textValue: String = "", secureTextEntry : Bool = false, keyboardType: UIKeyboardType = UIKeyboardType.default, autocorrectionType: UITextAutocorrectionType = UITextAutocorrectionType.default, autocapitalizationType: UITextAutocapitalizationType = UITextAutocapitalizationType.none, enablesReturnKeyAutomatically: Bool = true, returnKeyType : UIReturnKeyType = UIReturnKeyType.default, identifier : String? = nil) {
+	convenience init(textFieldWithAction action: StaticTableViewRowAction?, placeholder placeholderString: String = "", value textValue: String = "", secureTextEntry : Bool = false, keyboardType: UIKeyboardType = UIKeyboardType.default, autocorrectionType: UITextAutocorrectionType = UITextAutocorrectionType.default, autocapitalizationType: UITextAutocapitalizationType = UITextAutocapitalizationType.none, enablesReturnKeyAutomatically: Bool = true, returnKeyType : UIReturnKeyType = UIReturnKeyType.default, identifier : String? = nil, accessibilityLabel: String? = nil) {
```
mneuwert: @pablocarmu add some line breaks for better readability?

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -147,6 +147,9 @@ class ClientQueryViewController: UITableViewController, Themeable {
 		self.tableView.dragInteractionEnabled = true
 
 		let actionsBarButton: UIBarButtonItem = UIBarButtonItem(barButtonSystemItem: .add, target: self, action: #selector(uploadsBarButtonPressed))
+		actionsBarButton.accessibilityLabel = "Upload files".localized
+		actionsBarButton.setTitleTextAttributes([.font :UIFont.systemFont(ofSize: 10)], for: .normal)
```
mneuwert: @pablocarmu Couldn't we use` UIFont.preferredFont(forTextStyle:)` here?

ownCloud/Client/SortBar.swift
```diff
@@ -73,6 +74,7 @@ class SortBar: UIView, Themeable {
 		sortButton.accessibilityIdentifier = "sort-bar.sortButton"
 
 		leftButtonImage = Theme.shared.image(for: "folder-create", size: CGSize(width: 30.0, height: 30.0))!.withRenderingMode(.alwaysTemplate)
```
mneuwert: @pablocarmu Constant instead of plain value 30?

ownCloud/Client/ClientQueryViewController.swift
pablocarmu: After re-reading the Apple's accessibility guidelines again I've noticed that the navigation/toolbar items don't scale their font size like regular labels. Instead, you can provide a PDF vector inside an asset catalog and, when you long press on these elements, an alert-style description is provided (vector image + bigger message).

I think we can make this in the second iteration of this feature because we need to collect all the SVGs we're using and convert them to PDFs

ownCloud/UI Elements/StaticTableViewRow.swift
pablocarmu: I'll do it but also I need to modify the `swiftlint.yml` file because there is a bug in *SwiftLint* when the project/file use Tabs instead of Spaces.

need to add `vertical_parameter_alignment_on_call` to the excluded rules
Here it's the *SwiftLint* issue https://github.com/realm/SwiftLint/issues/1573

ownCloud/Client/SortBar.swift
pablocarmu: Sure, I'll do it!

ownCloud/Viewer/DisplayViewController.swift
pablocarmu: Sure, I'll do it!

ownCloud/Viewer/DisplayViewController.swift
```diff
@@ -41,6 +41,12 @@ protocol DisplayViewEditingDelegate: class {
 class DisplayViewController: UIViewController {
 
 	private let iconImageSize: CGSize = CGSize(width: 200.0, height: 200.0)
+	private let iconImageViewCenterYConstraintConstant: CGFloat = -60.0
```
mneuwert: @pablocarmu I am not convinced by creating constants just for the sake of creating constants. For example, i would ask a question in an expression like this `cancelButton!.topAnchor.constraint(equalTo: progressView!.bottomAnchor, constant: cancelButtonTopConstraintConstant`: what is `cancelButtonTopConstraintConstant`? Is it spacing between UI elements? Is it some margin etc.? I mean I would expect something more generic.

Issue #222, opened 2019-01-16
[UI] Refresh Passcode UI


hosy:
The passcode UI should get a more iOS like UI.

The current passcode has in my opinion some problems:

* Square Button not really iOS like
* Font size to small, not so easy to read
* Button Background Color to dark

I have two design suggestions which can you see below.

* Increased the font size
* Round Button, also used in iOS for passcodes

I added two different designs, please vote which one to you prefer or give suggestions, what we can do better.

Important for the development should be, to test it with:

* External keyboard connected
* Accessibility enabled (use accessibility label)
* Check iPad UI (use `viewController.modalPresentationStyle = UIModalPresentationFormSheet;`)

Draft 1: | Draft 2:
-- | --
![passcode draft 1](https://user-images.githubusercontent.com/736109/51236502-93507300-1972-11e9-8727-6fbcad2eed2d.png) | ![passcode draft 2](https://user-images.githubusercontent.com/736109/51236514-9b101780-1972-11e9-9cb7-1c2ad3efcade.png)

haukman:
I prefer Draft 1 as it looks a bit "lighter" and not so heavy as Draft 2.

PR #242, opened 2019-01-29
[UI] Refresh Passcode UI 

hosy:
## Description

- Passcode digit buttons are now rounded buttons
- Increased font size of passcode buttons
- support for external keyboards

## Related Issue
#222 

## Motivation and Context
The passcode UI should get a more iOS like UI.
I have two design suggestions which can you see below.

- Increased the font size
- Round Button, also used in iOS for passcodes
- External keyboard connected

## How Has This Been Tested?
Enable **Passcode lock** in app settings and **Lock Application** **Always**. Set app in background mode and back in foreground mode. Passcode view does appears and you set the changed UI. Please connect an external keyboard for testing keyboard support.
Furthermore disable **Passcode lock** in app settings and check if **Cancel** button can be triggered with ESC key of external keyboard.

| simulator screen shot - iphone xs max - 2019-01-29 at 20 56 35 | simulator screen shot - ipad pro 12 9-inch 3rd generation - 2019-01-29 at 20 56 57 |
|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|

## Types of changes

## Checklist:


ownCloud/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -257,4 +265,54 @@ class PasscodeViewController: UIViewController, Themeable {
...
+    
+    // MARK: - External Keyboard Commands
+    
+    @objc func performKeyCommand(sender: UIKeyCommand) {
+        guard let key = sender.input else {
+            return
+        }
+        
+        switch key {
+        case "\u{8}":
```
pablocarmu: Maybe you could code `"\u{8}"` into a constant to provide more information. Then you can reuse this constant in line `300`

ownCloud/Theming/UI/ThemeRoundedButton.swift
```diff
@@ -0,0 +1,41 @@
...
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        styleButton()
+    }
+    
+    private func styleButton() {
+        if self.frame.size.height < self.frame.size.width {
+            self.layer.cornerRadius = self.frame.size.height / 2
```
mneuwert: @hosy is it intended not to round the result of division?

ownCloud/Theming/NSObject+ThemeApplication.swift
```diff
@@ -69,7 +69,7 @@ extension NSObject {
 
 				case .bigTitle:
 					themeButton.themeColorCollection = collection.neutralColors
-					themeButton.titleLabel?.font = UIFont.systemFont(ofSize: 22)
+					themeButton.titleLabel?.font = UIFont.systemFont(ofSize: 34)
```
mneuwert: @hosy Would not use magic value as font size (at least a constant would be nice). And also have seen in another PR, that we started using dynamic type. Would it be an option here?

ownCloud/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -257,4 +265,54 @@ class PasscodeViewController: UIViewController, Themeable {
...
+            deleteLastDigit()
+        case UIKeyCommand.inputEscape:
+            cancelHandler?(self)
+        default:
+            appendDigit(digit: key)
+        }
+        
+    }
+    
+    override var keyCommands: [UIKeyCommand]? {
```
mneuwert: @hosy why is it optional? Below you initialize an empty array which is always returned.. Also can we do lazy init here instead of creating new array whenever getter is called?

ownCloud/Settings/Passcode/PasscodeViewController.swift
mneuwert: @hosy Would suggest to use human-readable constant instead of `"\u{8}"`

ownCloud/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -257,4 +265,54 @@ class PasscodeViewController: UIViewController, Themeable {
...
+            keyCommands.append(
+                UIKeyCommand(input:String(i),
+                             modifierFlags: [],
+                             action: #selector(self.performKeyCommand(sender:)),
+                             discoverabilityTitle: String(i))
+            )
+        }
+        
+        keyCommands.append(
+            UIKeyCommand(input: "\u{8}",
```
mneuwert: @hosy Would suggest to use human-readable constant instead of `"\u{8}"`

ownCloud/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -199,36 +199,44 @@ class PasscodeViewController: UIViewController, Themeable {
...
+    
+    private func appendDigit(digit: String) {
+        if !keypadButtonsEnabled || keypadButtonsHidden {
+            return
+        }
+        
+        if let currentPasscode = passcode {
+            // Enforce length limit
+            if currentPasscode.count < passcodeLength {
+                self.passcode = currentPasscode + String(digit)
```
mneuwert: @hosy digit is already String. String(String) doesn't make sense

ownCloud/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -199,36 +199,44 @@ class PasscodeViewController: UIViewController, Themeable {
...
+            }
+        } else {
+            self.passcode = String(digit)
+        }
+        
+        // Check if passcode is complete
+        if let passcode = passcode {
```
mneuwert: @hosy personally find constructs `if let x = x` confusing..

ownCloud/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -199,36 +199,44 @@ class PasscodeViewController: UIViewController, Themeable {
...
+            if passcode.count == passcodeLength {
+                // Delay to give feedback to user after the last digit was added
+                OnMainThread(after: 0.1) {
```
mneuwert: @hosy please use `let` constant instead of magic number

ownCloud/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -199,36 +199,44 @@ class PasscodeViewController: UIViewController, Themeable {
...
 	@IBAction func deleteLastDigit(_ sender: UIButton) {
-		if passcode != nil, passcode!.count > 0 {
-			passcode?.removeLast()
-			updatePasscodeDots()
-		}
+        deleteLastDigit()
 	}
+    
+    private func deleteLastDigit() {
+        if passcode != nil, passcode!.count > 0 {
```
mneuwert: @hosy wouldn't `if passcode?.count > 0` be shorter?

ownCloud/Settings/Passcode/PasscodeViewController.swift
hosy: Makes sense. I put the code in a new method and did not make changes to the existing code

ownCloud/Settings/Passcode/PasscodeViewController.swift
hosy: Does not work, because of a compiler error

ownCloud/Settings/Passcode/PasscodeViewController.swift
hosy: Thank you for pointing to this improvement

ownCloud/Theming/NSObject+ThemeApplication.swift
hosy: Should be addressed in a new issue.

ownCloud/Settings/Passcode/PasscodeViewController.swift
mneuwert: @hosy right.. i was wrong

jesmrec:
Code approved, let's move to QA

jesmrec:
All work OK. The only thing i was not able to get is the following:

https://user-images.githubusercontent.com/736109/51936427-7606b480-2408-11e9-89b0-65b7ed3f833e.png

Issue #199, opened 2019-01-06
[UX] Custimize Launch Screen 


hosy:
At some moment the launch screen has a solid white background. 
The launch screen should be customized with the ownCloud branding.

jesmrec:
creation a splash screen with the ownCloud log.

PR #244, opened 2019-01-31
Launch Screen

pablocarmu:
## Description

Fixes https://github.com/owncloud/ios-app/issues/199
## Related Issue

https://github.com/owncloud/ios-app/issues/199
## Types of changes

## Checklist:


ownCloud.xcodeproj/project.pbxproj
```diff
@@ -1486,6 +1489,7 @@
 				DCE5E8AD2080D781005F60CE /* folder-starred.tvg in Resources */,
 				DCE5E8A22080D781005F60CE /* folder-external.tvg in Resources */,
 				DC85572D20513B8C00189B9A /* ServerListTableViewController.xib in Resources */,
+				6E938522220322F30049D676 /* SplashImage.png in Resources */,
```
hosy: Why do you not store the SplashImage.png in Assets.xcassets and why do you not provide @2x @3x resources?

ownCloud/Resources/Base.lproj/LaunchScreen.storyboard
```diff
@@ -13,7 +17,18 @@
                     <view key="view" contentMode="scaleToFill" id="Ze5-6b-2t3">
                         <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
-                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                        <subviews>
+                            <imageView clearsContextBeforeDrawing="NO" userInteractionEnabled="NO" contentMode="scaleAspectFit" horizontalHuggingPriority="251" verticalHuggingPriority="251" image="SplashImage.png" translatesAutoresizingMaskIntoConstraints="NO" id="WtJ-CJ-2he">
+                                <rect key="frame" x="0.0" y="20" width="375" height="647"/>
```
hosy: I would scale the image canvas a little bit down, with a padding to the left and right side. Looks like a little bit oversized.

ownCloud.xcodeproj/project.pbxproj
pablocarmu: It's a 2048x2048 Image, so iOS reescalation works fine in this case.

We're not using assets catalog because of branding options. It's easier to exchange two images rather than make an asset catalog and exchange it with the actual one.

About the background color, we're used to switch the color modifying directly the *.XIB* file in the automatic-branded build process.

If you want me to use assets catalog I can do it but then we need to think how to create one in automatic-branded build process in our CI/build systems.

ownCloud/Resources/Base.lproj/LaunchScreen.storyboard
pablocarmu: How much do you want me to scale it? 30 points each side should be enough?

hosy: I think you should be include a scaling factor. If you use a static constant, this does not make sense for every device size. On a iPad screen it needs much more spacing. Please can you play with a scaling factor and post a screenshot on a iPad 12.9 and iPhone SE?

ownCloud.xcodeproj/project.pbxproj
hosy: Please keep the current behavior. It makes sense for branding. Maybe we will change it later.

ownCloud/Resources/Base.lproj/LaunchScreen.storyboard
pablocarmu: Take a look at the latest code, I've updated the constraints for every size class and I think looks better now.

hosy: Thank you! Looks nice!

jesmrec:
Launch screen correctly shown in iPhone and iPad, portrait and landscape.

Approved

Issue #232, opened 2019-01-23
[UX] Full-width buttons does not fit Apple design lines 


michaelstingl:
The full-width button doesn't look good in 2019. That's also what Apple recommends in the HIG:

> A button that extends to the edges of the screen might not look like a button. Respect the standard UIKit margins on the sides of full-width buttons. A full-width button appearing at the bottom of the screen looks best when it has rounded corners and is aligned with the bottom of the safe area—which also ensures that it doesn't conflict with the Home indicator.

(Source: [Apple HIG](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/))

In our project, I see another problem. Depending on branding colors, full-width buttons could be extra-confusing. Imagine the section separators "User and Groups" and "Public Links" from the sharing screen of the old app would have the same color as the buttons. An easy to recognize button shape will make user experience better.

Login view | Account confirmation | Old iOS app
:--: | :--: | :--:
![2019-01-23 01 26 37](https://user-images.githubusercontent.com/214010/51574535-7c4bdc80-1eae-11e9-840c-978e342fcdad.png) | ![2019-01-23 01 26 56](https://user-images.githubusercontent.com/214010/51574536-7ce47300-1eae-11e9-85cd-9ce5bac8881e.png) | ![2019-01-23 01 27 53](https://user-images.githubusercontent.com/214010/51574537-7ce47300-1eae-11e9-9515-83dc379e50db.png)
😄 | 👎 | 😕 

/cc @wuenschedesign

Issue #47, opened 2018-05-13
[UX] Confirm bookmark edit action on top-right


michaelstingl:
I was thinking about a button on the top-right to confirm bookmark edit action. This would also prevent the keyboard to cover the "Connect" button.

| Step 1 | Step 2 |
| --- | --- |
| ![img_3700](https://user-images.githubusercontent.com/214010/39971284-b57dc118-56f8-11e8-9ce5-5fb02329ef14.jpg) | ![img_3703](https://user-images.githubusercontent.com/214010/39971285-b5984c86-56f8-11e8-99d2-0d7655929967.jpg) |

(thanks @frankaellen for the feedback! 👍 )

felix-schwarz:
I've looked at how other apps handle this.

One of my favorite approaches is to remove the blue "Continue" button from the table view entirely (so it no longer scrolls with it) and show it permanently at the bottom of the screen (moving up if the keyboard is shown, so it never gets hidden by it).

That - or the approach you suggested.

michaelstingl:
@felix-schwarz How does the the floating button work in landscape mode? (above keyboard, with thethering and recording bar on top of the screen) . How mich vertical space remains?

michaelstingl:
@mmodrow FYI

michaelstingl:
We should put it on the top-right.

michaelstingl:
Next TestFlight build will bring the button to the top-right:

![img_0519](https://user-images.githubusercontent.com/214010/52496762-8e2ebe80-2bd4-11e9-83ea-9107d9aa113b.PNG) | ![img_0520](https://user-images.githubusercontent.com/214010/52496763-8e2ebe80-2bd4-11e9-8de3-56afeb4156bc.PNG)
-- | --


Not everybody in the team is a fan, but we'd like to test-drive this…

PR #262, opened 2019-02-07
StaticTableViewButtons corner radius

pablocarmu:
## Description

Fix for https://github.com/owncloud/ios-app/issues/232

## Related Issue

Fix https://github.com/owncloud/ios-app/issues/232
## Types of changes

## Checklist:


ownCloud/UI Elements/RoundedTableViewCell.swift
```diff
@@ -0,0 +1,41 @@
...
+
+	required init?(coder aDecoder: NSCoder) {
+		fatalError("init(coder:) has not been implemented")
+	}
+
+	override func layoutSubviews() {
+		super.layoutSubviews()
+
+		let maskLayer = CAShapeLayer()
+		let width = self.bounds.width - ((lateralInset * 2) + (safeAreaInsets.left * 2))
```
hosy: safeAreaInsets.left can be differ from safeAreaInsets.right
I think you should add the safeAreaInsets.right instead of multiply

ownCloud/UI Elements/StaticTableViewRow.swift
```diff
@@ -293,11 +293,15 @@ class StaticTableViewRow : NSObject, UITextFieldDelegate {
 
 	// MARK: - Buttons
 	convenience init(buttonWithAction action: StaticTableViewRowAction?, title: String, style: StaticTableViewRowButtonStyle = StaticTableViewRowButtonStyle.proceed, identifier : String? = nil) {
+
+		let lateralInset: CGFloat = 20
+		let cornerRaidus: CGFloat = 6
```
hosy: cornerRaidus should be replaced by cornerRadius (typo)

pablocarmu: Fixed

ownCloud/UI Elements/RoundedTableViewCell.swift
pablocarmu: You're right, I changed it.

michaelstingl:
@pablocarmu this became also relevant in context of https://github.com/owncloud/ios-app/issues/47 . I hope we'll get some feedback about it…

jesmrec:
Sent an additional commit to fix the UI tests after the new ` Continue` location. @hosy @pablocarmu, you can take a look if you want to review

jesmrec:
My question: 

I know that the ' Continue' location on the top-right is more Apple-styled, but i think that we are losing a bit of of UX. I mean:

The current login process starts on the top of the screen by asking for the URL. When it is correct, there are displayed credentials etc... so the UI flow moves from the top to the bottom. 

With the previous implementation, the button is always close to the tapping finger (just below the cell). With the current, the finger must move to the top every step. And every step is farther from the top. 

This is my opinion (not blocker of course)

PR #273, opened 2019-02-12
Use displayname instead of username

mneuwert:
## Description
In case of LDAP connected server, the logged in user name can be a full name which shall be shown instead of short user name (e.g. <first_name> <last name>

## Related Issue
#133 

## Types of changes

## Checklist:


michaelstingl:
@jesmrec this needs good testing. you have the most experience…

mneuwert:
> @jesmrec this needs good testing. you have the most experience…

@michaelstingl agree, because I just tested that `displayName` is at least equal to `shortName` when the user is logged in, but I don't have access to servers which have LDAP integration in place for instance.

michaelstingl:
Yeah, there are crazy setups out there…

jesmrec:
I will take care

ownCloud/Server List/ServerListTableViewController.swift
```diff
@@ -92,6 +92,8 @@ class ServerListTableViewController: UITableViewController, Themeable {
 		Theme.shared.register(client: self)
 
 		welcomeOverlayView.layoutSubviews()
+
+        self.tableView.reloadData()
```
hosy: Please indent code.
Why is here a reload necessary now?

mneuwert: @hosy fixed indentation (may be too aggresively). Reload is necessary, since usually you would have an LDAP display name after you have connected, and then you want to see this name when you return back to server list. If the table view won't be updated, you would see just a login name.

mneuwert:
@jesmrec Matthias has approved code changes. Can we proceed with QA?

jesmrec:
yes, i will start asap

ownCloud/Server List/ServerListTableViewController.swift
```diff
@@ -300,72 +302,72 @@ class ServerListTableViewController: UITableViewController, Themeable {
...
-								})
-							}, for: bookmark)
-						}))
 
-						self.present(alertController, animated: true, completion: nil)
-					}
-				}),
+									if let removeIndex = self.lockedBookmarks.index(of: bookmark) {
+										self.lockedBookmarks.remove(at: removeIndex)
+									}
```
jesmrec: correct indentation in this file?

mneuwert: @jesmrec I re-indented the file using following Xcode settings. Should be fine
screenshot 2019-02-14 at 15 11 18

jesmrec:
Display name is retrieved from the server after file list is loaded. When the bookmark is created, in the bookmark list is displayed the username (this is not a bad behaviour, just to be sure we are all in the same page)

jesmrec:
I have checked with regular servers and LDAP (using uid and 

- [X] Displayname with blanks
- [X] Displayname with special characters
- [X] Displayname very long

Some comments:

- Display name is set when file list is loaded, not before. When the bookmark is added the first time, in the bookmark list you'll see the username.
- Changes on the displayname are available when the app is killed and reopened. 

These two are not bad behaviours, i only comment to have available the current behaviour.

The improvement here would be to set the display name in the bookmark list as part of the URL. I mean: if you set a bookmark name, the displayname will not be visible nowhere:

screen shot 2019-02-18 at 09 22 05

the line below could include the displayname in @ notation:

screen shot 2019-02-18 at 09 22 05

what do you think? could be developed here, or improvement in a different PR? @mneuwert @michaelstingl

jesmrec:
Also, tests are failing in the authentication section:

```
CreateBookmarkTests.testLoginBasicAuthRightCredentials()
CreateBookmarkTests.testLoginBasicAuthWarningIssue()
CreateBookmarkTests.testLoginOAuth2RightCredentials()
```

mneuwert:
@jesmrec Regarding:
> * Display name is set when file list is loaded, not before.
Display name is only available when we have a connected core. And right now the connection is established, when user taps the bookmark in the list, and `ClientRootViewController` is created.

mneuwert:
@jesmrec @felix-schwarz please check out the latest more lightweight implementation of this feature.

jesmrec:
Will do a new round to check the new implementation. 

i triggered tests with the latest commit, and keep failing in https://github.com/owncloud/ios-app/blob/789d5363b1e0a22fe8a725ae3bfb78d0813b3e19/ownCloud/Bookmarks/BookmarkViewController.swift#L448

felix-schwarz:
> @jesmrec @felix-schwarz please check out the latest more lightweight implementation of this feature.

Looking good to me!

jesmrec:
i will add https://github.com/owncloud/ios-app/pull/273#issuecomment-464633452 as a sepparate improvement

Issue #278, opened 2019-02-19
Wrong spacing for Sort menu in file list


hosy:
### Steps to reproduce
1. Open the file list 

### Expected behaviour
Sort menu should be centered and should have the same spacing to left and right.

### Actual behaviour
Less spacing to the right side.

### Client
**iOS version:**
12.1.4

**ownCloud app version:**
1.0 (104)

**Device model:**
iPhone X, iPhone 8 

![img_cbb592f16bfc-1](https://user-images.githubusercontent.com/736109/53007975-a2768500-3438-11e9-82e6-95e839e934f7.jpeg)

jesmrec:
Done and merged

PR #280, opened 2019-02-21
Fixed wrong spacing for Sort menu in file list

hosy:
## Description
StackView has only two elements, spacing on the right side is to less and sort menu is not centered in view. Added the width value of the left button to the right anchor of the stack view to get the sort menu centered.

## Related Issue
#278 

## Motivation and Context
UI fix, because I can not unsee, that element is not centered in view.

## How Has This Been Tested?
Open file list on iPhone. Please check that everything is fine also in landscape mode.
Test also on iPad, if UISegmentedControl is centered.

simulator screen shot - iphone 8 plus - 2019-02-21 at 11 23 50

## Types of changes

## Checklist:


ownCloud/Client/SortBar.swift
```diff
@@ -163,7 +163,7 @@ class SortBar: UIView, Themeable {
 			stackView.topAnchor.constraint(equalTo: topAnchor),
 			stackView.bottomAnchor.constraint(equalTo: bottomAnchor),
 			stackView.leftAnchor.constraint(equalTo: safeAreaLayoutGuide.leftAnchor, constant: 20),
-			stackView.rightAnchor.constraint(equalTo: safeAreaLayoutGuide.rightAnchor, constant: -20)
+			stackView.rightAnchor.constraint(equalTo: safeAreaLayoutGuide.rightAnchor, constant: -50)
```
mneuwert: @hosy I am not a friend of magic numbers spread over the code. Would rather prefer using let constants. But may this is something we could do later.

michaelstingl: > magic numbers

👎

hosy: That's correct. There a many magic numbers to fix in this implementation. This was just a quick fix but I can do this now.

mneuwert: @hosy @michaelstingl I think this is not only UI class which uses magic numbers. We could create a separate task to move spacing, margin values into a separate file defining these constants.

michaelstingl:
@hosy Portrait and Landscape? With notch and without notch?

hosy:
@michaelstingl yes, tested in all variants.

hosy:
@michaelstingl safeAreaLayoutGuide is used for left and right padding (top and bottom is not relevant for this view)

jesmrec:
Approved.

Issue #58, opened 2018-05-30
[QA] UI testing 


jesmrec:
This issue is intended to be an epic to discuss about automatic testing.

Earlgrey is the chosen tool to develop UI tests. The branch [earlgrey-login](https://github.com/owncloud/ios-app/tree/feature/earlgrey_login) contains a first draft of a couple of tests. 

We can face two different targets using Earlgrey:

- **UI testing**: check how the app reacts to the SDK's responses. Abstracting dependencies from SDK and mocking responses make posible to isolate the app. This kind of testing complements the unit test in the SDK to have a greater coverage through the full stack of the app. 

- **E2E testing**: the scope is the whole app, by using the SDK itself. In this case the whole stack is tested, but is more difficult to abstract tests, and positive results will not give the accurate point of the issue/bug. By using a real server, network/core performance issues can happen causing wrong positives and making the tests flaky. 

The point is that we need to abstract dependencies in the app for UI testing, **by making testable code**. 
Currently, the app calls directly the SDK interfaces to get info. We should be able to use the same calls with mocked objects, in a way similar to the following one:

1. Create mocked object(s)
2. Set the mocked object(s) with the desired response
3. Perform actions 
4. The mocked SDK answers with the canned response in 2.
5. Check and assert what happens in UI 

for each test. Options to achieve it:

- Put all dependencies as parameters. Therefore, by subclassing we can call the functions with mocks. Of course, this point has code design/structure implications.
- Use the SDK dependency as a mock. 
- ...  (more ideas here).

All additional and related ideas are welcome.

@felix-schwarz @javiergonzper @pablocarmu @michaelstingl

pablocarmu:
Here it is a little summary of the discussions/thoughts @felix-schwarz  and I had about how to make UI testing in the app more straightforward:

#### Host Simulator

There is a mechanism called Host Simulator in the SDK that intercepts previously selected network calls and mock the response for this particular call with a set response. There are examples of how to use this mechanism in some SDK tests, like [here](https://github.com/owncloud/ios-sdk/blob/master/ownCloudSDKTests/HostSimulatorTests.m)

The problem using the Host Simulator is that is going to be impossible to test particular app states in a scalable way. 

#### Key-Value Observing

Set the values of the SDK objects directly to that they need to be to test the app. For read-only properties or private variables, we can leverage the `-setValue:forKey:` method from KVO.

#### Method Swizzling

 Swizzle individual methods to replace them with versions to return the values needed for testing.

[...]

PR #285, opened 2019-02-27
Earlgrey/more view

javiergonzper:
## Description
- Added more UI Tests for the More View.
- Refactorized the Swizzling methods are called in order to not duplicate code
- Included the Copyright headers on several files of UITests

## Related Issue
https://github.com/owncloud/ios-app/issues/58


ownCloudTests/File List/MoreViewTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+			//Reset status
+			EarlGrey.select(elementWithMatcher: grey_accessibilityID("dimming-view")).perform(grey_tap())
+			EarlGrey.select(elementWithMatcher: grey_allOf([grey_accessibilityLabel("Back"), grey_accessibilityTrait(UIAccessibilityTraits.staticText)])).perform(grey_tap())
+		} else {
+			assertionFailure("File list not loaded because Bookmark is nil")
+		}
+	}
+
+	func showFileList(bookmark: OCBookmark, issue: OCIssue? = nil) {
```
hosy: This code is duplicated in some classes. Is it possible to avoid code duplication for this method?

javiergonzper: I were thinking about it. Probably moving it to "UtilsTests"

jesmrec:
It is OK from my side. I added an additional assertion to check that the correct item name is shown in the card.

Issue #302, opened 2019-03-06
Remove full-width button in passcode view


hosy:
### Steps to reproduce
1. Open Settings
2. Enable / Disable Pin-Lock

### Expected behaviour
- In passcode view the `Cancel` button should use new implementation (not full width)
- Error message (long translation like German) should be completely readable

### Actual behaviour
- `Cancel` button has a full-width
- Error message (long translation like German) is not completely readable

![img_7609](https://user-images.githubusercontent.com/736109/53878501-54eb4200-400c-11e9-9012-ea44ba1af409.PNG)

### Client
**iOS version:**
12.1.4

**ownCloud app version:**
1.0.0 (110)

**Device model:**
iPhone, iPad

PR #308, opened 2019-03-06
Changed full width button to a rounded button

hosy:
## Description
- changed full width button to a rounded button
- fixed passcode view, error label and cancel button for smaller display sizes (iPhone SE)

## Related Issue
#302 

## Motivation and Context
- remove full width buttons, because it doesn't look like a button (see #232)

## How Has This Been Tested?
- open settings
- enable / disable pass-lock 
- please test on iPhone SE, iPhone XS Max, iPad Pro 12,9

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Server List/ServerListTableViewController.xib
+++ b/ownCloud/Server List/ServerListTableViewController.xib
@@ -56,6 +56,9 @@
                             <rect key="frame" x="0.0" y="283" width="241.5" height="43"/>
                             <color key="backgroundColor" red="0.27450980390000002" green="0.54901960780000003" blue="0.7843137255" alpha="1" colorSpace="calibratedRGB"/>
                             <accessibility key="accessibilityConfiguration" identifier="addServer"/>
+                            <constraints>
+                                <constraint firstAttribute="width" relation="lessThanOrEqual" constant="280" id="ddm-Lk-mEG"/>
+                            </constraints>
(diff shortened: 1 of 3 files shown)


jesmrec:
Button has the same size in iPhone and iPad. In iPhone, the button takes more or less 2/3 of the screen width, but in iPad it'd bu about 1/3. Should the button has an scalable size?

iPad: 

screen shot 2019-03-07 at 08 43 51

iPhone:

screen shot 2019-03-07 at 08 44 52

michaelstingl:
Please align button-width behaviour across the app. (compare with button on initial empty server screen) 

Alternative: Add it as text-button to the top-right. (not beautiful, but works reliable)

hosy:
I tried it with a larger button on the iPad yesterday, but it looks not nice. Currently it has the same width as the buttons above.

hosy:
The current implementation looks not adapted to the iPad:
img_0035

hosy:
In my opinion it would looks better with a smaller button

michaelstingl:
> In my opinion it would looks better with a smaller button

works for me

hosy:
Changed max width of buttons. `Add Account` and `Cancel` looks now equal and not to large on iPad:

simulator screen shot - ipad pro 12 9-inch 3rd generation - 2019-03-07 at 19 56 18
simulator screen shot - iphone se - 2019-03-07 at 19 56 02

jesmrec:
ok, let's go for the approach. Approved.

Issue #48, opened 2018-05-13
[UX] Field navigation in edit bookmark screen


michaelstingl:
| any field | last field |
| --- | --- |
| ![img_3703-a](https://user-images.githubusercontent.com/214010/39971414-cdac2426-56fa-11e8-8ce8-aefd19d6b99f.jpg) | ![img_3703-b](https://user-images.githubusercontent.com/214010/39971415-cdc49c68-56fa-11e8-95dd-c74da1f28895.jpg) |

(thanks @frankaellen for the feedback! 👍 )

michaelstingl:
> how about "Continue" + "Go"?

👍

michaelstingl:
@mmodrow FYI

michaelstingl:
I just found, when using OAuth 2.0 for login, we already have arrows and `Done` in the `SFAuthenticationSession` view. We should have the same buttons on both screens. Can we control what buttons to show in the `SFAuthenticationSession` too?

| edit bookmark screen | `SFAuthenticationSession` 1st field in focus | `SFAuthenticationSession` last field in focus |
|---|---|---|
| ![2018-08-21 19 06 57](https://user-images.githubusercontent.com/214010/44449111-8743cd80-a5ed-11e8-82ad-6887f9df46d1.png) | ![2018-08-21 19 07 53](https://user-images.githubusercontent.com/214010/44449112-8743cd80-a5ed-11e8-855f-b54e18be6427.png) | ![2018-08-21 19 08 22](https://user-images.githubusercontent.com/214010/44449214-cbcf6900-a5ed-11e8-95d9-93b592da3c93.png) |

(screenshots show iOS12 beta)

michaelstingl:
Seems the arrows are Safari-only, nothing that can be used in own Apps 😞 

Seems, this has to do with [`inputAccessoryView`](https://developer.apple.com/documentation/uikit/uiresponder/1621119-inputaccessoryview) (attach an accessory view to the system-supplied keyboard ).

Of course, there are projects that try to mimic the Safari style:
https://github.com/danielinoa/TextFieldsTraversalController

hosy:
I added a `inputAccessoryView` for rows username and password. `ReturnKeyType` will also be set.
@michaelstingl you can checkout this branch if this is working for you.

[...]

PR #312, opened 2019-03-10
Added keyboard field navigation for username, password textfield in bookmark view

hosy:
## Description
Added a `inputAccessoryView ` for username and password text field for field navigation between these rows, also setting a `ReturnTypeKey`.

## Related Issue
#48 

## Motivation and Context
Better navigation between text fields form keyboard, because next web view for oAuth 2 does also have this field navigators.

## How Has This Been Tested?
- add or edit a bookmark
- set focus on username or password text field

Simulator Screen Shot - iPhone 8 - 2019-03-10 at 11 31 35

## Types of changes

## Checklist:


ownCloud/Bookmarks/BookmarkViewController.swift
```diff
@@ -140,6 +155,7 @@ class BookmarkViewController: StaticTableViewController {
 		// Credentials section + rows
 		usernameRow = StaticTableViewRow(textFieldWithAction: { [weak self] (_, sender) in
 			if (sender as? UITextField) != nil, self?.bookmark?.authenticationData != nil {
+				//sender.inputAccessoryView = inputToolbar
```
mneuwert: @hosy remove dead code

ownCloud/Bookmarks/BookmarkViewController.swift
```diff
@@ -727,6 +743,21 @@ class BookmarkViewController: StaticTableViewController {
 	func isAuthenticationMethodTokenBased(_ authenticationMethodIdentifier: OCAuthenticationMethodIdentifier) -> Bool {
 		return authenticationMethodTypeForIdentifier(authenticationMethodIdentifier) == OCAuthenticationMethodType.token
 	}
+	
+	// MARK: - Keyboard AccessoryView
+	@objc func toogleTextField (_ sender: UIBarButtonItem) {
```
mneuwert: @hosy Instead of comparing `sender.style` I would suggest to create separate action methods

hosy: Make sense in this case

jesmrec:
For me it is OK, but a tiny detail:

In the "Add account" scenario, you are lighting/shadowing the arrows like:

Screen Shot 2019-03-21 at 13 51 16

Screen Shot 2019-03-21 at 13 51 23

In case of "Edit account" with a basic auth server, there is no choice to move up/down through text fields, because only `Password` is enabled:

Screen Shot 2019-03-21 at 13 52 56

so, the blue up arrow does not do nothing. It should be shadowed/hidden...

Apart of that, this is OK for me.

hosy:
@jesmrec I just fixed it. Accessory View is now only available, if the needed text fields are enabled.

jesmrec:
Approved

Issue #267, opened 2019-02-10
OAuth 2.0 end-to-end experience


michaelstingl:
Screenshot 1 | Screenshot 2 | Screenshot 3 | Screenshot 4
-- | -- | -- | --
![image](https://user-images.githubusercontent.com/214010/52532903-360ad000-2d2c-11e9-895f-0fcbb5ca18b5.png) | ![image](https://user-images.githubusercontent.com/214010/52532906-40c56500-2d2c-11e9-9840-67e6957c7132.png) | ![image](https://user-images.githubusercontent.com/214010/52532913-4fac1780-2d2c-11e9-8964-412d3575858d.png) | ![image](https://user-images.githubusercontent.com/214010/52532924-5c307000-2d2c-11e9-891e-a25efbc879a8.png)

### Problem 1

- Regular users don't understand why the 2nd screen (see **Screenshot 1**) after entering the server URL is useful
- We could explain this feature, and users should have an option to skip this screen in the future

@felix-schwarz @mneuwert @hosy 

### Problem 2

- Name of the app is different
- In the warning message, it says `"ownCloud" Wants to Use…` (see **Screenshot 2**)
- The OAuth 2.0 login screen says `The application "iOS" is requesting…` (see **Screenshot 3** &  **Screenshot 4**)

@DeepDiver1975 we should rename it in the OAuth 2.0 app to "ownCloud iOS app".

@felix-schwarz Any chance it could say "ownCloud app" or "ownCloud iOs app" in the warning message?

### Steps to reproduce
1. Install [OAuth 2.0 master (`40fe0ab`)](https://github.com/owncloud/oauth2/tree/40fe0abf6ff63e97f9e79ad9f2969314bd1d0144) on ownCloud server
2. Add account to the ownCloud iOS app

### Client
**iOS version:**
iOS 12.1.4 (16D57)

**ownCloud app version:**
ownCloud beta version 1.0 build 105 (https://github.com/owncloud/ios-app/tree/b19e57b2f67234460ae6dec8f9587d62ed34d2f0, built from branch)

**Device model:**
Model: iPhone XS Max

### Server configuration

**ownCloud version:**
ownCloud server 10.1.0 with master-[OAuth 2.0 (`40fe0ab`)](https://github.com/owncloud/oauth2/tree/40fe0abf6ff63e97f9e79ad9f2969314bd1d0144) on ownCloud server

/cc @hodyroff

hosy:
> ### Problem 1
[...]

PR #325, opened 2019-03-19
Added Additional OAuth and Certificate Infos

hosy:
## Description
For OAuth 2 there was added a header view which shows an information text, why in the next step a alert will shown. If OAuth is not supported by the server, this view should not appear.
Furthermore we showing now more information about the certificate validation result. In the row `Certificate Details` there is a new label with a short validation result label with a special color for this state. As section footer there is a longer description for the validation result.

## Related Issue
#267 

## Motivation and Context
Inform the user before OAuth 2 alert was shown with additional informations. Furthermore show the user validation informations for the server certificate.

## How Has This Been Tested?
- Add a new (OAuth 2) account
- Edit an existing account
- please test in landscape and portrait mode

Simulator Screen Shot - iPhone X - 2019-03-18 at 17 33 42

## Types of changes

## Checklist:

____

BUGS & IMPROVEMENTS

- [X] (1) Additional step also needed for edition in https: https://github.com/owncloud/ios-app/pull/325#issuecomment-476138115
- [X] (2) Warning in edition https://github.com/owncloud/ios-app/pull/325#issuecomment-476140850
- [ ] (3) Cell down displaced https://github.com/owncloud/ios-app/pull/325#issuecomment-476534671


hosy:
@michaelstingl should we keep this description or do we have a better wording:
`If you 'Continue', you will be prompted to allow the 'ownCloud' App to open OAuth 2 login where you can enter your credentials.`

ownCloud/SDK Extensions/OCCertificate+Extension.swift
```diff
@@ -0,0 +1,59 @@
...
+			case .reject:
+				color = UIColor.red
+				shortDescription = "Rejected".localized
+				longDescription = "Certificate was rejected by user.".localized
+			case .promptUser:
+				color = UIColor.orange
+				shortDescription = "Warning".localized
+				longDescription = "Certificate has issues.\nOpen 'Certificate Details' for more informations.".localized
+			case .passed:
+				color = UIColor(red: 0.173, green: 0.745, blue: 0.306, alpha: 1.000)
```
mneuwert: @hosy shouldn't we put the colors into the app theme?

ownCloud/Theming/ThemeCollection.swift
```diff
@@ -174,6 +175,7 @@ class ThemeCollection : NSObject {
 		self.tableBackgroundColor = UIColor.white
 		self.tableGroupBackgroundColor = UIColor.groupTableViewBackground
 		self.tableSeparatorColor = nil
+		self.tableRowBorderColor = UIColor.black.withAlphaComponent(0.1)
```
mneuwert: @hosy I think it makes sense to create a constant for 0,1 since it is also used multiple times

ownCloud/UI Elements/RoundedInfoView.swift
```diff
@@ -0,0 +1,94 @@
...
+	deinit {
+		Theme.shared.unregister(client: self)
+
+		if messageThemeApplierToken != nil {
+			Theme.shared.remove(applierForToken: messageThemeApplierToken)
+			messageThemeApplierToken = nil
+		}
+	}
+
+	private func styleView() {
```
mneuwert: @hosy may be `styleView` doesn't fully reflect what it does.. since it is also laying out the view

ownCloud/Resources/de-DE.lproj/Localizable.strings
```diff
@@ -39,6 +39,15 @@
 "Authenticated as %@ via %@" = "Authentifiziert als 1%@ über 1%@";
 "Edit" = "Bearbeiten";
 "Credentials" = "Zugangsdaten";
+"Rejected" = "Abgelehnt";
```
mneuwert: @hosy I am a bit surprised to see german translation here, since I was thinking those are maintained separately in transifex

hosy: When I submit it here it will also appear in Transifex. When I can do a translation directly it is better to do it right now, before we forgot to translate it.

ownCloud/Theming/ThemeCollection.swift
hosy: Ok

ownCloud/SDK Extensions/OCCertificate+Extension.swift
hosy: I am not sure about that, because it matches currently all themes. This colors are also fixed in the SDK.

ownCloud/Theming/ThemeCollection.swift
hosy: Just checked the code, but for me this are individual values for every color. I cannot reuse this values for the other colors, because it I change the constant, other colors will also get a change, which may not be wanted.

ownCloud/UI Elements/RoundedInfoView.swift
hosy: Solved

ownCloud/SDK Extensions/OCCertificate+Extension.swift
hosy: After rethinking it, you are right and I now using theme colors.

jesmrec:
Additional step in http is OK and fix the issue mentioned above. But this is causing the same effect in case when editing  https servers:

Check these steps:

1. Edit an https (user accepted)-OAuth2 account
2. Click on `Delete Authentication Data` 
3. Tap on Continue:

Mar-25-2019 11-29-02

jesmrec:
### (2)

Edit an http, or https-passed OAuth2 server, you will see:

Screen Shot 2019-03-25 at 11 36 58


in this step, the OAuth2 warning should not appear, so there is no `Continue` button. It should appear when you `Delete authentication data`, so it is the point to get the OAuth2 flow

hosy:
@jesmrec now fixed in edit mode.

jesmrec:
after the fix, OAuth2 warning is there in the correct step. I noticed that the layout is slightly down displaced:

| `master` | `oauth-experience` |
| --- | --- |
| Screen Shot 2019-03-26 at 10 05 30 |  Screen Shot 2019-03-26 at 09 31 16 |

with the OAuth2 warning inclusion, maybe the cell location is not correctly calculated. What do you think?

jesmrec:
Now it's going pretty fine. Checks performed:

- [X] OAuth2 + http (create & edit)
- [X] OAuth2 + https user accepted (create & edit)
- [X] OAuth2 + https secure (create & edit)
- [X] Regression: basic auth

everything OK. Approved

Issue #246, opened 2019-02-01
Multiselection II


jesmrec:
Will continue the work after https://github.com/owncloud/ios-app/issues/79

Actually, Move and Delete are done.

Suggestion for new options to include:

- [ ] Open In (as included in Files app)
- [ ] Open In > Save multiple images or videos to Photos app (from #79 (comment))
- [ ] Duplicate
- [ ] Copy

Also, a `Select All` and `Select inverse` will make the feature more usable.

Other suggestions, welcome.

michaelstingl:
- Open In > Save multiple images or videos to Photos app (from https://github.com/owncloud/ios-app/issues/79#issuecomment-457217958)
- Favorites (oC favorites and/or iOS favorites)

PR #327, opened 2019-03-22
Multiselection Part 2

mneuwert:
## Description
Feature allowing to select / deselect multiple files and handle following actions: 
- Open In
- Duplicate 
- Copy

## Related Issue
#246 

## Motivation and Context
Improves UX and user's productivity

## How Has This Been Tested?
n/a

## Types of changes

## Checklist:


ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -39,34 +39,32 @@ class DuplicateAction : Action {
...
-					fileExtension = ".\(fileExtension)"
-				}
+					if fileExtension != "" {
+						fileExtension = ".\(fileExtension)"
+					}
 
-				name = "\(baseName) copy\(fileExtension)"
-			}
-		}
+					name = "\(baseName) copy\(fileExtension)"
```
hosy: There is no fallback, if `"\(baseName) copy\(fileExtension)"` also exists. This is a general problem by our app, not only here. Maybe we need a function for checking, if a file name exists. If it exists, we need a new name `copy copy` or `copy 2` and recheck this name again, until it succeeds.

ownCloud/Client/Actions/Actions+Extensions/OpenInAction.swift
```diff
@@ -22,61 +22,95 @@ class OpenInAction: Action {
...
+			}
+		}
+	}
+
+	fileprivate func showDownloadError() {
+
+		guard let viewController = context.viewController else { return }
+
+		let appName = OCAppIdentity.shared.appName ?? "ownCloud"
+		let alertController = UIAlertController(with: "Cannot connect to ".localized + appName, message: appName + " couldn't download file(s)".localized, okLabel: "OK".localized, action: nil)
```
hosy: Do you have the context, if you are downloading one or multiple files? If yes, you could use a message for a single file (like before) or if there are multiple files, you can present the current message.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
mneuwert: @hosy I actually created separate issue for that, since I think it makes sense to keep stuff separated and not to try to solve issues which are out of scope for now.

ownCloud/Client/Actions/Actions+Extensions/OpenInAction.swift
mneuwert: @hosy I could show file name at this point in case it is just one file, but I dont know if it is that useable.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
hosy: Did not see that you already created an issue for that. Thanks!

jesmrec:
### (1)

In an iPad:

1. Select two or more files (no matter kind of file, or if they are downloaded or not)
2. Open In 

Current: App crashes
Expected: Dialog to open in other apps displayed

With iPhone, it works fine

Device: iPad Air v12

mneuwert:
@jesmrec I fixed the iPad related crash as well as some quirks related to multi-selection and drag & drop getting in each other's way. Please check the latest version on bitrise built from this branch

jesmrec:
Crash fixed and dragging options work.

Approved from my side.

Issue #228, opened 2019-01-21
[FR] Show the used quota inside the app


ChrisEdS:
### Expected behaviour
In our Android app we have the possibility to display the used quota. In my opinion, this useful feature should also be available in the new iOS app.

![image](https://user-images.githubusercontent.com/33026403/51485622-dcebf400-1d9e-11e9-84ad-bf4298613153.png) | 
![reflector_screenshot](https://user-images.githubusercontent.com/214010/51489091-9c44a880-1da7-11e9-8556-bfabfb94a7e2.png)
-- | --

mneuwert:
@jesmrec In the server list we have no active connection yet, so of course we could try to connect, just to get the quota but not sure.. Alternatively we could e.g. show it in the prompt of navigation bar:
<img width="503" alt="Screenshot 2019-03-27 at 07 47 42" src="https://user-images.githubusercontent.com/228537/55055862-afece380-5064-11e9-977e-e162c1638e42.png">

jesmrec:
Ok, other option is setting on the bottom (just above the options)

PR #337, opened 2019-03-28
Show the used quota inside the app

mneuwert:
## Description
- Changed SDK which retrieves quota if available and stores it in the instance of OCItem
- Changed the UI to show quota in the prompt of the navigation bar
- Right now space occupied by the files in the current folder is shown as bytes used information

## Related Issue
#228

## Motivation and Context
As a user I want to see how much storage space is already in use

out

## Types of changes

## Checklist:

## Test plan

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/User%20Quota.md

## Bugs & improvements

- [X] (1) Quota not shown https://github.com/owncloud/ios-app/pull/337#issuecomment-478907673 [FIXED]
- [ ] (2) Values not correctly displayed https://github.com/owncloud/ios-app/pull/337#issuecomment-479567354


mneuwert:
@michaelstingl @jesmrec please check out the video showing how current implementation is working. As you can see if user navigates through the folder structure the quota in use changes and the value is corresponding to the total size of files in that particular folder, so only in the root folder you would see the total space allocated by the files of the user. I think on one hand it is useful information (per folder total size) but on the other hand I am not sure if it is intuitive. Also any other suggestions / wishes / ideas are welcome

hosy:
@mneuwert thank you for your implementation. I thought the quota should shown in the account list, below the Server URL. Your current implementation can be misunderstood, because user can think the used quota is the size of the current folder. But I like the idea. Maybe you can change it, to only show the size of the folder without "of 5,21 GB" -> "679 kb used" and display the total used size of quota in the account list.

Maybe it is possible to use the new custom OCQuery by @felix-schwarz to get the total size of Documents, Photos, Videos, Others, (...) like the iCloud Storage overview:

iCloud Storage

michaelstingl:
Vertical space is valuable. Not sure if we should show it on top of every folder. But quota could be different in every folder, so we could show it with the folder-info…

felix-schwarz:
A few thoughts:
- I think of quota as a global number. I find it confusing when I start in the root folder with "161.9 MB of 5.21 GB used" and then get to see "135.6 MB of 5.21 GB used" in the Documents folder. The latter gives me the impression that I have more free space available than I actually have.
- If more than one quota is possible (I guess in shared folders (cc @michaelstingl)?), then we should have a concept of "quota roots", so that I get to see the quota numbers of the "quota root". So f.ex. "161.9 MB of 5.21 GB used" for all directories in my account and then "20 MB of 100 MB" for the shared folder with quota and all its subfolders.
- thinking of where to put the quota, I have no favorite, but also think that it shouldn't occupy vertical space permanently (unless, maybe, the remaining space is less than a percentage or fixed amount of total storage). Maybe a cell at the end of the table view? Or a new "Show quota" action that shows a nice table that lists all "quota roots" and their available and total space.

jesmrec:
First preliminary check:

- Root folder is not showing the quota:

Screen Shot 2019-04-02 at 10 57 45

jesmrec:
### (2)

I checked the different cases with some findings:

- An user with `Default` quota displays in root folder:

Screenshot 2019-04-03 at 18 34 20

the propfind returns:

```
<d:quota-used-bytes>348737755</d:quota-used-bytes>
<d:quota-available-bytes>-3</d:quota-available-bytes>
```

`quota-used-bytes` is the number of bytes used in the account, but if you display in MB, the value is not correctly transformed, (348737755 bytes are 332.6 MB)

in case of `-3` nothing should be displayed as limit, like `332.6 MB used`


- An user with limited quota. For example, 1GB:

Screenshot 2019-04-03 at 18 10 23

in such case the propfind responses something like

```
<d:quota-used-bytes>348737755</d:quota-used-bytes>
<d:quota-available-bytes>725004069</d:quota-available-bytes>
```

The result should be something like:  `332.6 MB  of 1 GB used `

- An user with `Unlimited` quota:

Screenshot 2019-04-03 at 18 12 48

same case as `Default`

- An user with `Other` quota:

same case as limited quota.

mneuwert:
@jesmrec  strictly speaking it the MB calculation is correct. There was an effort to unify data amount units, and actually as in your comment 348737755 bytes are 332.6 MiB (Mebibyte) but not MB (Megabyte). macOS I think is used mixed formats and decimal style (SI prefixes) is used to reflect how much storage is occupied by files and the "old" style (where KB is 1024 Bytes) is used to e.g. show how much memory is used say by a running process.

jesmrec:
thanks for the explanation @mneuwert. Calculation was already fixed.

jesmrec:
Everything fixed. Approved.

PR #346, opened 2019-04-01
[fix/autolayout] Fix auto-layout constraints in ClientItemCell 

felix-schwarz:
## Description
- Fix contraint conflicts in ClientItemCell / ClientQueryViewController
- Replace text-based more dots with actual image to improve performance

## Motivation and Context
Previously, verbose warnings were logged by iOS for every item cell, during regular display as well as drag & drop – likely impacting performance in the process.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #347, opened 2019-04-02
[fix/sdk-updates] Update to latest SDK and various app-fixes

felix-schwarz:
## Description
Updates the app to the latest SDK and fixes various issues in the app (incl. #346, which this PR supersedes).

## Related Issue
#346

## Types of changes

## QA

https://github.com/owncloud/ios-app/pull/347#issuecomment-479503910


ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -76,12 +76,17 @@ class ClientQueryViewController: UITableViewController, Themeable, UIDropInterac
...
-		var title = (query.queryPath as NSString?)!.lastPathComponent
+		var title : String?
 
-		if title == "/", let shortName = core?.bookmark.shortName {
-			title = shortName
-		}
+		if let queryTitle = (query.queryPath as NSString?)?.lastPathComponent {
+			title = queryTitle
 
+			if title == "/" {
```
mneuwert: @felix-schwarz I think having some category method on `OCPath` (`isRoot`) would be nicer than to check if it is "/"

mneuwert:
@felix-schwarz  don't think generally it is a good idea to squash all possible (unrelated) app fixes + SDK fixes into one big PR

ownCloud/Client/ClientQueryViewController.swift
felix-schwarz: Yep. Thought about this yesterday, too. I'll add that property in the next SDK update.

felix-schwarz:
> @felix-schwarz don't think generally it is a good idea to squash all possible (unrelated) app fixes + SDK fixes into one big PR

I fully agree. However, sometimes I've little choice. Xcode 10.2 started to complain about missing nullability annotations in places where Xcode 10.1 didn't - and the only way to fix IIRC 200+ warnings that would otherwise burry useful warnings regarding new code - was to add them to the SDK where needed and update the app alongside, so it continues to compile, handles `nil` values correctly where it didn't before and also respect SwiftLint's allergic reaction to `as!`. :-D

This kind of update should be much rarer in the future, though, as we're close to having nullability annotations almost everywhere in the SDK now.

Regarding the other fixes - like the item cell constraints fixes - I also think they could (should) go into separate PRs. But as a late entry before the code freeze, I didn't want to cause more overhead/delay than necessary.

ownCloud/Client/ClientQueryViewController.swift
mneuwert: @felix-schwarz ok, cool!

jesmrec:
some checks performed over this PR. Mainly added nullability checks, so i put my focus on:

- [X] Cell items in file list
- [X] Thumbnails 
- [X] Browsability
- [X] Auth basic and OAuth2

It is ok from my side

Issue #320, opened 2019-03-15
Duplicate file shows an error when a copy of the file already exists


mneuwert:
If you have a file _"XY.jpg"_ and then duplicate it, _"XY copy.jpg"_ is created. If you try to duplicate _"XY.jpg"_ again, app shows an error that  _"XY copy.jpg"_ already exists. For example, the "Files" app can deal with that. Instead of fixed "copy" suffix, it adds a number which is incremented on every new duplicate. We could do that in similar way.

### Steps to reproduce
1. Choose a file
2. Duplicate it
3. Try to duplicate the original file again.

### Expected behaviour
Duplicate action should not fail. I think it is a valid use case to create multiple copies of the same file.

### Actual behaviour
App refuses to create a duplicate

### Client
**iOS version:**
12.x

**ownCloud app version:**
master / 868097f150b66e1fd9b79d5996c4e06b4f0ae025

**Device model:**
n/a

PR #348, opened 2019-04-02
Correct handling for duplicated file names

hosy:
## Description
Fixed action `Duplicate` to get the correct file name for a duplicated file or folder.
Now the `macOS` `Finder` logic is used.

## Related Issue
#320 

## Motivation and Context
App showed error message, if file already exists.

## How Has This Been Tested?
1. Open file list
2. Select one ore more files
3. Files should be duplicated with new file name
4. `Paris.jpg` > `Paris copy.jpg`
5. `Paris copy.jpg` > `Paris copy 1.jpg`
6. `Paris.jpg` > `Paris copy 2.jpg`
7. `Paris copy 1 .jpg` > `Paris copy 3.jpg`

Use the same handling as `macOS` `Finder`.

## Types of changes

## Checklist:

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Duplicate%20II.md


jesmrec:
Approved.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -20,11 +20,14 @@ import Foundation
...
 import ownCloudSDK
 
-class DuplicateAction : Action {
+class DuplicateAction : Action, OCQueryDelegate {
 	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.duplicate") }
 	override class var category : ActionCategory? { return .normal }
 	override class var name : String? { return "Duplicate".localized }
 	override class var locations : [OCExtensionLocationIdentifier]? { return [.moreItem, .moreFolder, .toolbar] }
+	var query : OCQuery?
+	var localizedCopy = "copy".localized
```
mneuwert: @hosy don't think it is worth storing as extra property.. also should be constant and not `var`. And also, shall we call it` "blah copy x"`? macOS for instance just uses numbers.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -34,37 +37,126 @@ class DuplicateAction : Action {
...
-					if error != nil {
-						Log.log("Error \(String(describing: error)) duplicating \(String(describing: item?.path))")
-						self.completed(with: error)
-					} else {
-						self.completed()
+				var fileCounter = -1
+				for anItem in items {
+					guard let baseName = anItem.baseName else { return }
+
+					let pattern = "\(self.localizedCopy)[ ]?[0-9]{0,}$"
```
mneuwert: @hosy this pattern seems to be repeated, candidate for class-wide constant?

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -34,37 +37,126 @@ class DuplicateAction : Action {
...
 			completed(with: NSError(ocError: .itemNotFound))
 			return
 		}
+		remainingItems.append(contentsOf: context.items)
+		proceed()
+	}
 
-		context.items.forEach({ (item) in
-			if let rootItem = item.parentItem(from: core) {
+	func proceed() {
```
mneuwert: @hosy I think` proceed(`) as a method name doesn't reflect what it does

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -34,37 +37,126 @@ class DuplicateAction : Action {
...
-				if let itemFileExtension = item.fileExtension, let baseName = item.baseName {
-					var fileExtension = itemFileExtension
+		query = OCQuery(forPath: rootItemPath)
+		query?.delegate = self
+		query?.queryItem = item
 
-					if fileExtension != "" {
-						fileExtension = ".\(fileExtension)"
-					}
+		let filterHandler: OCQueryFilterHandler = { (_, _, item) -> Bool in
```
mneuwert: @hosy I think for all this filtering stuff, one or two lines of comments wouldn't hurt

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -34,37 +37,126 @@ class DuplicateAction : Action {
...
+			completed(with: NSError(ocError: .itemNotFound))
+			return
+		}
+		guard let rootItem = item.parentItem(from: core) else {
+			completed(with: NSError(ocError: .itemNotFound))
+			return
+		}
+
+		let pattern = " \(localizedCopy)[ ]?[0-9]{0,}$"
+		var copyFileName = "\(itemBaseName.replacingOccurrences(for: pattern)) \(localizedCopy)"
```
mneuwert: @hosy can we assume that all NOT yet duplicated files doesn't have "copy x" in the name? What if the file was named like that intentionally. So, try to duplicate a file and then duplicate a resulting duplicate in macOS.. You'll see what I mean. So if you duplicate a file `a.png`, it will create `a 2.png` and then if you duplicate `a 2.png`, it will produce `a 2 2.png`

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -34,37 +37,126 @@ class DuplicateAction : Action {
...
+			guard let queryItem = query.queryItem else { return }
+
+			switch query.state {
+			case .contentsFromCache, .waitingForServerReply:
+				break
+			case .idle:
+				self.core?.stop(query)
+
+				let items = changeSet?.queryResult ?? []
+				guard items.first != nil else {
```
mneuwert: @hosy why do we have extra handling code for the first item? And then we have a loop for all of them? Can't we have a generic loop iterating over all of them?

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
```diff
@@ -34,37 +37,126 @@ class DuplicateAction : Action {
...
+			case .targetRemoved: break
+
+			default: break
 			}
-		})
+		}
+	}
+
+	// MARK: - Copy Action
+	func copy(item: OCItem, with counter: Int) {
```
mneuwert: @hosy I guess parameter name `counter` could be even more descriptive

ownCloud/Resources/de.lproj/Localizable.strings
```diff
@@ -233,6 +233,7 @@
 "Cannot connect to " = "Keine Verbindung möglich zu";
 " couldn't download this file" = "Datei konnte nicht geladen werden";
 "Actions" = "Aktionen";
+"copy" = "Kopie";
```
mneuwert: @hosy Using only numbers like macOS would also remove a need to localize it

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
hosy: It differs with a leading space

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
hosy: What I'm doing here, is exactly what you can find in macOS Finder. If you duplicate a file which was already duplicated, Finder names this also with ending` copy x`
Samples can be found in this PR **How Has This Been Tested?**

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
hosy: Maybe `items.first` is wrong. But if you look at the code, I'm working with delegate methods for OCQuery. If you put it into a loop it would not work as expected, because if you duplicate two files with the same name, but with ending ` copy x`, both files would get the same name. So you have to copy the first item, before you can name the second item.

ownCloud/Resources/de.lproj/Localizable.strings
hosy: Did you duplicate a file in Finder? For me I'm getting `Filename copy.extension` or `Filename copy 1.extension`

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
hosy: Just checked, that `items.first` is not wrong, because if it not exists, there is an important fallback code.

ownCloud/Resources/de.lproj/Localizable.strings
mneuwert: @hosy hm.. yes, looks like `option + drag` and `cmd + D` do produce different results in finder

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
mneuwert: @hosy sorry, sometimes I am just questioning the code which is a bit hard to read at a first glance)

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
mneuwert: @hosy see my another comment.. Finder behaviour seems to be different depending on how you do it)

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
hosy: In my opinion it is not so difficult to understand. It is only a check if a baseName has a prefix.

ownCloud/Client/Actions/Actions+Extensions/DuplicateAction.swift
mneuwert: @hosy ok

PR #350, opened 2019-04-03
[fix/gallery-view] Fix double-download issue in Gallery View

felix-schwarz:
## Description
Fixes an issue where a tapped file finishes downloading, only to be downloaded *again* by the GalleryView, because the latter was using an outdated version of the `OCItem`.

## Types of changes


jesmrec:
Fix checked in iPhone and iPad devices.

approved.

Issue #351, opened 2019-04-04
Share with user/groups


jesmrec:
_As an user, i want to share my contents with other users or groups, in my owncloud server or in another one_

AC (already defined in epic):

* User can share any file or folder with other users and groups, by searching them (using patterns) and selecting the desired one
* User will see if the file is shared and the list of sharees
* Every share has its own attributes:
   - Permissions: edit and share in case of files. add, edit, delete and share in case of folders
   - Password: will be mandatory depending capabilities
   - Expiration date: will be mandatory depending capabilities
* Federation will be allowed. with @ notation will be distinguished when the user is in a different oC instance (if capabilities allow it)

The capabilities affecting private sharing are:

- "Allow apps to use the Share API" -> if false, sharing is not available
- "Allow resharing" -> If true and the file is shared with the user, no sharing available
- "Allow sharing with groups" -> if false, groups will not appear in searching
- "Restrict users to only share with users in their groups" 
- "Restrict users to only share with groups they are member of"
- "Exclude groups from sharing"
- "Allow username autocompletion in share dialog. If this is disabled the full username needs to be entered."
- "Default user and group share permissions"
- "Extra field to display in autocomplete results"
- Federation: allow federation, 

Some UI drafts will be welcome to discuss implementation.

related: Public Share: #275

hosy:
I finished a first draft and I'm interested in your opinion.

Starting point is the Action-View, where you have action for an item and further informations like file name, file size, modifying date and file icon. 
My thoughts was to show there if the item is shared. If the item is shared, you can see all recipients for the share. I decided to put the names one after the other without further informations and actions, because in this view it should only give a first overview and the view has not so much space for more actions.
If the item currently is not shared, you have an action to add a new share recipient.

| ![IMG_8108-2](https://user-images.githubusercontent.com/736109/55956374-1cbacd00-5c64-11e9-8fed-390286fd916a.png) | ![IMG_8105-2](https://user-images.githubusercontent.com/736109/55956367-188eaf80-5c64-11e9-90af-aebbf7da03d2.png) |
|-------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------|


If you tapped on the share recipients you will get a new sharing view, where you can see all recipients grouped by Users, Groups and Remotes. For every recipient you can see the granted permissions, without any need to look at each recipient in the detail view.
If you want to add a new share recipient, start typing into the search bar and after selecting the result the recipient a new share will be created for the item.

[...]

Issue #292, opened 2019-02-28
"No matches" label in file search is hidden by keyboard


hosy:
### Steps to reproduce
1. Open File list and type a search string into the search field, with non matching files

### Expected behaviour
Labels "No matches" and "There is no results for this search" should be completely visible for the user.

### Actual behaviour
Labels are currently hidden by the keyboard

![img_7544](https://user-images.githubusercontent.com/736109/53557555-7482f600-3b46-11e9-89e6-525d3f3a4af1.PNG)

### Client
**iOS version:**
12.1.4

**ownCloud app version:**
1.0.0 109 (ce6108d)

**Device model:**
iPhone X

hosy:
Fixed in #358

jesmrec:
Fixed in #358

Issue #275, opened 2019-02-14
Share public


jesmrec:
_As an user, i want to generate a public link to any of the items in my list, so i can share my content with other people_

AC (already defined in epic):

* User can create as many public links as he/she wants. It was available from 10.0.4, pending to check if it was backported in any way. For older versions, only one public link per item is allowed.
* Public links have several attributes:

   - Link name
   - Permissions: like in web, it will depend on the kind of item (file, folder):
      - Download / View: files & folders
      - Download / View / Upload: folders
      - Upload only (file drop): folders
   - Password: will be mandatory depending capabilities
   - Expiration date: will be mandatory depending capabilities

Capabilities are also important to take in account. They apply to every account in the same server. The capabilities affecting public sharing are:

- "Allow apps to use the Share API" -> if false, Share View is not available from Files View
- "Allow users to share via link" -> If false, Public link UI is hidden
- "Allow resharing" -> If true and the file is shared privately, no public share available
- "Enforce password protection for..." -> Show in UI somehow, that the password is mandatory
- "Set Default Expiration Date" -> Set the date as default 
- "Enforce expiration date" -> only available if "Set Expiration Date is true". If true, link is not created with empty expiration date.

Some UI drafts will be welcome to discuss implementation.

hosy:
#83

PR #358, opened 2019-04-12
[feature/sharing-user-groups] Implementation of Sharing with Users, Groups and Remote Users, Public Links

hosy:
## Description
Implementation of User, Group and Remote Sharing for files and folders

## Related Issue
#351 
#275  
#292 

## Motivation and Context
As one of a main feature of ownCloud it is now possible the share files and folders to other users.

### Sharing with Users, Groups
- shared files and folders should have a shared icon in the file list (folders have a special folder icon)
- on a file, which is currently not shared, tap on the `More` button
	- if sharing is possible for this item (depends on server capability, permissions) there is a new item "Share file" or "Share folder"
	- Tapping on this item will open the `Sharing` overview, where you can search for a new recipient
- on an item which is shared by you, you will see how many recipients are on this share
- if the item has public links, these are also visible, but cannot edited at the moment
- in the Sharing overview you can edit, delete and add recipients
- by selecting a share recipient you can change permissions, based on the item type
- if you have reshared the item, you can edit permissions for this recipient

### Public Links
- if item is sharable, there is a menu item `Create Public Link`, if not there is no menu entry
- tap on it will create a new public link with known options
- please test it with all regarded server capabilities

### Private Link
- open the menu item `Create Public Link`
- on the top of the screen there is a button `Copy Private Link`
- this will copy the private link url to the pasteboard

**Sorting in file list**
- you can now sort items in file list, if it is shared

### In the process of implementation, the following was changed:
**Issue #292:**
- Enter a search term in file list, which has no matches
- Delete all share recipients of an item
- If the keyboard is on screen, all text labels should be visible and not hidden by keyboard

**Add Menu**
- In file list, tap on `Add` button
- Menu items are now left aligned and have an left aligned icon

**More Menu**
- Action items are now left aligned and have left aligned icons

### Quick Access
This view was necessary to show pending shares, which needs to be accepted by the user.

**Pending Shares**
- if the user has pending shares,

**Shares**

**Collections**

## Types of changes

## Checklist:

## Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Sharing.md

## Bugfixing:

- [X] (1) Folder with "+" character ==> can't create share, can't read shares/links [FIXED] https://github.com/owncloud/ios-app/pull/358#issuecomment-489798177
- [X] (2) Glitches in card https://github.com/owncloud/ios-app/pull/358#issuecomment-496568569 [FIXED]
- [X] (3) Current collaborators' alert https://github.com/owncloud/ios-app/pull/358#issuecomment-496583272 [DONE]
- [X] (4) Recipient sees himself as collaborator https://github.com/owncloud/ios-app/pull/358#issuecomment-496597162  [FIXED]
- [X] (5) Alert of increasing permissions to be improved https://github.com/owncloud/ios-app/pull/358#issuecomment-496809395 [WONT FIX]
- [X] (6) Lack of edit permission in edition from files app https://github.com/owncloud/ios-app/pull/358#issuecomment-496811238 [MOVE TO CORE] [TO DO]
- [X] (7) Edit subpermisions revoked do not uncheck the Can Edit permission https://github.com/owncloud/ios-app/pull/358#issuecomment-496817986 [DONE]
- [x] (8) Race condition in permission switching https://github.com/owncloud/ios-app/pull/358#issuecomment-496880939 [FIXED]
- [X] (9) Unshare wording https://github.com/owncloud/ios-app/pull/358#issuecomment-496889104 -> [DONE]
- [X] (10) `@` character collides with federated sharing https://github.com/owncloud/ios-app/pull/358#issuecomment-496899425 [FIXED]
- [x] (11) SharingAPI capability keeps visible the sharing icons https://github.com/owncloud/ios-app/pull/358#issuecomment-496951170 [WONT FIX]
- [X] (12) Resharing capability should hide some UI components https://github.com/owncloud/ios-app/pull/358#issuecomment-496957163 [FIXED]
- [X] (13) Additional info below the file name in card https://github.com/owncloud/ios-app/pull/358#issuecomment-496972715 [FIXED]
- [X] (14) Message for not shared content https://github.com/owncloud/ios-app/pull/358#issuecomment-496992062 -> Included the option to invite
- [X] (15) Private link issues https://github.com/owncloud/ios-app/pull/358#issuecomment-497259653 [FIXED]
- [X] (16) Public link overlaps the button https://github.com/owncloud/ios-app/pull/358#issuecomment-497618707 [FIXED]
- [X] (17) Public link with no name makes the app crash https://github.com/owncloud/ios-app/pull/358#issuecomment-497620653  [FIXED]
- [X] (18) N2H: Keyboard to be hidden after typing password https://github.com/owncloud/ios-app/pull/358#issuecomment-497629012 [DONE] 
- [X] (19) Different names for public links https://github.com/owncloud/ios-app/pull/358#issuecomment-497630759 [DONE]
- [X] (20) Edit public link name is not posible https://github.com/owncloud/ios-app/pull/358#issuecomment-497632439 [FIXED] 
- [X] (21) Default expiration date not saved https://github.com/owncloud/ios-app/pull/358#issuecomment-497653084 [FIXED]
- [X] (22) Reaction when expiration date changes https://github.com/owncloud/ios-app/pull/358#issuecomment-498142129 [WONT FIX]
- [X] (23) Removal of expiration date https://github.com/owncloud/ios-app/pull/358#issuecomment-498145363 [FIXED] 
- [X] (24) Default expiration date is not set https://github.com/owncloud/ios-app/pull/358#issuecomment-498194753 [FIXED]
- [X] (25) Labelling "password" with "*" when it is mandatory https://github.com/owncloud/ios-app/pull/358#issuecomment-498195925 [DONE]
- [x] (26) Offline behaviour https://github.com/owncloud/ios-app/pull/358#issuecomment-498197477 [WONT FIX HERE]
- [X] (27) Some issues regarding #292 https://github.com/owncloud/ios-app/pull/358#issuecomment-498705578 [FIXED]
- [X] (28) Share permission in list with no Sharing allowed by capability https://github.com/owncloud/ios-app/pull/358#issuecomment-498748530 [FIXED]
- [X] (29) Declination in federated shares takes long and shows an incorrect notification value https://github.com/owncloud/ios-app/pull/358#issuecomment-498990842 [FIXED]
- [X] (30) Unshare in federation not working https://github.com/owncloud/ios-app/pull/358#issuecomment-498994980 [FIXED]
- [x] (31) Explanation tip in public links https://github.com/owncloud/ios-app/pull/358#issuecomment-499129185 [WONT FIX HERE]
- [X] (32) Multiple pending invitations https://github.com/owncloud/ios-app/pull/358#issuecomment-499394445 [FIXED] weird effect, not blocker
- [X] (33) Unsharing content inside a shared folder https://github.com/owncloud/ios-app/pull/358#issuecomment-499806459 [FIXED]
- [x] (34) No connection error message https://github.com/owncloud/ios-app/pull/358#issuecomment-499806824 [WONT FIX HERE]
- [X] (35) Default permissions are hidden instead of unchecked https://github.com/owncloud/ios-app/pull/358#issuecomment-499836346  [FIXED]


michaelstingl:
### (1)

1. Add account test:test@demo.owncloud.com
2. Create "New+Folder"
3. Action: Share this folder ==> focus not in search bar ??
4. Share with "admin" ==> Error: Adding User to Share failed (Wrong path, file/folder doesn't exist)

Encoding issue?

App version: 1.0 build 118 (https://github.com/owncloud/ios-app/commit/e4e2cf187f496b9da5415732c04a3752b261a5e6)
Devices: iPhone XS Max and iPad Pro 10.5
Server: 10.1.0

hosy:
> ### (1)
> 1. Add account test:[test@demo.owncloud.com](mailto:test@demo.owncloud.com)
> 2. Create "New+Folder"
> 3. Action: Share this folder ==> focus not in search bar ??
> 4. Share with "admin" ==> Error: Adding User to Share failed (Wrong path, file/folder doesn't exist)

Could not reproduce the error. Maybe item was not available online, when try to share!? 
Focus will be fixed!

hosy:
> > ### (1)
> > 
> > 1. Add account test:[test@demo.owncloud.com](mailto:test@demo.owncloud.com)
> > 2. Create "New+Folder"
> > 3. Action: Share this folder ==> focus not in search bar ??
> > 4. Share with "admin" ==> Error: Adding User to Share failed (Wrong path, file/folder doesn't exist)
> 
> Could not reproduce the error. Maybe item was not available online, when try to share!?
> Focus will be fixed!
@michaelstingl 
Bug is now fixed!

ownCloud/Client/Actions/Action.swift
```diff
@@ -281,4 +321,152 @@ class Action : NSObject {
...
+
+				viewController.present(navigationController, animated: true, completion: nil)
+			}
+		}, title: title, style: .plain, image: UIImage(named: "group"), alignment: .left, identifier: "share-add-group")
+
+		return addGroupRow
+	}
+
+	class func shareAsPublicLinkRow(item : OCItem, presentingController: UIViewController, context: ActionContext) -> StaticTableViewRow? {
+		if context.core!.connection.capabilities?.publicSharingEnabled == true, item.sharedByPublicLink() == false, item.isShareable {
```
felix-schwarz: Comparisons like `== true` and `== false` are most useful when you deal with an Optional (`Bool?`) value. When dealing with a plain `Bool`, `!value` or `value` is often preferable:
```suggestion
		if context.core!.connection.capabilities?.publicSharingEnabled == true, !item.sharedByPublicLink(), item.isShareable {
```

ownCloud/Client/Actions/Action.swift
```diff
@@ -281,4 +321,152 @@ class Action : NSObject {
 	var position : ActionPosition {
 		return type(of: self).applicablePosition(forContext: context)
 	}
+
+	// MARK: - Sharing
```
felix-schwarz: Please put the Sharing methods into an extension to the `Action` class:
```suggestion
}

// MARK: - Sharing
extension Action {
```

ownCloud/SDK Extensions/OCCore+Extension.swift
```diff
@@ -0,0 +1,74 @@
...
+			let shares = query.queryResults.filter({ (share) -> Bool in
+				if share.itemPath == item.path {
+					return true
+				}
+				return false
+			})
+			initialPopulationHandler(shares)
+		}
+
+		return shareQuery
```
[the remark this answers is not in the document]
hosy: Should I stop it on the call site like so?
`			let sharedWithMeQuery = self.core?.sharesSharedWithMe(for: self.item, initialPopulationHandler: { (sharesWithMe) in
				OnMainThread {
					if sharesWithMe.count > 0 {
						var shares : [OCShare] = []
						shares.append(contentsOf: sharesWithMe)
						shares.append(contentsOf: sharesWithReshares)
						self.shares = shares
						self.removeShareSections()
						self.addShareSections()
					}

					self.addActionShareSection()
				}
			})
			if let query = sharedWithMeQuery {
				self.core?.stop(query)
			}`

ownCloud/Client/Actions/Action.swift
```diff
@@ -259,10 +293,16 @@ class Action : NSObject {
...
 	func provideAlertAction() -> UIAlertAction? {
-		return UIAlertAction(title: self.actionExtension.name, style: actionExtension.category == .destructive ? .destructive : .default, handler: { (_ alertAction) in
+		let alertAction = UIAlertAction(title: self.actionExtension.name, style: actionExtension.category == .destructive ? .destructive : .default, handler: { (_ alertAction) in
 			self.willRun()
 			self.run()
 		})
+
+		let image = self.actionExtension.image
+		alertAction.setValue(image, forKey: "image")
+		alertAction.setValue(CATextLayerAlignmentMode.left, forKey: "titleTextAlignment")
```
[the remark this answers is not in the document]
hosy: This is not working and it does not respond to the given selector.

wuenschedesign:
hereby an improved draft for the iOS App. 

Accounts | Filelist
-- | --
ownCloud-iOS-Account-List | ownCloud-iOS-Startscreen

ownCloud/SDK Extensions/OCCore+Extension.swift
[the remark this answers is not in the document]
felix-schwarz: That could stop it too soon.

If - as it looks right now - you don't really need the `OCShareQuery` returned from `sharesSharedWithMe(for:…)`, but just are interested in the initial results, I'd suggest extending the relevant methods with a `keepRunning` parameter, which controls if the query should be stopped right inside the `initialPopulationHandler`. The code would then look like this:

```swift
@discardableResult func sharesSharedWithMe(for item: OCItem, initialPopulationHandler: @escaping (_ shares: [OCShare]) -> Void, keepRunning: Bool = false) -> OCShareQuery? {
	let shareQuery = OCShareQuery(scope: .sharedWithUser, item: item)
	start(shareQuery!)
	shareQuery?.initialPopulationHandler = { [weak self] query in
		let shares = query.queryResults.filter({ (share) -> Bool in
			if share.itemPath == item.path {
				return true
			}
			return false
		})
		initialPopulationHandler(shares)

		if !keepRunning {
			self?.stop(query)
		}
	}

	return keepRunning ? shareQuery : nil
}

@discardableResult func sharesWithReshares(for item: OCItem, initialPopulationHandler: @escaping (_ shares: [OCShare]) -> Void, keepRunning: Bool) -> OCShareQuery? {
	let shareQuery = OCShareQuery(scope: .itemWithReshares, item: item)
	start(shareQuery!)
	shareQuery?.initialPopulationHandler = { query in
		initialPopulationHandler(query.queryResults)

		if !keepRunning {
			self?.stop(query)
		}
	}

	return keepRunning ? shareQuery : nil
}
```

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
```diff
@@ -0,0 +1,465 @@
...
+				var title = ""
+				if recipient.type == .user {
+					guard let displayName = recipient.displayName else { continue }
+					title = displayName
+				} else {
+					guard let displayName = recipient.displayName else { continue }
+					let groupTitle = "(Group)".localized
+					title = "\(displayName) \(groupTitle)"
+				}
+
```
felix-schwarz: From the department of nice-to-have-but-not-required: would be awesome an icon representing a `user` or `group` would be in front of each line in the search results.

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
```diff
@@ -0,0 +1,465 @@
...
+		} else if searchController.isActive {
+			resetTable(showShares: false)
+		}
+	}
+
+	func searchBarCancelButtonClicked(_ searchBar: UISearchBar) {
+		self.resetTable(showShares: true)
+		self.searchController?.searchBar.isLoading = false
+	}
+
```
felix-schwarz: Adding this code snippet will pre-populate with suggestions from the server when the search bar is focused the first time:
```swift

	func searchBarShouldBeginEditing(_ searchBar: UISearchBar) -> Bool {
		if recipientSearchController?.recipients == nil, recipientSearchController?.searchTerm == nil {
			recipientSearchController?.search()
		}
		return true
	}
```
Bildschirmfoto 2019-05-17 um 16 09 45

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
```diff
@@ -0,0 +1,465 @@
...
+				return false
+			}
+			self.shares = sharesWithReshares
+			OnMainThread {
+				self.removeShareSections()
+				self.addShareSections()
+
+				self.addActionShareSection()
+			}
+		}
```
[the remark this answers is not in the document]
hosy: @michaelstingl do not want this and I removed setting focus to the search field.

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	// MARK: - UISearchResultsUpdating Delegate
+
+	func updateSearchResults(for searchController: UISearchController) {
+		guard let text = searchController.searchBar.text else { return }
+		if text.count > core?.connection.capabilities?.sharingSearchMinLength?.intValue ?? 1 {
+			recipientSearchController?.searchTerm = text
+		} else if searchController.isActive {
+			resetTable(showShares: false)
```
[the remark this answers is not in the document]
hosy: I used the second option, because showing default suggestions my not be wanted.

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
hosy: @felix-schwarz I have implemented the `OCUser.avatar` but this is always nil. I set the profile picture in the settings for the user.

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
hosy: This may be not wanted.

ownCloud/Client/Actions/Action.swift
```diff
@@ -162,11 +162,42 @@ class Action : NSObject {
...
+					progressView.startAnimating()
+
+					let row = StaticTableViewRow(rowWithAction: nil, title: "Searching Shares...".localized, alignment: .left, accessoryView: progressView, identifier: "share-searching")
+					self.updateSharingSection(sectionIdentifier: "share-section", rows: [row], tableViewController: tableViewController)
+
+					core.unifiedShares(for: item, completionHandler: { (shares) in
+						OnMainThread {
+							let shareRows = self.shareRows(shares: shares, item: item, presentingController: moreViewController, context: context)
+							self.updateSharingSection(sectionIdentifier: "share-section", rows: shareRows, tableViewController: tableViewController)
+							moreViewController.preferredContentSize = tableViewController.tableView.contentSize
```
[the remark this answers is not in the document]
hosy: Could not reproduce this problem. I had this, and I solved this setting the new tableView height, after building the UI: `moreViewController.preferredContentSize = tableViewController.tableView.contentSize`

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
```diff
@@ -0,0 +1,473 @@
...
+	@objc func dismissView() {
+		if let query = self.shareQuery {
+			self.core?.stop(query)
+		}
+		dismissAnimated()
+	}
+
+	// MARK: - Header View
+
+	func addHeaderView() {
```
felix-schwarz: This code is shared between `GroupSharingTableViewController` and `PublicLinkTableViewController`. Maybe create a parent class to move this code to, so it can be reused in both.

ownCloud/Client/Actions/Action.swift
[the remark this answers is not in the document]
hosy: please try again, I set now a placeholder row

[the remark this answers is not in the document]
felix-schwarz: Fixed, thanks!

ownCloud/Library/CustomFilelistTableViewController.swift
```diff
@@ -0,0 +1,381 @@
...
+				UserDefaults.standard.setValue(newValue.rawValue, forKey: "sort-method")
+			}
+
+			get {
+				let sort = SortMethod(rawValue: UserDefaults.standard.integer(forKey: "sort-method")) ?? SortMethod.alphabeticallyAscendant
+				return sort
+			}
+		}
+
+		func makeActionProgressHandler() -> ActionProgressHandler {
```
felix-schwarz: `CustomFilelistTableViewController`, `SharesFilelistTableViewController` and `ClientQueryViewController` share quite a bit of code.

Would be great to either them `ClientQueryViewController` subclasses or move the shared code to a shared superclass, so any future additions or fixes in the shared code benefits all three at once.

ownCloud/Client/ClientItemCell.swift
```diff
@@ -139,7 +159,9 @@ class ClientItemCell: ThemeTableViewCell {
 			iconView.topAnchor.constraint(equalTo: self.contentView.topAnchor, constant: verticalIconMargin),
 			iconView.bottomAnchor.constraint(equalTo: self.contentView.bottomAnchor, constant: -verticalIconMargin),
 
-			titleLabel.rightAnchor.constraint(equalTo: cloudStatusIconView.leftAnchor, constant: -horizontalSmallMargin),
+			titleLabel.rightAnchor.constraint(equalTo: sharedStatusIconView.leftAnchor, constant: -horizontalSmallMargin),
+			sharedStatusIconView.rightAnchor.constraint(equalTo: publicLinkStatusIconView.leftAnchor, constant: -horizontalSmallMargin),
+			publicLinkStatusIconView.rightAnchor.constraint(equalTo: cloudStatusIconView.leftAnchor, constant: -horizontalSmallMargin),
```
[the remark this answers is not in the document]
hosy: Please can you have a look at my ClientItemCell.swift implementation. I think there is something nor correct and a constraint for `publicLinkStatusIconView` is missing, it this image is nil.

hosy:
@michaelstingl should every `Private Link` title removed with `Link`?

michaelstingl:
@hosy updated 3rd screenshot in the 2nd row. Also for public link creation, a confirmation would be preferred.

ownCloud/Client/ClientItemCell.swift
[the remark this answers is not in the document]
felix-schwarz: I think you have accidentally picked the wrong constraint to control the constant of. These control the width (A):
```swift
sharedStatusIconViewWidthConstraint = sharedStatusIconView.widthAnchor.constraint(equalToConstant: 0)
publicLinkStatusIconViewWidthConstraint = publicLinkStatusIconView.widthAnchor.constraint(equalToConstant: 0)
```

But the width is already automatically controlled by setting an image or `nil` via (B):
```swift
sharedStatusIconView.setContentHuggingPriority(.required, for: .vertical)
sharedStatusIconView.setContentHuggingPriority(.required, for: .horizontal)
sharedStatusIconView.setContentCompressionResistancePriority(.required, for: .vertical)
sharedStatusIconView.setContentCompressionResistancePriority(.required, for: .horizontal)

publicLinkStatusIconView.setContentHuggingPriority(.required, for: .vertical)
publicLinkStatusIconView.setContentHuggingPriority(.required, for: .horizontal)
publicLinkStatusIconView.setContentCompressionResistancePriority(.required, for: .vertical)
publicLinkStatusIconView.setContentCompressionResistancePriority(.required, for: .horizontal)
```

Thanks to (B), you don't need to take care of the width. What needs to be taken care of is the margin to the right of the image views, which doesn't collapse to zero automatically.

To take care of this, remove (A), keep (B) and put these two constraints into f.ex. `sharedStatusIconViewRightMarginConstraint` and `publicLinkStatusIconViewRightMarginConstraint` variables instead (C):
```swift
sharedStatusIconView.rightAnchor.constraint(equalTo: publicLinkStatusIconView.leftAnchor, constant: 0)
publicLinkStatusIconView.rightAnchor.constraint(equalTo: cloudStatusIconView.leftAnchor, constant: 0)
```

Then, change the constant value between `0` and `-horizontalSmallMargin` as needed.

I also noticed a couple of unneeded / duplicate / conflicting constraints that can be removed without replacement:
```swift
// Duplicate (and now in (C))
sharedStatusIconView.rightAnchor.constraint(equalTo: publicLinkStatusIconView.leftAnchor, constant: -horizontalSmallMargin),
publicLinkStatusIconView.rightAnchor.constraint(equalTo: cloudStatusIconView.leftAnchor, constant: -horizontalSmallMargin),

// Conflicting / not needed
sharedStatusIconView.rightAnchor.constraint(lessThanOrEqualTo: publicLinkStatusIconView.leftAnchor, constant: -horizontalSmallMargin),
publicLinkStatusIconView.rightAnchor.constraint(lessThanOrEqualTo: cloudStatusIconView.leftAnchor, constant: -horizontalSmallMargin),
```

I've tested this and put the resulting `ClientItemCell.swift` with all changes applied up at https://gist.github.com/felix-schwarz/7ff0ebb1258b03dfe93d2a53e4fa7082

Also noticed Xcode complaining about a missing constraint in `UITableViewController+Extension.swift`:

```swift
coloredView.leftAnchor.constraint(equalTo: self.tableView.leftAnchor),
```

Put the changed version of the file into that gist, too.

ownCloud/Client/Sharing/PublicLinkTableViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+			var rows : [StaticTableViewRow] = []
+
+			self.core?.retrievePrivateLink(for: self.item, completionHandler: { (error, url) in
+
+				guard let url = url else { return }
+				if error == nil {
+					OnMainThread {
+						let privateLinkRow = StaticTableViewRow(headerRowWithAction: { [weak self] (row, _) in
+							guard let self = self else { return }
+							self.retrievePrivateLink(for: self.item, in: row)
```
felix-schwarz: At this point, the private link URL is already known. As far as I know that URL doesn't change, so it shouldn't be necessary to retrieve it again in `retrievePrivateLink(for:in:)`. Instead, the private link URL can be "saved" in this block and copied from clipboard when run.

ownCloud/Client/Sharing/PublicLinkTableViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+			OnMainThread {
+				self.resetTable(showShares: false)
+			}
+		}
+	}
+
+	// MARK: - Private Link Section
+
+	func addPrivateLinkSection() {
+		let identifier = "private-link-section"
```
felix-schwarz: Design-wise, I'd prefer this section to have a title "Private Link", the private link URL to be shown below it as regular row. And then "Copy" as a button on the right side in the same row. Feels a bit busy right now.

ownCloud/Client/Sharing/PublicLinkTableViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+						self.navigationController?.pushViewController(editPublicLinkViewController, animated: true)
+					}, title: share.name!, subtitle: share.permissionDescription(), accessoryType: .disclosureIndicator) )
+				} else {
+					shareRows.append( StaticTableViewRow(rowWithAction: nil, title: share.name!, subtitle: share.permissionDescription(), accessoryType: .none) )
+				}
+			}
+		} else {
+			shareRows.append( StaticTableViewRow(buttonWithAction: { [weak self] (_, _) in
+				self?.addPublicLink()
+			}, title: "Create Public Link".localized, style: StaticTableViewRowButtonStyle.plain))
```
felix-schwarz: I understand the idea of only showing the "Create Public Link" button when there are no public links created yet. It took me a while to find the "+" in the upper-left corner, though. In fact, I only recognized it when I already wanted to dismiss the view controller.

So, from a UX perspective, having this "button" always available may be preferable.

ownCloud/Client/Sharing/PublicLinkEditTableViewController.swift
```diff
@@ -0,0 +1,590 @@
...
+			footer = String(format: "Shared since: %@".localized, dateFormatter.string(from: date))
+		}
+
+		let deleteSection = StaticTableViewSection(headerTitle: nil, footerTitle: footer)
+
+		if let share = self.share, let shareURL = share.url {
+			deleteSection.add(rows: [
+				StaticTableViewRow(headerRowWithAction: { (_, _) in
+					UIPasteboard.general.url = shareURL
+				}, headerTitle: String(format:"%@", shareURL.absoluteString), title: "Copy Public Link".localized, accessoryView: .none)
```
felix-schwarz: Would IMO look better and make it less busy to
- have that section with a header "Public Link"
- put the link url into a regular row
- have a "Copy" button on the right side of that row.
- have the delete row just labelled "Delete" then

ownCloud/Client/Sharing/GroupSharingEditTableViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+						if enabled {
+							share.permissions.insert(permissionValue)
+						} else {
+							share.permissions.remove(permissionValue)
+						}
+					}
+				}, completionHandler: { (error, share) in
+					if error == nil {
+						guard let changedShare = share else { return }
+						self.share?.permissions = changedShare.permissions
```
felix-schwarz: After completing an update, the SDK fetches the shares metadata anew, creating a completely new share that exactly represents the share on the server.

IMO it'd therefore be better to use that `OCShare` instance and set it as the view controller's `share`. A `didSet` block on the view controller's `share` could then update the view controller as needed to match the updated share.

I write this only once here, but it also applies to all similar uses of `core.update(share, afterPerformingChanegs:,completionHandler:)` in this view controller and also `PublicLinkEditTableViewController`.

ownCloud/Client/Sharing/GroupSharingTableViewController.swift
```diff
@@ -0,0 +1,462 @@
...
+			var rows : [StaticTableViewRow] = []
+			for recipient in recipients {
+
+				guard let itemPath = self.item.path else { continue }
+				var title = ""
+				var image: UIImage?
+				if recipient.type == .user {
+					guard let displayName = recipient.displayName else { continue }
+					title = displayName
+					image = UIImage(named: "person")
```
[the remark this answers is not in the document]
hosy: Fixed

ownCloud/Client/Actions/Action.swift
```diff
@@ -247,7 +277,7 @@ class Action : NSObject {
 		return StaticTableViewRow(buttonWithAction: { (_ row, _ sender) in
 			self.willRun()
 			self.run()
-		}, title: actionExtension.name, style: actionExtension.category == .destructive ? .destructive : .plain, identifier: actionExtension.identifier.rawValue)
+		}, title: actionExtension.name, style: actionExtension.category == .destructive ? .destructive : .plain, image: self.icon, alignment: .left, identifier: actionExtension.identifier.rawValue)
```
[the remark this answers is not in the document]
hosy: Cannot get it working for .default style, because setting constraints are not working. Please give me a hint or do I need a subclass?

ownCloud/Client/Sharing/GroupSharingEditTableViewController.swift
hosy: IMP would should move this to later, because this needs more time to reorganize the current view.

ownCloud/Client/Sharing/GroupSharingEditUserGroupsTableViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+		guard let share = share else { return }
+
+		var canEdit = false
+		if !createShare, share.canUpdate || share.canCreate || share.canDelete {
+			canEdit = true
+		}
+		var canShare = false
+		if !createShare, share.canShare {
+			canShare = true
+		}
```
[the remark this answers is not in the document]
hosy: Fixed!

ownCloud/Client/Sharing/GroupSharingEditTableViewController.swift
felix-schwarz: Agreed!

ownCloud/Client/Actions/Action.swift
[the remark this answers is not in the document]
felix-schwarz: Couldn't get to work this properly, either, so I implemented a different approach: adding some padding to the images so they all have the same width. Also noticed that `Rename` was localized as ` Rename` (leading space) and fixed that.

jesmrec:
### (1)

A couple of small glitches in the card:

1- If you swipe the card to the top, the thumbnail is a bit displaced to the top, and the `Delete` option to the bottom. By releasing the finger, the `Delete` option is hidden 

May-28-2019 17-29-55

2- In landscape, by swiping the options, the `Actions` title overlaps with the options (it shouldn't):

May-28-2019 17-32-04

iPhone 6Plus v12
Server 10.1

jesmrec:
### (3) (improvement)

1. Invite a collaborator
2. Invite again the same collaborator
3. Tap on Save

Current: Error is correctly displayed, but the permission view stays. The only way to dismiss the view is tapping on  `Cancel`

Expected: some options:

- Current collaborators are not included in the matches
- If a current collaborator is selected from the matches, the alert is displayed instead the permissions view, staying in collaborators view
- With the same flow, the alert is shown in permissions view. When it is "OK", the view is dismissed.

jesmrec:
### (4)

1. "User1" invites collaborator "User2" of any item.
2 "User2" opens collaborators view of the shared item

Current: "User2" sees "User2" as collaborator
Expected: "User2" sees only those collaborators he has shared with (web UI behaviour)

jesmrec:
### (5)

1. "User1" invites "User2" with no Edit permission granted over one item
2. "User2" invites (reshares) "User2" over the same item, giving Edit permission:

Current: The following alert is displayed:

Screenshot 2019-05-29 at 08 49 39

- The file path should only include the file name, not the whole path
- Wording does not pass with the feature: ´Collaborator` instead of `User`,  `Invite` instead of `Share` @michaelstingl @pmaier1 ?

jesmrec:
### (6)

1. "User1" invites "User2" with no Edit permission granted over one item
2. "User2" edit the file in Files App

Current: File is updated locally. It is intended to be uploaded to the server, but the lack of permissions makes the spinner in file list to be there forever.

Expected: User is notified the issue in any way (alert view, or directly dismiss the upload)

jesmrec:
### (7) (improvement)

1. Invite a collaborator over a folder
2. Revoke create, change, delete permissions

Current:

Screenshot 2019-05-29 at 09 21 06

Expected: 

When three edit options are removed, the whole section is vanished (`Can Edit` unchecked')

Screenshot 2019-05-29 at 09 24 24

hosy:
@jesmrec (2) solved

hosy:
> ### (4)
> 1. "User1" invites collaborator "User2" of any item.
> 2. "User2" opens collaborators view of the shared item
> 
> Current: "User2" sees "User2" as collaborator
> Expected: "User2" sees only those collaborators he has shared with (web UI behaviour)

@jesmrec 
I do not agree, because as "User2" I want to see which permission I have for this share. If I remove "User2" from this view, I cannot see my permissions.

hosy:
> ### (5)
> 1. "User1" invites "User2" with no Edit permission granted over one item
> 2. "User2" invites (reshares) "User2" over the same item, giving Edit permission:
> 
> Current: The following alert is displayed:
> 
> Screenshot 2019-05-29 at 08 49 39
> 
> * The file path should only include the file name, not the whole path
> * Wording does not pass with the feature: ´Collaborator`instead of`User`,  `Invite`instead of`Share` @michaelstingl @pmaier1 ?

@jesmrec resharing to the same user is no longer possible (because of (4) solved). Alert wording will returned from server.

hosy:
@jesmrec (7) solved

jesmrec:
> @jesmrec resharing to the same user is no longer possible (because of (4) solved)

sorry, i meant "User2 shares with User3". The point is that the item owner shares with no re-share permission to a third one.

jesmrec:
### (8)

Editing share permissions should be synchronous to avoid bad behaviours. Check these steps:

1. Create a share over a folder with all permissions (31)
2. Enter to edit the share
3. Disable Delete and Change (in this order) with two taps and go back by tapping "Collaborations"

Current behaviour:

Read, Share, Create and Delete permissions are displayed 

Expected:

only Read, Share and Create  are displayed

This happens because every tap triggers an asynchronous PUT request 

First tap (disable "Delete" permission) triggers a PUT with permissions = 23
Second tap (disable "Change" permission) triggers a PUT with permissions = 29 (delete enabled)

so, a race condition happens. One option is showing a spinner after a permission switching, that will make waiting till the operation/request is fully completed.

jesmrec:
### (9)

Justo for being in the same page of the web UI, if an user is the recipient, the option "Delete"of the item turns on "Unshare":

Web UI:

Screenshot 2019-05-29 at 12 50 07

App:

Screenshot 2019-05-29 at 12 59 31

jesmrec:
### (10)

Usernames can start with '@' character, so it collides with the federate sharing flow

1. Create a user which username is  `@user`
2. Share something with such user

Current: After saving, an error is displayed `Invalid federated Cloud ID`
Expected: Shared with such user

Just an idea: in the Android client, in case an username containing the `@` character, we show two options: one for the user itself, another one for remote sharing option:

Screenshot 2019-05-29 at 13 35 07

hosy:
@jesmrec (8) solved

jesmrec:
### (11)

In case the capability `api_enabled` is false, the options in the card are disabled. But, you still see the sharing icons in the file list. 

Would be a good idea to remove them in case no sharing is allowed by capability? Both public and private shares are not reachable for no one.

CC @michaelstingl

jesmrec:
### (12) (improvement)

With `resharing`capability in  false:

- Option `Can Share` in `Invite Collaborators` should be hidden and unchecked. Recipient will not have the option to share.
- When recipient opens the "Shared by ..." view, the searching bar is not useful since he is not allowed to reshare. I'd hide it as the `Links` option in the card is.

michaelstingl:
### (9)

@jesmrec @hosy wording is still not finalized yet. (Low priority for now)

jesmrec:
### (14) (improvement)

Invite collaborators over an item with no collaborations yet. This is the view:

Screenshot 2019-05-29 at 17 39 23

i'd set there some kind of message like 'No Collaborators` or similar, as the one it is used for empty folders or empty searchs. What do you think?

hosy:
> ### (14) (improvement)
> Invite collaborators over an item with no collaborations yet. This is the view:
> 
> Screenshot 2019-05-29 at 17 39 23
> 
> i'd set there some kind of message like 'No Collaborators` or similar, as the one it is used for empty folders or empty searchs. What do you think?

@jesmrec yes, we discussed this today. There will be added some infos what to do. Thank you!

hosy:
@jesmrec (12), (13) fixed

hosy:
> ### (10)
> Usernames can start with '@' character, so it collides with the federate sharing flow
> 
> 1. Create a user which username is  `@user`
> 2. Share something with such user
> 
> Current: After saving, an error is displayed `Invalid federated Cloud ID`
> Expected: Shared with such user
> 
> Just an idea: in the Android client, in case an username containing the `@` character, we show two options: one for the user itself, another one for remote sharing option:
> 
> Screenshot 2019-05-29 at 13 35 07

@felix-schwarz is this SDK or server api related?

hosy:
@jesmrec (14) fixed

jesmrec:
With commit `1bce17044e55f01a6a968cd9cc5be98fa6ec64a6`

(1), (2), (3), (7), (9), (13), (14) -> Fixed / Done

(4), (5) -> ellaborated next message

(8) -> still reproducible

(12) -> Option to share is not displayed in permissions

Screenshot 2019-05-30 at 10 02 43

but, it is in the list of collaborators after saving

Screenshot 2019-05-30 at 10 02 56

> When recipient opens the "Shared by ..." view, the searching bar is not useful since he is not allowed to reshare. I'd hide it as the Links option in the card is.

still there

jesmrec:
(4)

> @jesmrec
I do not agree, because as "User2" I want to see which permission I have for this share. If I remove "User2" from this view, I cannot see my permissions.

the point is if the recipient should know its own permissions (neither web UI nor other clients show such info). Also, the row cell is in the same list as the reshared users, making there a mixture that could be confusing.

Also i noticed a bad behaviour related:

When the recipient reshares with a third user, the third user is set in the first position of the list, overlapped with the recipient that was there before:

May-30-2019 10-22-47

When deleting collaborator, happens the same (bad list handling)

hosy:
@jesmrec please can you add a table of contents to the initial PR comment, because we are loosing the overview. To mich comments here or you can hide fixed comments.

jesmrec:
### (16)

About private link

If the URL is long, the link overlaps with the `Copy` button:

Screenshot 2019-05-30 at 10 56 39

On the other hand, is it really useful to display the link URL? Maybe setting somewhere a representative icon, and by tapping on it the private link will be copied to the clipboard.

jesmrec:
> @jesmrec please can you add a table of contents to the initial PR comment, because we are loosing the overview. To mich comments here or you can hide fixed comments.

in the fist comment there is a list to check the status of every report

- [ ] -> Not finished, something to do
- [X] -> Done/Fixed/Finished

michaelstingl:
> On the other hand, is it really useful to display the link URL?

Requested by @hodyroff

jesmrec:
### (16)

Same as (15) for the public links

Screenshot 2019-05-31 at 10 09 12

jesmrec:
### (17)

Create a public link with empty name

Current: app crashes
Expected: link created with default name

iOS12.2
iPhone 6Plus

jesmrec:
### (18)  (nice to have)

1. Create public link
2. Enable password and set 
3. Enable Expiration 

Soft keyboard overlaps the date picker. It is posible to swipe the view up, but would be better if the keyboard is vanished in any way.  

May-31-2019 10-31-58

Just as a idea: the old iOS app has a "Done" option just above the keyboard, that hides it after finishing typing

Screenshot 2019-05-31 at 10 42 16

jesmrec:
### (19) (improvement)

1. Create two public links over the same item, keeping the defaults

Current: Both links will have the same name
Expected: web UI add automatically  number to distinguish, like:

Screenshot 2019-05-31 at 10 48 45

jesmrec:
### (20)

1. Create a public link
2. Try to edit the public link name

Current: No way to submit the change
Expected: Button, icon, "Save" option, or even automatically after some seconds without typing

hodyroff:
> > On the other hand, is it really useful to display the link URL?
> 
> Requested by @hodyroff

Thats not a good argument.
IMHO it is useful as I like to see the link I am going to send to somebody when its created.
Just to make sure its not telling - eg.
Just for the personal comfort of the creator ... 
Of course it should be possible to see the full link or the beginning and the end if it is a space challenge but that can be treated like any name view.

jesmrec:
> > > On the other hand, is it really useful to display the link URL?
> > 
> > 
> > Requested by @hodyroff
> 
> Thats not a good argument.
> IMHO it is useful as I like to see the link I am going to send to somebody when its created.
> Just to make sure its not telling - eg.
> Just for the personal comfort of the creator ...
> Of course it should be possible to see the full link or the beginning and the end if it is a space challenge but that can be treated like any name view.

This is not a matter of space. IHMO, the link itself is ugly to see and does not give extra info since the only action you can do with it is copying. I'd hide it with an eye toggle (like the one used in password fields, if technically posible), just in case the owner to be sure about what it is being sent.

jesmrec:
### (21)

1. Create a public link
2. Enable Expiration date, but let the default date
3. Save

Current: Expiration date not saved
Expected: Expiration date saved.

hodyroff:
> > > > On the other hand, is it really useful to display the link URL?
> > > 
> > > 
> > > Requested by @hodyroff
> > 
> > 
> > Thats not a good argument.
> > IMHO it is useful as I like to see the link I am going to send to somebody when its created.
> > Just to make sure its not telling - eg.
> > Just for the personal comfort of the creator ...
> > Of course it should be possible to see the full link or the beginning and the end if it is a space challenge but that can be treated like any name view.
> 
> This is not a matter of space. IHMO, the link itself is ugly to see and does not give extra info since the only action you can do with it is copying. I'd hide it with an eye toggle (like the one used in password fields, if technically posible), just in case the owner to be sure about what it is being sent.

Yes, eye toggle is perfect. Just some way for the user to inspect the URL if wanted.

michaelstingl:
> Yes, eye toggle is perfect.

Not sure if eye-toggle for URL feels iOS native, but @hosy you get the idea… Otherwise keep it visible and only take care about the copy-button-collision.

hosy:
> > Yes, eye toggle is perfect.
> 
> Not sure if eye-toggle for URL feels iOS native, but @hosy you get the idea… Otherwise keep it visible and only take care about the copy-button-collision.

@jesmrec @hodyroff 
I would suggest to show only a `Copy Private Link` Button and if you touch on it, the URL gets copied and you will see the URL for e.g. 3 seconds instead of the `Copy Private Link` Button on the same place. After 3 seconds you will see the `Copy Private Link` Button again.

hodyroff:
> > > Yes, eye toggle is perfect.
> > 
> > 
> > Not sure if eye-toggle for URL feels iOS native, but @hosy you get the idea… Otherwise keep it visible and only take care about the copy-button-collision.
> 
> @jesmrec @hodyroff
> I would suggest to show only a `Copy Private Link` Button and if you touch on it, the URL gets copied and you will see the URL for e.g. 3 seconds instead of the `Copy Private Link` Button on the same place. After 3 seconds you will see the `Copy Private Link` Button again.

Could work for me. Same for public link, please.

jesmrec:
> @jesmrec @hodyroff
I would suggest to show only a Copy Private Link Button and if you touch on it, the URL gets copied and you will see the URL for e.g. 3 seconds instead of the Copy Private Link Button on the same place. After 3 seconds you will see the Copy Private Link Button again.

works for me!

jesmrec:
### (22) 

UI suggestion

1. Create a public link
2. Edit the public link by changing the expiration date. 

The expiration date is correctly changed, but there is no indicator about it. A regular user changes the date but there is no way to know if the change was submitted. Every user action should have a reaction.

jesmrec:
### (23)

1. Create a public link with expiration date (in app or web UI)
2. Edit the link in the app, removing the expiration date.

Current: Expiration date is not removed
Expected: Expiration date removed

pmaier1:
Regarding https://github.com/owncloud/ios-app/pull/358#issuecomment-496899425:

- We could add some more validation for that the "remote" entry will only be shown once a user types a compliant federation ID, like we do for Guests. This would solve this specific problem but once a user enters a whole e.g. mail address we still need a solution to not confuse the user. Long standing problem, discussed also here, for instance, https://github.com/owncloud/enterprise/issues/2344.
Such validation should happen on the server side / API, right?
- The "remote" entry has been renamed to "federated" in the meanwhile

jesmrec:
### (24)

1. Enable capability "Set default expiration date" with 30 days (as example)
2. In the app, create a public link and enable expiration date.

Current: tomorrow as default
Expected: 30 days after today

jesmrec:
### (25) (improvement)

In case the capability "Enforce password" is enabled, it would be handy to show the label "Password" with an '*' to mark this is mandatory.

jesmrec:
### (26)

Related with offline behaviour

If the device is offline, no way to check shares. Is this expected behaviour?

hosy:
> ### (26)
> Related with offline behaviour
> 
> If the device is offline, no way to check shares. Is this expected behaviour?

Yes, because sharing informations are only available if server is reachable.

hosy:
> ### (25) (improvement)
> In case the capability "Enforce password" is enabled, it would be handy to show the label "Password" with an '*' to mark this is mandatory.

A better solution would be to disable the password switch and set it to state on.

jesmrec:
> A better solution would be to disable the password switch and set it to state on.

or both, the '*' always clarify the mandatory condition

jesmrec:
### (27)

Regarding https://github.com/owncloud/ios-app/issues/292 (fix included here)

- Font color inside the searching textview is not visible in dark and classic themes

Before:

Screenshot 2019-06-04 at 16 40 54

Now:

Screenshot 2019-06-04 at 16 44 32

- In landscape mode, the text `No matches`  is not visible 

Screenshot 2019-06-04 at 16 45 16

- also in landscape, the view is scrollable and makes some weird effect (more visible in ownCloud classic theme)

Jun-04-2019 16-47-50

jesmrec:
### (28)

This was reported here: https://github.com/owncloud/ios-app/pull/358#issuecomment-497245346, but not for losing, here it is:

1. Disable "Allow resharing" capability
2. Share with another user, share option is not displayed (correct):

https://user-images.githubusercontent.com/14894746/58618135-22c34680-82c2-11e9-829f-850a62dc750e.png

3. Save the share

Current:  Share option is in the permissions list:

https://user-images.githubusercontent.com/14894746/58618233-5e5e1080-82c2-11e9-9155-9530ff40376c.png

Expected: Share is not in the permissions list

jesmrec:
Rechecked the [READY TO TEST] issues with commit `f115830d022575f2a152d9dc75838e28020556c7`, many of them are already fixed (check first message), but the following ones:

- (8): Still reproducible:

Jun-04-2019 17-53-48

- (17): Crash fixed but a little difference with web UI. If name is removed during creation, in the list is show the hash code of the public link:

Screenshot 2019-06-04 at 18 23 42

in the app, no name is shown:

Screenshot 2019-06-04 at 18 25 10

- (19) Not very important, but just to know: with the same link name, web UI starts the numeration in the second link:

Public Link
Public Link (1)
Public Link (2)
etc...

Current behaviour starts with (1) from the first:

Public Link (1)
Public Link (2)
Public Link (3)
etc...

this should not be a blocker and it is not a must to fix, can be addressed separately. In fact, (19) has been marked as [DONE]. Just a consideration.

- (24): still reproducible

Today: 4th June
Expiration: 30 days (links will expire on 4th July):

Jun-04-2019 18-17-33

- (25): still reproducible. With the password enforced the switch is visible, and there is no mark of mandatoriness:

Jun-04-2019 18-20-44

hosy:
> ### (22)
> UI suggestion
> 
> 1. Create a public link
> 2. Edit the public link by changing the expiration date.
> 
> The expiration date is correctly changed, but there is no indicator about it. A regular user changes the date but there is no way to know if the change was submitted. Every user action should have a reaction.
> 
> Same for password

For me, this is a won't fix, because we are using the iOS behavior. If there is something wrong, we will show an error. If everything is fine, we doing nothing.

hosy:
> > A better solution would be to disable the password switch and set it to state on.
> 
> or both, the '*' always clarify the mandatory condition

This is also a won't fix. If we would do this, we have to add an '*' to every text field and option. Password field is displayed without the switch, so you have to enter something. If you don't enter anything, you will get an error.

jesmrec:
> This is also a won't fix. If we would do this, we have to add an '*' to every text field and option. Password field is displayed without the switch, so you have to enter something. If you don't enter anything, you will get an error.

the switch is visible with the password enforced

hosy:
> > This is also a won't fix. If we would do this, we have to add an '*' to every text field and option. Password field is displayed without the switch, so you have to enter something. If you don't enter anything, you will get an error.
> 
> the switch is visible with the password enforced

please try again with updated code

jesmrec:
### (29)

1. Share an item with an user in another server
2. Enter the recipient
3. Decline the invitation in Quick access

Current: the cloud icon in the bottom bar shows a (2) during some seconds. Finally, the invitation is correctly declined

Jun-05-2019 10-30-28

Expected: (2) value is not correct. The declination seems to be not an immediate operation, it can take a bit log. So, setting an spinner or something to show the user that it is running will improve the UX

jesmrec:
### (30)

1. Share some content with federation
2. Recipient accepts invitation
3. Recipients open the card and select "Unshare"

Current: Content is not unshared
Expected: Content unshared

jesmrec:
### (31) (improvement)

Open "Link" view. Private link is explained, Public Link is not. I would add a little explanation to distinguish correctly both kind of links:

Screenshot 2019-06-05 at 17 15 55

what do you think?

hosy:
@felix-schwarz please can you have a look at QA finding [(6)](https://github.com/owncloud/ios-app/pull/358#issuecomment-496811238)?

hosy:
> ### (31) (improvement)
> Open "Link" view. Private link is explained, Public Link is not. I would add a little explanation to distinguish correctly both kind of links:
> 
> Screenshot 2019-06-05 at 17 15 55
> 
> what do you think?

Not requested by management at the moment.

jesmrec:
About (30), the share is correctly rejected, but added again to pending invitations:

Jun-06-2019 09-53-13

jesmrec:
### (32)

1. Share several items with an user in another server (federated)
2. Open the recipient

Current: Notification is correct, but in the list of pending invites, only one share appears
Expected: List of pending invites

felix-schwarz:
@jesmrec (30) is fixed by the latest commit

felix-schwarz:
@jesmrec Regarding (6):

> ### (6)
> 1. "User1" invites "User2" with no Edit permission granted over one item
> 2. "User2" edit the file in Files App
> 
> Current: File is updated locally. It is intended to be uploaded to the server, but the lack of permissions makes the spinner in file list to be there forever.
> 
> Expected: User is notified the issue in any way (alert view, or directly dismiss the upload)

Looking into this, this looks like a bug in the ownCloud Server, which returns wrong incorrect permissions for shared single files. User `demo` shares the file `San Francisco.jpg` with `admin` with these permissions:

Bildschirmfoto 2019-06-06 um 10 42 41

A PROPFIND by `admin` still returns this XML for the shared file:
```xml
<d:response>
    <d:href>/remote.php/dav/files/admin/San%20Francisco.jpg</d:href>
    <d:propstat>
        <d:prop>
            <d:resourcetype/>
            <d:getlastmodified>Thu, 06 Jun 2019 08:02:40 GMT</d:getlastmodified>
            <d:getcontentlength>216071</d:getcontentlength>
            <d:getcontenttype>image/jpeg</d:getcontenttype>
            <d:getetag>"bec4e2276fc12a43dc81360be43e1d46"</d:getetag>
            <oc:id>00000012oc1hstuj49jp</oc:id>
            <oc:size>216071</oc:size>
            <oc:permissions>SDNVW</oc:permissions>
            <oc:favorite>0</oc:favorite>
            <oc:share-types/>
            <oc:owner-id>demo</oc:owner-id>
            <oc:owner-display-name>demo</oc:owner-display-name>
        </d:prop>
        <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
</d:response>
```

The key passage here is `<oc:permissions>SDNVW</oc:permissions>`, with `W` indicating the file is writable.

The permissions are correct only for shared folders and their contents.

Sharing `Photos` with these permissions …
Bildschirmfoto 2019-06-06 um 10 49 44

… the `PROPFIND` returns correct permission information:

```xml
<d:response>
    <d:href>/remote.php/dav/files/admin/Photos%20(2)/San%20Francisco.jpg</d:href>
    <d:propstat>
        <d:prop>
            <d:resourcetype/>
            <d:getlastmodified>Thu, 06 Jun 2019 08:02:40 GMT</d:getlastmodified>
            <d:getcontentlength>216071</d:getcontentlength>
            <d:getcontenttype>image/jpeg</d:getcontenttype>
            <d:getetag>"bec4e2276fc12a43dc81360be43e1d46"</d:getetag>
            <oc:id>00000012oc1hstuj49jp</oc:id>
            <oc:size>216071</oc:size>
            <oc:permissions>S</oc:permissions>
            <oc:favorite>0</oc:favorite>
            <oc:share-types/>
            <oc:owner-id>demo</oc:owner-id>
            <oc:owner-display-name>demo</oc:owner-display-name>
        </d:prop>
        <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
</d:response>
```

Here `<oc:permissions>S</oc:permissions>` simply indicates that this item is shared. The `W` permission is missing and the read-only nature of this file properly propagated by the File Provider: the Files app will not show an annotation/paint icon for such files:

Bildschirmfoto 2019-06-06 um 10 52 27

felix-schwarz:
@jesmrec Regarding (10): that's fixed with the latest commit

hosy:
> ### (32)
> 1. Share several items with an user in another server (federated)
> 2. Open the recipient
> 
> Current: Notification is correct, but in the list of pending invites, only one share appears. Once you accept or decline the next one is there
> Expected: List of pending invites

@jesmrec fixed!

jesmrec:
about (10), not posible to share if the username last character is  '@' 
Jun-06-2019 11-52-49

jesmrec:
with commit `a951f4d` , (10) is still reproducible. Is it pointing to the right SDK commit?

jesmrec:
(10) -> fixed!

jesmrec:
### (33)

1. Share with user/group a folder 
2. Browse into the folder and try to delete items (permission was granted in the creation)

Current: Unshare option available, no way to delete.
Expected: Unsharing is an option for the "source" item (shared folder in step 1.), the items inside can be removed (if granted)

jesmrec:
### (34) (improvement)

1. Open "Share this file/folder"
2. Search and choose a recipient
3. Remove connection
4. Save

Current: error can be improved

Screenshot 2019-06-07 at 10 42 26

hosy:
> ### (33)
> 1. Share with user/group a folder
> 2. Browse into the folder and try to delete items (permission was granted in the creation)
> 
> Current: Unshare option available, no way to delete.
> Expected: Unsharing is an option for the "source" item (shared folder in step 1.), the items inside can be removed (if granted)

@jesmrec good catch!

felix-schwarz:
> ### (33)
> 1. Share with user/group a folder
> 2. Browse into the folder and try to delete items (permission was granted in the creation)
> 
> Current: Unshare option available, no way to delete.
> Expected: Unsharing is an option for the "source" item (shared folder in step 1.), the items inside can be removed (if granted)

@jesmrec Fixed as of c06f6b9b978519cd2e2c535c2daad850c97f95fb. The "Unshare" option is only shown for the actually shared "root item", not the items inside it though. Would've found it confusing otherwise.

jesmrec:
### (35)

In web UI:

Screenshot 2019-06-07 at 12 25 27


In app:

Screenshot 2019-06-07 at 12 26 16

Current: Share and Delete are hidden
Expected: Share and Delete are visible and unchecked

hosy:
> ### (35)
> In web UI:
> 
> Screenshot 2019-06-07 at 12 25 27
> 
> In app:
> 
> Screenshot 2019-06-07 at 12 26 16
> 
> Current: Share and Delete are hidden
> Expected: Share and Delete are visible and unchecked

@jesmrec it was decided to hide elements, which cannot be selected -> [WON`T FIX]

jesmrec:
But they can be selected, this option is to mark by default when the share is being created.

F. ex, in webUI, with the set up:

Screenshot 2019-06-07 at 12 25 27

when you create a share:

Screenshot 2019-06-07 at 13 51 00

which behaviour is the correct, then? 

CC @pmaier1

hosy:
> But they can be selected, this option is to mark by default when the share is being created.
> 
> F. ex, in webUI, with the set up:
> 
> Screenshot 2019-06-07 at 12 25 27
> 
> when you create a share:
> 
> Screenshot 2019-06-07 at 13 51 00
> 
> which behaviour is the correct, then?
> 
> CC @pmaier1

@jesmrec what do you mean exactly? Sharing and Deleting is not allowed or is not a default permission?

jesmrec:
i mean.

You see this in the webUI

Screenshot 2019-06-07 at 13 51 00

options "Share" and "delete"  unchecked, but you can check them if you want to grant those permissions.

In the app:

Screenshot 2019-06-07 at 14 14 47

"Share" and "Delete" are hidden, and you do no have the choice to grant those permissions.

Issue #319, opened 2019-03-15
ClientDirectoryPickerViewController shows "Empty folder" when selecting target directory for Copy / Move


mneuwert:
### Steps to reproduce
1. Select a file and tap on "Copy" or "Move"
2. Select a directory which doesn't contain subdirectories

### Expected behaviour
Since as a user I know that the selected folder is not empty (e.g. it contains a lot of files which are just filtered out in this case), it is irritating that "empty folder" is shown. May be in this case nothing should be shown at all.

### Actual behaviour
Prominent "Empty Folder" message is shown in the middle of the screen

### Client
**iOS version:**
12.x

**ownCloud app version:**
master / 868097f150b66e1fd9b79d5996c4e06b4f0ae025

**Device model:**
any device

michaelstingl:
How dos this look in other products with directory tree and files listing? iOS Files? 3rd party?

mneuwert:
@michaelstingl When selecting target folder for copy / move operation, iOS Files app shows a tree structure in a single view and expands the folder when you tap it, so that the subfolders are shown. Dropbox opens every folder in a new view (similar to our app), but it is also showing files contained in that folder. The files are however greyed out since you can't open them.

jesmrec:
Done!

PR #366, opened 2019-04-29
Not showing empty folder message on copy / move

mneuwert:
## Description
Previously even if target folder had files, message "Empty folder" was shown due to the query configured to return only subfolders. Now the message is not shown during target folder selection for copy or move operation.

## Related Issue
#319

## Motivation and Context
Better UX, less confusion for the user

## Types of changes

## Checklist:


jesmrec:
I have one question here. We have two different situations:

- The target folder does not contain any subfolder in which move the item to
- The target folder is empty

The placeholder image that was removed in the current PR was intended to cover the second one. Do you think it should be removed? is it posible to keep for the second situation?

mneuwert:
@jesmrec I think the main problem is that we create a query which excludes the files and therefore we don't know if the particular folder does contain any file or not. I think it would be Ok not to show "empty folder" in this case. But we can check with @michaelstingl.

jesmrec:
just for clarifying: the move picker looks like that in both situations (no subfolders, no content):

Screenshot 2019-05-10 at 08 41 43

Bofore, both cases were displayed like that:

Screenshot 2019-05-10 at 08 45 49

In the first case (current PR implementation), you do not know if  the folder is empty or lacks of subfolders. In the `master` implementation, if the target folder lacks of subfolders you will see it as empty.

jesmrec:
Accorded fix is OK. I only miss some order in the list. For example, in case i have in root folder the following subfolders:

- AA
- bb
- CC
- Dd

if i try to move something there, i see in the picker:

Screenshot 2019-05-14 at 17 30 00

the sorting is neither by size, nor by name, nor by date. 

How are the items listed @mneuwert ?

mneuwert:
@jesmrec Sorry, missed this point.. Added alphabetical sorting with last commit. Would it be fine?

jesmrec:
Absolutely fine.

Approved

PR #369, opened 2019-04-30
[feature/se-concurrency] Sync Engine concurrency support

felix-schwarz:
## Description
This PR adds Sync Engine concurrency support to the app, so that more than one Sync Action (like move, copy, delete, download, upload, …) can be executed at once. To enable this, the Sync Engine: 
- builds a tree of dependencies between the scheduled actions by putting them in Sync Lanes 
- performs actions without dependencies on other actions immediately
- queues actions that depend on the completion of preceding actions and performs them when it is their turn

## Types of changes

### Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Concurrency.md

### Bugs & improvements:

- [X] (1) Bottom progress bar gets crazy https://github.com/owncloud/ios-app/pull/369#issuecomment-488992655 [FIXED]
- [x] (2) Countdown in progress message https://github.com/owncloud/ios-app/pull/369#issuecomment-490436459


jesmrec:
### (1)

About the progress bar in concurrency:

Check these steps:

1. Upload a big file
2. Then upload many smaller files

Current: They start to upload concurrently, but the progress status bar in the files view becomes crazy (i guess, trying to show the upload progress of every ongoing upload):

May-03-2019 09-52-29

I also detected that the progress bar does not appear anymore for downloads (only the round progress indicator inside the cell). Maybe, removing the bottom progress bar for uploads is a solution, and keep the indicator inside the cell, to avoid such glitch . More ideas welcome.

felix-schwarz:
(1) should be addressed by https://github.com/owncloud/ios-app/pull/369/commits/d744a44beab9b9e0220b73cf46b72496c811a55f with the following changes:

- progress objects of the same type are now automatically grouped and summarized like this:
PNG-Bild 4

- more recent progress objects (and groups) are now prioritized over older ones (it was the other way round before), so that if you're uploading 100 files, it initially shows the summary, but will switch to the progress info of whatever you decided to do after that for as long as that other action is ongoing.

felix-schwarz:
Re (1) again: re-added the missing lines, so download progress is now once again shown in the progress summary bar

hosy (approved):

jesmrec:
### (2) 

With the last update related the behaviour of the progress bar, i notice this:

Using the default budget:

```
OCSyncActionCategoryTransfer : @(6),	// Limit total number of concurrent transfers to 6
OCSyncActionCategoryUpload   : @(3),	// Limit number of concurrent upload transfers to 3
OCSyncActionCategoryDownload : @(3)	// Limit number of concurrent download transfers to 3
```

1.  Tap on seven files to download 
2. Three of them start to download, remaining four enqueued -> `Downloading 7 of 7 files`
3. Two of them finish -> `Downloading 5 of 7 files`
4. From the file list add three extra files to download

Current: `Downloading 8 of 10 files`. But in the list, there are 8 files instead of 10. When everything is finished, new transfers trigger a new countdown.

Suggested (more ideas welcome):

if the progress message is `Downloading X of Y files`:

X -> ongoing transfers (maximum 6 for the default budget, or separate for uploads/downloads)
Y -> pending/non-finished transfers (number of cells in status view)

so, it counts only the pending stuff because the finished one is out of scope in terms of Status.

what do you think? CC @michaelstingl

jesmrec:
Ok, i got your idea and make sense to do it in that way. Let's give a try.

jesmrec (approved):

jesmrec:
Approved on my side, correct application of concurrency taking in account the budgets.

Issue #365, opened 2019-04-29
Colliding names when uploading from Albums


jesmrec:
Look at these steps:

1. i have the following Albums, from which i want to upload content to my account:

![Screenshot 2019-04-25 at 09 38 01](https://user-images.githubusercontent.com/14894746/56718141-8051f980-673e-11e9-8d88-7fead83a2546.png)

2. I select "screenshots" (84 files), "Select all" and "Upload", to upload everything there to oC
3. All files are set in the target folder and start to upload. You can see that:

![Screenshot 2019-04-25 at 09 38 42](https://user-images.githubusercontent.com/14894746/56718246-b55e4c00-673e-11e9-9811-753bd631d227.png)

there are several enqueued files with the same name, that will collide in the moment the second one is uploaded, with the following dialog:

![Screenshot 2019-04-25 at 09 39 00](https://user-images.githubusercontent.com/14894746/56718356-e0e13680-673e-11e9-8eed-e363cda8f97f.png)

The dialog is OK, since the user can choose the action. But if you upload a very huge number of files, the number of collisions raise a lot (with 84 files it was 17,  ~20%), making the user decide every time what to do.  User only wanted to upload content in the album, regardless the name if the files (with files stored in Files App is different).

PR #370, opened 2019-04-30
Fixing #365: Colliding names when uploading from Albums

mneuwert:
## Description
- Offloaded photo import operations off the main thread
- Using fullsize image URL to derive filename

## Related Issue
#365 

## Motivation and Context
@felix-schwarz reported that UI thread is blocked while uploading a bunch of images.
@jesmrec reported in #365 that there are name conflicts on his device when uploading images from 'Screenshots' album.

## Types of changes

## Checklist:

____

Bugs & improvements:

- [ ] (1) Colliding name https://github.com/owncloud/ios-app/pull/370#issuecomment-488618221

### DIFF
--- a/ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
+++ b/ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
@@ -7,13 +7,13 @@
 //
 
 /*
- * Copyright (C) 2019, ownCloud GmbH.
- *
- * This code is covered by the GNU Public License Version 3.
- *
- * For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
- * You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
- *
+* Copyright (C) 2019, ownCloud GmbH.
+*
+* This code is covered by the GNU Public License Version 3.
+*
+* For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
+* You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+*
 */
 
 import UIKit
@@ -36,20 +36,30 @@ class UploadBaseAction: Action {
 	}
 
 	// MARK: - Upload
-	func upload(itemURL: URL, to rootItem: OCItem, name: String, completionHandler: ClientActionCompletionHandler? = nil) {
-		if let progress = core?.importFileNamed(name, at: rootItem, from: itemURL, isSecurityScoped: false, options: nil, placeholderCompletionHandler: nil, resultHandler: { (error, _ core, _ item, _) in
-			if error != nil {
-				Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path))")
-				completionHandler?(false)
-			} else {
-				Log.debug("Success uploading \(Log.mask(name)) to \(Log.mask(rootItem.path))")
-				completionHandler?(true)
-			}
+	func upload(itemURL: URL, to rootItem: OCItem, name: String, completionHandler: ClientActionCompletionHandler? = nil, importByCopy:Bool = false) {
+		if let progress = core?.importFileNamed(name,
+												at: rootItem,
+												from: itemURL,
+												isSecurityScoped: false,
+												options: importByCopy ? [OCCoreOption.importByCopying : true] : nil,
+												placeholderCompletionHandler: { (error, item) in
+													Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path)), error: \(error?.localizedDescription ?? "" )")
+													completionHandler?(false, error)
+
+		},
+												resultHandler: { (error, _ core, _ item, _) in
+													if error != nil {
+														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path)), error: \(error?.localizedDescription ?? "" )")
+														completionHandler?(false, error)
+													} else {
+														Log.debug("Success uploading \(Log.mask(name)) to \(Log.mask(rootItem.path))")
+														completionHandler?(true, nil)
+													}
 		}) {
 			self.publish(progress: progress)
 		} else {
 			Log.debug("Error setting up upload of \(Log.mask(name)) to \(Log.mask(rootItem.path))")
-			completionHandler?(false)
+			completionHandler?(false, nil)
 		}
 	}
 }
--- a/ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
+++ b/ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
@@ -43,9 +43,10 @@ class UploadBaseAction: Action {
 												isSecurityScoped: false,
 												options: importByCopy ? [OCCoreOption.importByCopying : true] : nil,
 												placeholderCompletionHandler: { (error, item) in
-													Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path)), error: \(error?.localizedDescription ?? "" )")
-													completionHandler?(false, error)
-
+													if error != nil {
+														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path)), error: \(error?.localizedDescription ?? "" )")
+														completionHandler?(false, error)
+													}
 		},
 												resultHandler: { (error, _ core, _ item, _) in
 													if error != nil {
--- a/ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
+++ b/ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
@@ -20,6 +20,10 @@ import UIKit
 import ownCloudSDK
 
 class UploadBaseAction: Action {
+
+	typealias UploadCompletionHandler = (_ success: Bool, _ item:OCItem?) -> Void
+	typealias UploadPlaceholderCompletionHandler = (_ item:OCItem?, _ error:Error?) -> Void
+
 	// MARK: - Action Matching
 	override class func applicablePosition(forContext: ActionContext) -> ActionPosition {
 		// Only available for a single item ..
@@ -36,7 +40,7 @@ class UploadBaseAction: Action {
 	}
 
 	// MARK: - Upload
-	func upload(itemURL: URL, to rootItem: OCItem, name: String, completionHandler: ClientActionCompletionHandler? = nil, importByCopy:Bool = false) {
+	func upload(itemURL: URL, to rootItem: OCItem, name: String, completionHandler: UploadCompletionHandler? = nil, placeholderHandler:UploadPlaceholderCompletionHandler? = nil, importByCopy:Bool = false) {
 		if let progress = core?.importFileNamed(name,
 												at: rootItem,
 												from: itemURL,
@@ -45,16 +49,16 @@ class UploadBaseAction: Action {
 												placeholderCompletionHandler: { (error, item) in
 													if error != nil {
 														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path)), error: \(error?.localizedDescription ?? "" )")
-														completionHandler?(false, error)
 													}
+													placeholderHandler?(item, error)
 		},
 												resultHandler: { (error, _ core, _ item, _) in
 													if error != nil {
 														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path)), error: \(error?.localizedDescription ?? "" )")
-														completionHandler?(false, error)
+														completionHandler?(false, item)
 													} else {
 														Log.debug("Success uploading \(Log.mask(name)) to \(Log.mask(rootItem.path))")
-														completionHandler?(true, nil)
+														completionHandler?(true, item)
 													}
 		}) {
 			self.publish(progress: progress)
--- a/ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
+++ b/ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
@@ -231,7 +231,7 @@ class UploadPhotosAction: UploadBaseAction {
 							// Upload to the cloud
 							performUpload(sourceURL: localURL, copySource: false)
 						} else {
-							completion(false, false)
+							completion(false, true)
 						}
 					}
 				} else {
--- a/ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
+++ b/ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
@@ -99,55 +99,61 @@ class UploadPhotosAction: UploadBaseAction {
 
 		guard let userDefaults = OCAppIdentity.shared.userDefaults else { return }
 
-		let ressources = PHAssetResource.assetResources(for: asset)
+		guard let rootItem = context.items.first else { return }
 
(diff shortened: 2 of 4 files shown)


jesmrec:
I found the following issues:

- With iPhone6Plus and v12.2, the app crashes when uploading all stuff in any album. 

- With iPadAir, it does not crashes but the following behaviour:

May-02-2019 11-56-17

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -95,59 +121,82 @@ class UploadPhotosAction: UploadBaseAction {
...
+								// First write an image to a file stored in temporary directory
+								try imageData?.write(to: localURL)
+
+								cleanUpCoreImageRessources()
+
+								// Upload to the cloud
+								self.upload(itemURL: localURL, to: rootItem, name: localURL.lastPathComponent, completionHandler: { (actionPerformed) in
+									// Delete the temporary asset file
+									try? FileManager.default.removeItem(at: localURL)
+									completion(actionPerformed)
```
felix-schwarz: The `completionHandler` will only be called by the upload-method after the image has finished uploading, so you shouldn't rely on it.

To catch critical errors (like I/O errors moving the file to local storage), please use the `placeholderCompletionHandler`. Right now `UploadBaseAction.upload()` does not support it, so you could either extend that action - or you'd need to call the `OCCore.importFileNamed()` method directly.

By default, `OCCore.importFileNamed()` (which `UploadBaseAction.upload()` also uses under the hood) does move the file from the provided location into its internal storage, so that deletion is not necessary if the `placeholderCompletionHandler` does not indicate an error.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -95,59 +121,82 @@ class UploadPhotosAction: UploadBaseAction {
...
+							var imageData : Data?
+
+							func cleanUpCoreImageRessources() {
+								// Release memory consuming resources
+								imageData = nil
+								image = nil
+								ciContext.clearCaches()
+							}
+
+							if storeAsHEIF {
```
felix-schwarz: Unless I miss something, the current implementation
- will always decode and re-compress the file - even if the file is already in the right format (possibly adding additional compression artifacts, changing the colorspace from grayscale to RGB, increase file size, …) 
- also convert JPEG originals to HEIF files if HEIF is selected as output format

IMO the implementation should
- preserve the original file whenever possible
- implement the option `Convert HEIC to JPEG` as described:
    - when `off`: upload files "as-is", using the original file and format
    - when `on`: convert HEIF files to JPEGs before upload, leave JPEGs unchanged

ownCloud/Resources/en.lproj/Localizable.strings
```diff
@@ -267,7 +267,7 @@
 "Deselect All" = "Deselect All";
 "All Photos" = "All Photos";
 "Albums" = "Albums";
-"Importing '%@' from photo library" = "Importing '%@' from photo library";
+"Importing asset from photo library" = "Importing asset from photo library";
```
felix-schwarz: While not wrong, I think `item` may be more familiar to regular users.

jesmrec:
Last check with commit: `d48b2bf7e3c6696f4bc91ad46b5418b49023ca66p`

First of all, the app is not crashing anymore with the steps mentioned [here](https://github.com/owncloud/ios-app/pull/370#issuecomment-488618221). Same steps (huge number of pics selected from an album), makes the app crash in `master` so the improvement is cool. 

About the name collisions, i have tested with different devices, obtaining different results:

**iPhone 6Plus v12.2  and  iPadAir2 v12.2** : Effect with different files called `FullSizeRender.jpg` like https://github.com/owncloud/ios-app/pull/370#issuecomment-488618221 happens, but in average ~1-2 times every 30 uploads.

**iPhoneX v12.2**: everything OK, no collisions.

jesmrec:
Maybe this one out of scope (we can address in a separate issue), so does not have to do with name collisions:

If i select manually some pictures (no matter the number) to upload from any album, they upload correctly. If i upload after choosing "Select All" option in the album, only are uploaded some images, but not all of them. For example: 
  
- With the iPhone6 Plus: with 120 pics in the album -> Select all -> only 24 are finally uploaded (20%)
- With iPad: 109 pics -> Select all -> 10 uploaded 
- With iPhoneX : An album with 164 pics was uploaded correctly. But trying with another album containing ~400 pics, only 5 were uploaded. 

(all of them with 12.2 and enough free storage)

i got logs with info and debug logging level from iPhone 6Plus

felix-schwarz:
Just made another observation: if I select many images for upload and then exit from the account, images are never scheduled for upload.

There is an internal `OCCore` API to indicate the beginning and end of activities during which the core should be kept alive. Let me know if its of interest to address the issue described above and I'll add a public version of that API. @mneuwert

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz Fixed finding except for:

> * also convert JPEG originals to HEIF files if HEIF is selected as output format

Since it was not requested in the original issue.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz fixed

ownCloud/Resources/en.lproj/Localizable.strings
mneuwert: @felix-schwarz fixed

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -80,11 +82,35 @@ class UploadPhotosAction: UploadBaseAction {
...
+								_ = uploadGroup.wait(timeout: .now() + 0.5)
+
+							} else {
+								// Escape on first failed download
+								break
+							}
+						}
+
+						uploadGroup.notify(queue: DispatchQueue.main, execute: {
+							self.completed( with: uploadFailed ? NSError(ocError: .internal) : nil)
```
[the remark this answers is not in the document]
felix-schwarz: @mneuwert `self.completed()` is still called only after all uploads have been scheduled, via `uploadGroup.notify()`.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
felix-schwarz: @mneuwert I saw you extended `UploadBaseAction.upload()` to also call the completion block from the `OCCore.importFileNamed(placeholderCompletionHandler)`.

The regular `completionHandler` (that is called after the actual upload has finished) and the `placeholderCompletionHandler` (that is called after updating the local metadata table/item cache) should be strictly separate, however. Otherwise, the completion block passed to `UploadBaseAction.upload()` will be called more than once most of the time.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
felix-schwarz: Looks fine now. Thanks for making the changes!

ownCloud/Resources/en.lproj/Localizable.strings
felix-schwarz: "Importing from photo library" is even better. Thanks!

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -95,59 +123,98 @@ class UploadPhotosAction: UploadBaseAction {
...
+			self.unpublish(progress: progress)
+
+			self.uploadSerialQueue.async {
+				if let input = input, let url = input.fullSizeImageURL {
+					let fileName = url.lastPathComponent
+
+					if !userDefaults.convertHeic || input.uniformTypeIdentifier == "public.jpeg" {
+						// No conversion of the image data, upload as is
+						self.upload(itemURL: url, to: rootItem, name:fileName, completionHandler: { (actionPerformed, _) in
 							// Delete the temporary asset file
```
felix-schwarz: Left-over comment right there.

mneuwert:
@felix-schwarz I fixed findings and hope that I am using the new API in a correct way

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz fixed

ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
```diff
@@ -36,20 +36,31 @@ class UploadBaseAction: Action {
...
+	func upload(itemURL: URL, to rootItem: OCItem, name: String, completionHandler: ClientActionCompletionHandler? = nil, importByCopy:Bool = false) {
+		if let progress = core?.importFileNamed(name,
+												at: rootItem,
+												from: itemURL,
+												isSecurityScoped: false,
+												options: importByCopy ? [OCCoreOption.importByCopying : true] : nil,
+												placeholderCompletionHandler: { (error, item) in
+													if error != nil {
+														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(rootItem.path)), error: \(error?.localizedDescription ?? "" )")
+														completionHandler?(false, error)
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz fixed, added second call-back in  `UploadBaseAction.upload()`

jesmrec:
With the latest commit `150ffcb682a3e0deca7ff0df8916e2d9439462de`

Name collisions are not fixed, but the rate of collision warning is again reduced to minimal. That only happens if you upload all with the specific option, not manually. 

About the other problem, i will open a new issue after merging the current one, so this issue is intended to fix the problems of the name collisions.

The most important thing is that crashes are fixed and memory handling has been improved a lot.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -95,59 +135,109 @@ class UploadPhotosAction: UploadBaseAction {
...
-						imageData = CIContext().heifRepresentation(of: fullImage, format: CIFormat.RGBA8, colorSpace: colorSpace)
-						print(fullImage.properties)
-					} else {
-						localURL = localURL.deletingPathExtension().appendingPathExtension("jpg")
-						imageData = CIContext().jpegRepresentation(of: fullImage, colorSpace: colorSpace)
+						let fileName = sourceURL.lastPathComponent
+						self.upload(itemURL: sourceURL, to: rootItem, name:fileName, completionHandler: { ( success, _) in
+							completion(success, true)
+						}, placeholderHandler: { (_, error) in
+							if copySource && error != nil {
```
felix-schwarz: I believe this should be `!copySource` instead
```suggestion
							if !copySource && error != nil {
```

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -75,16 +79,52 @@ class UploadPhotosAction: UploadBaseAction {
...
+								if uploadFailed == false {
+									// Upload image on a background queue
+									uploadGroup.enter()
+
+									self.upload(asset: asset, completion: { (success, uploadFinished) in
+										if !success {
+											uploadFailed = true
+										}
+										if uploadFinished {
+											uploadGroup.leave()
```
felix-schwarz: (2/3) … but the group is only left when `self.upload(itemURL:…)` is called in L212.

The group is not left if there's f.ex. a conversion error…

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -75,16 +79,52 @@ class UploadPhotosAction: UploadBaseAction {
...
+									_ = uploadGroup.wait(timeout: .now() + 0.5)
+
+								} else {
+									// Escape on first failed download
+									break
+								}
+							}
+
+							uploadGroup.notify(queue: queue, execute: {
+								runningCoreCompletion()
```
felix-schwarz: (3/3) … so that `runningCoreCompletion()` is never called, either in that case, keeping the core running indefinitely.

Regarding general use of this API, it should be as brief as possible.

Since it's the job of the Core/Sync Engine to ensure the files are uploaded once submitted to it, it's safe to call `runningCoreCompletion()` when all `placeholderCompletionHandler`s were called.

Otherwise, the core is kept from stopping until all uploads have finished, eventually blocking users that have "logged out" from the server from logging back in.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
```diff
@@ -75,16 +79,52 @@ class UploadPhotosAction: UploadBaseAction {
...
+
+						self.core?.perform(inRunningCore: { (runningCoreCompletion) in
+
+							let uploadGroup = DispatchGroup()
+							var uploadFailed = false
+
+							for asset in assets {
+								if uploadFailed == false {
+									// Upload image on a background queue
+									uploadGroup.enter()
```
felix-schwarz: (1/3) The group is always entered…

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz actually I made that change intentionally to leave the group only if actual upload has failed or succeeded to keep the OCCore alive as long as uploads are not finished. But actually.. yes, failed conversion was an open point. I fixed it with c78014a706c23a0254c370a78bce3c6de9f7a621

felix-schwarz: Ok. If the core should be kept alive & open until the uploads have completed while *also* avoiding the blocking issues described in (3/3), another API may actually be better suited: call `OCCoreManager.requestCore()` to "retain" the core and `OCCoreManager.returnCore()` when you're done.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz yes, right. Fixed.

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz implemented as suggested

ownCloud/Client/Actions/Actions+Extensions/UploadPhotosAction.swift
mneuwert: @felix-schwarz made sure that enter and leave are balanced

jesmrec:
I will approve this, and check the final status of the name collisions after https://github.com/owncloud/ios-app/pull/369 is merged as well.

Issue #260, opened 2019-02-05
Clear storage


jesmrec:
Check the following scenario:

1. I have a huge file in the server
2. I download the file, because i want to preview it
3a. I have seen the file, so i do not want it to use my storage anymore
3b. I want to keep the file in the device, because tomorrow i want to view it again avoiding wasting time and bandwidth to download.

So, a new option to clear internal storage is useful, as well as an available offline for files (https://github.com/owncloud/ios-app/issues/134) and(or folders (https://github.com/owncloud/ios-app/issues/135)

michaelstingl:
In the first iteration, I think about a one-time action.

- This should only delete implicit downloaded files
- This should NOT delete files a user selected for "available offline"
- New option for every account
  - Show used space on the device
  - Show available space on the device??
  - Button to delete downloaded data
- Also show used space in Settings (across all accounts)
- New file action for every file

jesmrec:
First step will be removing everything downloaded in a specific bookmark. To remove, we can use:

- Info row action 
- Swipe left in bookmark list

SDK is involved

jesmrec:
Done!

PR #376, opened 2019-05-08
Clear storage

mneuwert:
## Description
Added option in settings to clean up offline file storage. Also showing how much space is occupied by downloaded files and how much space is available on the device.

## Related Issue
#260 

## Motivation and Context
Helps to avoid cluttering user's device with downloaded offline data.

Screenshot 2019-05-08 at 20 08 22

## Types of changes

## Checklist:

## Test Plan: 

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Clear%20Storage.md


ownCloud/Bookmarks/BookmarkViewController.swift
```diff
@@ -810,6 +862,23 @@ class BookmarkViewController: StaticTableViewController {
...
 	}
 
+	func updateStorageInfo() {
+		if bookmark != nil {
+			if let vaultURL = OCVault(bookmark: bookmark!).filesRootURL {
+				FileManager.default.calculateDirectorySize(at: vaultURL) { (size) in
+					if size != nil {
+						let occupiedSpace = self.byteCounterFormatter.string(fromByteCount: size!)
+						self.offlineStorageInfoRow?.value = occupiedSpace
+					}
```
felix-schwarz: Error handling is missing for `size == nil`… maybe could show "not available" or "n/a" in that case.

ownCloud/Tools/FileManager+Extension.swift
```diff
@@ -0,0 +1,53 @@
...
+					if let fileAttrs = try? self.attributesOfItem(atPath: fileURL.path) {
+						if let size = (fileAttrs[FileAttributeKey.size] as? NSNumber)?.int64Value {
+							totalSize! += size
+						}
+					}
+				}
+			}
+
+			OnMainThread {
+				completion(totalSize)
```
felix-schwarz: Like the extension! However, returning to the main thread should be part of the `completion` block itself, so the method can be used purely in the background, too.

mneuwert:
> Already looking good, just a few points:
> 
> * I assumed the plan was to create an additional "Info" view (with just "Certificates" + "Free space/Compacting" for now) instead of adding this to the `BookmarkViewController`. By adding it there, I have to choose between "Cancel" and "Save" to return to the server list, which could be confusing to some users.
> * please update the `ios-sdk` to the latest `develop` commit, so the Files app running side by side will get notified of updates, too

@felix-schwarz In our last meeting I suggested to put the storage information inside the "Edit" view to avoid adding third swipe action in the server list. And.. it felt just right to do so, since all this stuff is related. So, we agreed ion that approach. I will fix the 2 findings you mentioned.

ownCloud/Bookmarks/BookmarkViewController.swift
mneuwert: @felix-schwarz right now it shows "unknown" from the beginning, since the calculation is done on a background thread and takes a couple of seconds. So if size is nil, it will just stay "unknown".

ownCloud/Tools/FileManager+Extension.swift
mneuwert: @felix-schwarz Glad you like it! Moved `OnMainThread { ... }`

mneuwert:
@felix-schwarz updated SDK in the branch to` develop/LATEST`

ownCloud/Bookmarks/BookmarkViewController.swift
felix-schwarz: Missed that. Thanks!

mneuwert:
> Compared to that a separate "Info" or "Manage" option would have clearer labelling, show the options without scrolling and just have a "Close", "Back" or "Done" button at the top. It would also be reusable in a branding context where `BookmarkViewController` is not used.

@felix-schwarz Addressed this.. I agree that especially given the context of non-destructive changes which are cancellable, storage management options are better placed in a different dedicated view.

hosy:
@mneuwert looks cool! I would suggest to change the background Color for the row action to make a difference to the edit action.

mneuwert:
> @mneuwert looks cool! I would suggest to change the background Color for the row action to make a difference to the edit action.

Yes, absolutely, changed in latest commit however was not sure if the color should be defined in the color scheme.
Screenshot 2019-05-13 at 15 21 40

michaelstingl:
Not sure if I like so many items in the swipe buttons? Make "Manage" a sub-view from from "Edit"? (Like "Advanced" sub-view in the system Settings)

mneuwert:
@michaelstingl I actually think (as already pointed out by @felix-schwarz) it is not a good idea to put storage management under "Edit account" since this screen has options to cancel and save changes, but "delete offline files" is rather an action which is performed immediately and will not e.g. be undone if user taps "cancel". May be we could restructure "Edit account" UI in some which which would allow us to separate these concerns.

michaelstingl:
Yeah, open for ideas…

jesmrec:
Approved. 

Great job!

jesmrec:
Sorry for the confusion with labels, i thought i git a weird behaviour but it was my fault. 

Approved! great job!

Issue #334, opened 2019-03-27
Implement option for log rotation


hosy:
### Description

Would be great to have an option in app settings to enable / disable log rotation. This means that only the last 24 hours logged messages available in the log file. This will prevent to large log files and it is easier to read for developers.
SDK is also involved into this issue.

### Expected behaviour
Log file contains the log messages of the last 24 hours.

### Actual behaviour
Log file saves all log messages, if enabled

### Client
**ownCloud app version:**
1.0.0 build 113

michaelstingl:
Check permanent logging feature in the desktop sync client:
https://doc.owncloud.org/desktop/troubleshooting.html#logging-to-a-temporary-directory

PR #382, opened 2019-05-16
Implement log rotation

mneuwert:
## Description
- Archiving logs kept for 24 hours with name ownCloud.log.<timestamp>
- Keeping up to 10 logs archived and removing older logs
- Added UI showing list of log files and allowing to delete / share individual logs

## Related Issue
#334

## Motivation and Context
- Keeping logs fairly short
- Not to clutter device storage with huge log files

Screenshot 2019-05-16 at 10 11 19

## Types of changes

## Checklist:

## Test Plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Log%20Rotation.md

## Bugfixing and improvements:

- [x] (1) Option to browse hidden https://github.com/owncloud/ios-app/pull/382#issuecomment-493897390 . Moved to https://github.com/owncloud/ios-app/issues/397
- [x] (2) File deleted and not generated anymore https://github.com/owncloud/ios-app/pull/382#issuecomment-493913833 [FIXED]
- [x] (3) `Delete all` does not delete today's file https://github.com/owncloud/ios-app/pull/382#issuecomment-493917754 [FIXED]
- [x] (4) Log files have the same name https://github.com/owncloud/ios-app/pull/382#issuecomment-493922065 [FIXED]
- [x] (5) Logging disabled after disabling `Log file` https://github.com/owncloud/ios-app/pull/382#issuecomment-493927466 [FIXED]
- [x] (6) 11 log files are retained https://github.com/owncloud/ios-app/pull/382#issuecomment-493949452  [FIXED]
- [ ] (7) Remove legacy log file https://github.com/owncloud/ios-app/pull/382#issuecomment-494275301


ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,243 @@
...
+
+	var shareAction : ((_ cell:UITableViewCell) -> Void)?
+
+	override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+		super.init(style: .subtitle, reuseIdentifier: reuseIdentifier)
+
+		let shareButton = UIButton(type: .system)
+		shareButton.setImage(UIImage(named: "open-in"), for: .normal)
+		shareButton.addTarget(self, action: #selector(shareButtonTapped), for: .touchUpInside)
+		shareButton.frame = CGRect(origin: CGPoint(x:0.0, y:0.0), size: shareButton.imageView!.image!.size)
```
felix-schwarz: This will crash if there's no image set (i.e. when there's no longer an image named `open-in` bundled in the app). Maybe wrap the whole shareButton segment inside `if let image = UIImage(named: "open-in") {}`?

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,243 @@
...
+				if let attributes = try? FileManager.default.attributesOfItem(atPath: logURL.path) {
+					creationDate = attributes[FileAttributeKey.creationDate] as? Date
+					fileSize = (attributes[FileAttributeKey.size] as? NSNumber)?.int64Value
+				}
+
+				let logEntry = LogEntry(name: name, size: fileSize, creationDate: creationDate)
+				self.logFileEntries.append(logEntry)
+			}
+		}
+
```
felix-schwarz: It will probably work fine the way it is, but in case APFS will not return the log files in chronological order in the future, it would be awesome to sort them by date.

Also, I feel there's an opportunity to avoid duplicating code by removing the `LogEntry` class and instead providing an equivalent from the SDK, so that `logFiles` provides an array of f.ex. `OCLogFileRecord` instances - with properties for file size, creation date and URL.

I also see the date of each log file - and especially the way to retrieve it - as an internal implementation detail. But right now there are separate implementations to get it in both the app and the SDK.

If `OCLogFileWriter.logFiles` returned an array of `OCLogFileRecord` (or any other name, really) instances with said properties, the UI would be strictly just a consumer of data provided by the SDK.

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,243 @@
...
+			shareViewController.popoverPresentationController?.sourceView = self.view
+		}
+		self.present(shareViewController, animated: true, completion: nil)
+	}
+
+	private func removeLog(with fileName:String, indexPath:IndexPath) {
+		guard let appGroupLogsURL = OCAppIdentity.shared.appGroupLogsContainerURL else { return }
+		let logURL = appGroupLogsURL.appendingPathComponent(fileName)
+		do {
+			try FileManager.default.removeItem(at: logURL)
```
felix-schwarz: Removal, too, should be managed through the same method both in the SDK and in the app, f.ex. through a `-[OCLogWriter deleteLog:(OCLogFileRecord *)log completionHandler:(void(^)(NSError *error))completionHandler]` method. As it is, the actual implementation to delete an individual log file exists twice - once in the app and once in the SDK.

michaelstingl:
@mneuwert add a line below to explain the retention period? (24h, 10 logs)

ownCloud/Settings/LogFilesViewController.swift
mneuwert: @felix-schwarz for me it is a philosophical question: let the app crash because there is some problem and find the crash early in the QA process or may be ship not crashing app accidentally slipping through QA but containing visual glitches.

ownCloud/Resources/de-DE.lproj/Localizable.strings
```diff
@@ -171,6 +171,11 @@
 "ownCloud Classic" = "ownCloud Klassik";
 
 /* Log settings */
+"Log Files" = "Protokolle";
+"Share" = "Teilen";
+"Delete all log files?" = "Alle Protokolle entfernnen?";
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,243 @@
...
+		self.tableView.register(LogFileTableViewCell.self, forCellReuseIdentifier: LogFileTableViewCell.identifier)
+		self.title = "Log Files".localized
+	}
+
+	override func viewWillAppear(_ animated: Bool) {
+		super.viewWillAppear(animated)
+		OCLogger.shared.pause()
+		self.populateLogFileList()
+
+		let removeAllButtonItem = UIBarButtonItem(title: "Delete All".localized, style: .done, target: self, action: #selector(removeAllLogs))
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Settings/LogFilesViewController.swift
mneuwert: @felix-schwarz done

ownCloud/Settings/LogFilesViewController.swift
mneuwert: @felix-schwarz done

mneuwert:
> @mneuwert add a line below to explain the retention period? (24h, 10 logs)

@michaelstingl added a note
Screenshot 2019-05-17 at 08 58 08

ownCloud/Settings/LogSettingsViewController.swift
```diff
@@ -139,63 +139,13 @@ class LogSettingsViewController: StaticTableViewController {
...
-
-					row.viewController?.present(alert, animated: true, completion: nil)
-				}, title: "Reset log file".localized, style: .destructive, identifier: "reset-logfile"))
+				// Creation of the frequency row.
+				let logsRow = StaticTableViewRow(subtitleRowWithAction: { [weak self] (_, _) in
+						let logFilesViewController = LogFilesViewController()
+						self?.navigationController?.pushViewController(logFilesViewController, animated: true)
+					}, title: "Log Files".localized, accessoryType: .disclosureIndicator, identifier: "viewLogs")
+				logOutputSection?.add(row: logsRow)
+				logOutputSection?.footerTitle = "Up to 10 recent logs are kept on the device and each log can accumulate messages for the last 24 hours of the app use".localized
```
felix-schwarz: Since the File Provider also generates log output if logging is enabled, I'd like to suggest this instead:
"The last 10 log files are kept on the device - with each log file covering up to 24 hours of usage. You can export and delete log files by selecting "Browse log files"."

Regarding the "Mask private data" option, I noticed there's already some text pointing out its limitations, so I guess no change is required there.

Adding a notice below the "Enable logging" toggle might be a good idea, though, pointing out that log files
- may end up containing sensitive information
- can impact performance
- are stored on the device and not sent to ownCloud unless explicitly shared by the user themselves.
- … anything else? @michaelstingl @hosy

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,222 @@
...
+				self?.removeLogRecord(at: indexPath)
+			})
+		]
+	}
+
+	// MARK: - Private Helpers
+
+	private func populateLogFileList() {
+		guard let logFileWriter = OCLogger.shared.writer(withIdentifier: .writerFile) as? OCLogFileWriter else { return }
+		self.logRecords = logFileWriter.logRecords()
```
felix-schwarz: Maybe sort the result in reverse order so the latest log file is always on top. (I noticed you sort it from oldest to latest in the SDK as part of the log rotation routine, so the reverse sorting should likely happen in the app in order not to impact that)

ownCloud/Settings/LogSettingsViewController.swift
```diff
@@ -139,63 +139,13 @@ class LogSettingsViewController: StaticTableViewController {
...
-						})
-					})
-
-					row.viewController?.present(alert, animated: true, completion: nil)
-				}, title: "Reset log file".localized, style: .destructive, identifier: "reset-logfile"))
+				// Creation of the frequency row.
+				let logsRow = StaticTableViewRow(subtitleRowWithAction: { [weak self] (_, _) in
+						let logFilesViewController = LogFilesViewController()
+						self?.navigationController?.pushViewController(logFilesViewController, animated: true)
+					}, title: "Log Files".localized, accessoryType: .disclosureIndicator, identifier: "viewLogs")
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz agree, changed

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,222 @@
...
+		self.toolbarItems = [flexibleSpaceButtonItem, removeAllButtonItem, flexibleSpaceButtonItem]
+		self.navigationController?.setToolbarHidden(false, animated: false)
+
+		NotificationCenter.default.addObserver(self, selector: #selector(handleLogRotationNotification), name:NSNotification.Name.OCLocalLogRotationNotificationName , object: nil)
+	}
+
+	override func viewWillDisappear(_ animated: Bool) {
+		super.viewWillDisappear(animated)
+		self.navigationController?.setToolbarHidden(true, animated: false)
+		NotificationCenter.default.removeObserver(self)
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,222 @@
...
+		super.viewWillAppear(animated)
+		self.populateLogFileList()
+
+		let removeAllButtonItem = UIBarButtonItem(title: "Delete all".localized, style: .done, target: self, action: #selector(removeAllLogs))
+		let flexibleSpaceButtonItem = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
+
+		self.toolbarItems = [flexibleSpaceButtonItem, removeAllButtonItem, flexibleSpaceButtonItem]
+		self.navigationController?.setToolbarHidden(false, animated: false)
+
+		NotificationCenter.default.addObserver(self, selector: #selector(handleLogRotationNotification), name:NSNotification.Name.OCLocalLogRotationNotificationName , object: nil)
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Settings/LogSettingsViewController.swift
mneuwert: @felix-schwarz I changed the first part.. And regarding adding another note to "Enable Logging" switch, I came up with following text:

_When activated, logs may impact performance and include sensitive information. However the logs are not subject to automatic submission to ownCloud servers. Sharing logs with others is sole user responsibility._

ownCloud/Settings/LogFilesViewController.swift
mneuwert: @felix-schwarz fixed

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -0,0 +1,235 @@
...
+					logFileWriter.cleanUpLogs(true)
+				}
+			})
+		})
+
+		self.present(alert, animated: true, completion: nil)
+	}
+
+	@objc private func handleLogRotationNotification() {
+		populateLogFileList()
```
felix-schwarz: Just noticed during testing: this notification can be delivered on a background thread and then lead to `tableView.reloadData()` being called on a thread different than the main thread.
```suggestion
		OnMainThread {
			self.populateLogFileList()
		}
```

jesmrec:
Let's QA this

jesmrec:
### (1) (improvement)

1. In Settings -> Logging, disable the option "Log File"

Current: Browse log files is displayed, but no logs will be stored there anymore
Expected: Browse log option is hidden, so no file is generated. 

The special case is keep the option visible to check older logs. Not sure if this makes sense, since the logs are something that become outdated very fast

jesmrec:
### (2)

1. Generate some logs
2. Settings -> Logging -> Browse log files
3. Swipe in the file and remove it

Current: By generating more logs (performing actions in file list), log file is not generated again 
Expected: File is created again with new logs

iPhoneX v12.2

jesmrec:
### (3)

1. With logs enabled, `Delete all`

Current: Content of today's file is deleted, but a new one is created at the moment. By swiping, today's file can be deleted. 
Expected: Everything deleted in the same way if you do one by one.

This is not a bad behaviour, because today's file is deleted and generated again... so, it could be addressed separately in case you think it is out of scope.

iPhoneX v12.2

jesmrec:
### (4) (improvement)

Every generated file does have the same name `ownCloudApp.log`. I'd distinguish them in any way, for example, including the date, f. ex.: `ownCloudApp_20190521.log` or `ownCloudApp_21May2019.log` or similar. In case you need to send more than one, will be less confusing.

jesmrec:
### (5)

1. Install the app from scratch
2. Go to Settings -> Logging 
3. Disable `Log File` 
4. Kill the app, and reopen it

Current: `Enable logging` is disabled
Expected: `Enable logging` is enabled, and `Log file` disabled

iPhoneX v12.2

May-20-2019 12-20-48

mneuwert:
> ### (3)
> 1. With logs enabled, `Delete all`
> 
> Current: Content of today's file is deleted, but a new one is created at the moment. By swiping, today's file can be deleted.
> Expected: Everything deleted in the same way if you do one by one.
> 
> This is not a bad behaviour, because today's file is deleted and generated again... so, it could be addressed separately in case you think it is out of scope.
> 
> iPhoneX v12.2

@jesmrec I think this is expected behavior, since app is constantly logging and the logger will create a new file automatically as soon as new message is logged.

jesmrec:
> @jesmrec I think this is expected behavior, since app is constantly logging and the logger will create a new file automatically as soon as new message is logged.

ok, i was wondering about this, but it is OK for me.

mneuwert:
> ### (4) (improvement)
> Every generated file does have the same name `ownCloudApp.log`. I'd distinguish them in any way, for example, including the date: `ownCloudApp_20190521.log` or `ownCloudApp_21May2019.log` or similar. In case you need to send more than one, will be less confusing.

@jesmrec The files are in fact following the following scheme ownCloud.log.<date_time>. However in the log file management UI, I just strip the timestamp. We agreed with @felix-schwarz to do so to not to clutter the UI with long file names since the creation timestamp is anyway shown below the name. When you share the file however, the timestamp will be preserved in the filename.

mneuwert:
> ### (1) (improvement)
> 1. In Settings -> Logging, disable the option "Log File"
> 
> Current: `Browse log files` is displayed, but no logs will be stored there anymore
> Expected: `Browse log` option is hidden, because no file is generated.
> 
> The special case is keeping the option visible to check older logs. Not sure if this makes sense, since the logs are something that become outdated very fast

@jesmrec I can do it, but it will require significant code change and will need some time to rework log settings view controller.

mneuwert:
> ### (2)
> 1. Generate some logs
> 2. Settings -> Logging -> Browse log files
> 3. Swipe in the file and remove it
> 
> Current: By generating more logs (performing actions in file list), log file is not generated again
> Expected: File is created again with new logs
> 
> iPhoneX v12.2

@jesmrec some how this is contradicting with (3)

jesmrec:
> @jesmrec The files are in fact following the following scheme ownCloud.log.<date_time>. However in the log file management UI, I just strip the timestamp. We agreed with @felix-schwarz to do so to not to clutter the UI with long file names since the creation timestamp is anyway shown below the name. When you share the file however, the timestamp will be preserved in the filename.

I was aware about the UI/UX decision not to include the timestamp but in sharing, but it was not included in file name. I have just noticed that it is included in the file name, only in case there are more than one stored log after installing. Nothing very important, but i'd include even in that case. What do you think?

jesmrec:
> @jesmrec some how this is contradicting with (3)

no, they are different. in (2), i do not see a log file anymore, even performing actions:

May-20-2019 13-02-59

(3) is only kind of improvement, as you told, we can skip it because logs are immediately generated.

jesmrec:
> @jesmrec I can do it, but it will require significant code change and will need some time to rework log settings view controller.

we can add a separate issue then. This is not functionally incorrect.

jesmrec:
### (6)

Retain is 10 logs max

Current: 11 log files are retained:

Screenshot 2019-05-20 at 13 34 45

mneuwert:
> ### (6)
> Retain is 10 logs max
> 
> Current: 11 log files are retained:
> 
> Screenshot 2019-05-20 at 13 34 45

@jesmrec It is 10 archived + 1 current

mneuwert:
> > @jesmrec The files are in fact following the following scheme ownCloud.log.<date_time>. However in the log file management UI, I just strip the timestamp. We agreed with @felix-schwarz to do so to not to clutter the UI with long file names since the creation timestamp is anyway shown below the name. When you share the file however, the timestamp will be preserved in the filename.
> 
> I was aware about the UI/UX decision not to include the timestamp but in sharing, but it was not included in file name. I have just noticed that it is included in the file name, only in case there are more than one stored log after installing. Nothing very important, but i'd include even in that case. What do you think?

@jesmrec I think I know why.. basically if you share current log, name stays unchanged. I can fix that and always add "log creation date" as suffix when sharing.

jesmrec:
> @jesmrec I think I know why.. basically if you share current log, name stays unchanged. I can fix that and always add "log creation date" as suffix when sharing. 

this would be an extra ball as well.

Let's summarize then:

(1) -> improvement, not to do here
(2) -> bad behaviour. it should be fixed
(3) -> expected behaviour. no fixes needed
(4) -> timestamp included with several files. Not extra work then
(5) -> bad (hidden) behaviour. it should be fixed
(6) -> 11 log files: 10 closed + 1 opened. Maybe changing the wording:  `The last 10 logs are kept...` for `The last 10 closed logs are kept...` ? (just an idea, i was confused with that)

mneuwert:
@jesmrec i fixed (2), added timestamp permanently to shared files (4) irrespective of if it is current or archived log. Changed wording for (6) to make it more explicit. As for (5) I was able to reproduce it only once after fresh install but reviewing code / debugging didn't bring any insight in what it could be. @felix-schwarz as you are more familiar with log level handling, do you have any ideas?

felix-schwarz:
Re (5) @jesmrec:

If this is the first launch, logging shouldn't be enabled as the default is that logging is off.

So I think what's happening here is this:
- the app is launched via Xcode, which probably uses the environment variable `oc:log.log-level=0`:
Bildschirmfoto 2019-05-20 um 23 27 01
- this overrides the user defaults and enabled logging at debug level directly, which helps debugging
- killing the app ends the debugging session with Xcode
- the next launch is without the environment variable set, so that now the default setting (logging = off) applies

If you either untick the checkbox in front of that environment variable in Xcode - or always launch it without Xcode, that issue should not occur.

jesmrec:
(2), (4), (6) -> Fixed
(5) -> by unchecking the log option in XCode, works properly. Thanks for the tip!!
(1) -> will open a new issue
(3) -> wont fix

I performed a second round with another device, everything OK

Approved

felix-schwarz:
### (7) Legacy log file

Didn't think about it earlier, but just noticed: if someone has been using the previous betas, the log file is in a different location and will no longer be used. Since some beta testers could have gigabytes worth of logs there, I think that the app should look for an old log file and simply delete it if one exists.

michaelstingl:
> I think that the app should look for an old log file and simply delete it if one exists.

👍

mneuwert:
> ### (7) Legacy log file
> Didn't think about it earlier, but just noticed: if someone has been using the previous betas, the log file is in a different location and will no longer be used. Since some beta testers could have gigabytes worth of logs there, I think that the app should look for an old log file and simply delete it if one exists.

@felix-schwarz Fixed with d8cf15faafd6235a63edd7dbfc68ec9f7cdaba46. Please approve if you agree with the change.

jesmrec:
Fix is ok from my side as well

jesmrec:
@michaelstingl done.

Issue #364, opened 2019-04-29
Move/Copy with no target folder


jesmrec:
If the root folder contains only files (no folders) moving/copying these files should not be available, since there is no posible target folder there. 

In this case, the copy/move options should be disabled, or the user warned about such situation.

Ir comes from https://github.com/owncloud/ios-app/issues/237#issuecomment-483606734

jesmrec:
Fixed

PR #392, opened 2019-05-21
[fix/disable-move-copy] Disable move/copy action in toolbar if no folder is available

hosy:
## Description
If there is no folder available in the root folder and a file is selected, the toolbar actions move/copy are disabled, because items cannot moved or copied to an other directory.

Furthermore this is also important for:
- Menu Cardview
- Drag and Drop on toolbar items

## Related Issue
#364

## Motivation and Context
Do not show user options, which are not possible.

## How Has This Been Tested?
- open file list
- be sure, there is no directory available in the root folder
- select a file
- toolbar actions move and copy should be disabled
- create a new folder
- select file again
- toolbar actions move and copy should be enabled

**Repeat same tests for Drag & Drop and Card Menu**

## Types of changes

## Checklist:


ownCloud/Client/Actions/Actions+Extensions/CopyAction.swift
```diff
@@ -27,6 +27,9 @@ class CopyAction : Action {
 
 	// MARK: - Extension matching
 	override class func applicablePosition(forContext: ActionContext) -> ActionPosition {
+		if forContext.items.contains(where: {$0.type == .file}), let path = forContext.query?.queryPath, path.isRootPath, let containsFolder = forContext.preferences?["containsFolders"] as? Bool, !containsFolder {
```
mneuwert: @hosy would it make sense to split up a bit for better readability

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -81,6 +81,26 @@ class ClientQueryViewController: UITableViewController, Themeable, UIDropInterac
...
+		let folders = items.filter { (item) -> Bool in
+			if item.type == .collection {
+				return true
+			}
+			return false
+		}
+		if folders.count == 0 {
+			return false
+		}
+		return true
```
mneuwert: @hosy I would return `false` by default and I think check for `folders.count` is superflous, since if a folder is found, it would be caught by condition above

jesmrec:
any choice to disable in the card as well?

hosy:
> any choice to disable in the card as well? @hosy

@jesmrec fixed, also for drag and drop in toolbar

jesmrec:
Fixed and approved. There are some conflicts and branch is outdated. Could you take a look @hosy ?

hosy:
@jesmrec I will take care of the merge conflicts. Thank you!

Issue #368, opened 2019-04-30
[Release] 1.0.1


jesmrec:
Release a new version
## TASKS:

Git & Code

- [x] [GIT] Create branch `release/1.0.1` (freeze the code) @jesmrec 
- [X] [DEV] Update version number `version_1.0.1`
- [x] [DIS] Update [changelog](https://github.com/owncloud/ios-app/blob/master/CHANGELOG.md) https://github.com/owncloud/ios-app/pull/407
- [x] [TRFX] Update translations from transifex branch.
- [x] [QA] Design Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version_1.0.1.md @jesmrec 
- [x] [QA] Regression Test plan @jesmrec 
- [ ] [DOC] Update user manual with the new functionalities https://github.com/owncloud/ios-app/pull/398
- [ ] [DOC] Update owncloud.org/download version numbers (notify rocketchat #marketing)
- [x] [DEV] Settings: Check Help URL
- [ ] [DEV] Settings: Feedback subject line: `$version ownCloud iOS app feedback` or something… (ticket system often crops the end of the subject line)
- [x] [DEV] Settings: Check Recommend URL

App Store 

- [x] [DIS] App Store Connect: Create a new version following the `[major].[minor].[patch]`
- [x] [DIS] App Store Connect: Update screenshots if needed
  - [x] Empty accounts (`Welcome`) `[10_ios_accounts_welcome_demo.png]`
  - [x] Multiple accounts `[11_ios_accounts_list_demo.png]`
  - [x] File Listing `/Photos/*` `[20_ios_files_list_demo.png]`
  - [x] File Actions `/Photos` `[21_ios_files_actions_demo.png]`
  - [x] PDF preview `/ownCloud Manual.pdf` (Page 13) `[22_ios_files_preview_pdf_demo.png]`
  - [x] Quick Access `[40_ios_quick_access_demo.png]`
  - [x] Settings (Passcode enabled) `[60_ios_settings_demo.png]`
  - [x] give up-to-date screenshot of the sharing and settings screens to @lefherz 
- [x] [DIS] Upload the binary to the App Store
- [x] [DIS] App Store Connect: When release (manually, automatically, automatically not early than date)
- [x] [DIS] App Store Connect: How to release (immediately to all the users o 7-day period phased release)
- [x] [DIS] App Store Connect: Reset iOS Summary Rating (Keep existing or reset the rating)
[...]

PR #402, opened 2019-06-10
[Release] 1.0.1

jesmrec:
## Description
To perform release changes: distribution, QA...

## Related Issue
#368

## Motivation and Context
First new iOS app release

## How Has This Been Tested?

Test Plan: 

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Bugfixing:

- [X] (1) Visibility of sharing options in card https://github.com/owncloud/ios-app/pull/402#issuecomment-500337180 [FIXED]
- [x] (2) Open in option does not work if file is not downloaded https://github.com/owncloud/ios-app/pull/402#issuecomment-500389176 [FIXED]
- [x] (3) Number of downloading items does not match the downloads https://github.com/owncloud/ios-app/pull/402#issuecomment-500414744
- [X] (4) "Go to page" in PDF viewer is missing https://github.com/owncloud/ios-app/pull/402#issuecomment-500712526 [FIXED] -> new issue to address three-dot-button
- [X] (5) Uploads interrupted https://github.com/owncloud/ios-app/pull/402#issuecomment-500739425 [FIXED]
- [x] (6) Deletions are not resumed after recovering connection https://github.com/owncloud/ios-app/pull/402#issuecomment-500747340
- [X] (7) Download a file makes the app crash https://github.com/owncloud/ios-app/pull/402#issuecomment-500848377 [NOT REPRODUCIBLE]
- [X] (8) Files app: move between accounts, let files in the source account https://github.com/owncloud/ios-app/pull/402#issuecomment-500914200 [WONT FIX] Apple problem
- [X] (9) UI glitch card https://github.com/owncloud/ios-app/pull/402#issuecomment-501009853 [FIXED]
- [X] (10) Friendly UI for "Open In" https://github.com/owncloud/ios-app/pull/402#issuecomment-501011461 [DONE]
- [X] (11) Files are not attached to Mail with Open In https://github.com/owncloud/ios-app/pull/402#issuecomment-501229523 [FIXED]
- [X] (12) Thumbnails in Files app https://github.com/owncloud/ios-app/pull/402#issuecomment-501325001 [NOT REPRODUCIBLE] if it becomes reproducible, take video and logs
- [X] (13) Glitch after cancelling download https://github.com/owncloud/ios-app/pull/402#issuecomment-501651948 [FIXED]
- [X] (14) ownCloud name in some Strings https://github.com/owncloud/ios-app/pull/402#issuecomment-502011183 [FIXED]
- [X] (15) Log file name https://github.com/owncloud/ios-app/pull/402#issuecomment-502596255 [FIXED]
- [X] (16) Select log file sharing action https://github.com/owncloud/ios-app/pull/402#issuecomment-502776492 [FIXED]
- [X] (17) Public Link password text field not shown https://github.com/owncloud/ios-app/pull/402#issuecomment-502979051 [FIXED]
- [X] (18) Format string, when overwrite a unsynced file https://github.com/owncloud/ios-app/pull/402#issuecomment-503053582 [FIXED]
- [X] (19) Reshare items inside a shared folder is not posible https://github.com/owncloud/ios-app/pull/402#issuecomment-503493536 [FIXED]
- [X] (20) UI glitches when creating new folder https://github.com/owncloud/ios-app/pull/402#issuecomment-503599004 [FIXED]


jesmrec:
### (1)

Share options in the card are only visible is the item is already visible. Related with capability handling (latest commits in sharing branch)

jesmrec:
### (2)

1. On a non-downlaoded file, open the card
2. Select "Open In"

Current: `Downloading` is in foreground forever. Cancel button does not cancel the view
Expected: Card with options to open in is displayed.

jesmrec:
### (3)

I did not get specific steps to reproduce this one. Once several files has been downloaded, the countdown does not go till zero. This means, you have only one downloading item, but the status message shows  `Downloading X items`  (with X > 1):

Jun-10-2019 15-09-03

any additional info is required to help here?

jesmrec:
### (4)

1. Open PDF file

Current: "Go to page" is missing
Expected: not missing

iPhoneX v12

felix-schwarz:
@jesmrec Couldn't reproduce the issue but did some spelunking in the progress summarizer code and addressed the places where progress tracking *could* possibly break. Please check if you can still reproduce (3).

jesmrec:
### (4)

Not sure if we addressed something like this in the past:

1. Upload a bunch of files
2. Connection is lost during the transaction
3. Connection is recovered

Current: Uploads are resumed, but user is asked standly for fails:

Jun-11-2019 10-14-20

Expected: resumed uploads are smooth

jesmrec:
### (6)

in the same path as (5)

1. Delete a bunch of files by using multiselection
2. Remove connection during the deletion
3. After a while, recover connection

Current: Deletions are not resumed
Expected: Deletions resumed

iPhoneX v12

jesmrec:
### (7) 

Downloading an item makes the app crash

jesmrec:
### (8)

1. In Files App, select several items in non-root folder and the option "Move"
2. As target, select another oC account (its root folder)

Current: Items moved to the source account, in the root folder

We had some issues in moving stuff between accounts due to an Apple issues, or improvements to do:

https://github.com/owncloud/ios-app/issues/237#issuecomment-481308589
https://github.com/owncloud/ios-app/issues/237#issuecomment-483163528

And more related issues in that release attempt, but i did not notice the current behaviour. If i select as target a different account, i do not expect the content copied to the same account. I expect:

- Copied to the selected target
- Not copied (error message or something?)

felix-schwarz:
### (9)

When tapping the dots, the card first overreaches, then drops back into place.

Expected: a smooth transition without overreach and drop back.
Actual: he card first overreaches, then drops back into place.

Fixed as of https://github.com/owncloud/ios-app/pull/402/commits/43f0769d245d36750f09aa0f4d459fa56e0c83d3

felix-schwarz:
### (10)

"Open in" action fragile, not offering full multi selection support.

Expected: after selecting "Open in", the app shows accurate download progress information and allows cancelling the downloads entirely.
Actual: the app shows only the download progress of the last started download and will allow cancelling only the last started download

Fixed as of https://github.com/owncloud/ios-app/commit/43f0769d245d36750f09aa0f4d459fa56e0c83d3

felix-schwarz:
> ### (2)
> 1. On a non-downlaoded file, open the card
> 2. Select "Open In"
> 
> Current: `Downloading` is in foreground forever. Cancel button does not cancel the view
> Expected: Card with options to open in is displayed.

Missed this. Should be the same as (10) and fixed.

felix-schwarz:
@jesmrec Fixed (5) and (6) as of https://github.com/owncloud/ios-app/pull/402/commits/0151119f767fb10e40c8348afa762a798eea5096. The reason was that not all network error codes indicating failure due to missing network connectivity were checked for - and some requests could get silently dropped for that reason. Now also includes things like DNS lookup failure, data not being allowed, international roaming to be disabled, etc.

jesmrec:
### (11)

After the fixes in the ' Open In` feature:

1. Select a file and "Open In"
2. Select Mail as target

Current: File is not attached
Expected: File is attached

jesmrec:
I have checked the fixes for (5) and (6)

(5) -> first two times i checked, it worked fine. The third time, the problem happened again. From here, checked several times, it happened only sometimes.

(6) -> i reproduced it again, every time:


Jun-12-2019 13-59-13

Anything needed, please ping me. I'm trying to scope it a bit more.

felix-schwarz:
> ### (8)
> 1. In Files App, select several items in non-root folder and the option "Move"
> 2. As target, select another oC account (its root folder)
> 
> Current: Items moved to the source account, in the root folder
> 
> We had some issues in moving stuff between accounts due to an Apple issues, or improvements to do:
> 
> [#237 (comment)](https://github.com/owncloud/ios-app/issues/237#issuecomment-481308589)
> [#237 (comment)](https://github.com/owncloud/ios-app/issues/237#issuecomment-483163528)
> 
> And more related issues in that release attempt, but i did not notice the current behaviour. If i select as target a different account, i do not expect the content copied to the same account. I expect:
> 
> * Copied to the selected target
> * Not copied (error message or something?)

I could not reproduce this. Both when trying to move single items as well as multiple items, I get the expected error:

1 item | 3 items 
-- | --
IMG_9808 | IMG_9809

felix-schwarz:
Regarding (5) and (6): I can't reproduce this. Only way I can still think of this happening would be if iOS returns a different error code on your device that's not currently handled. Can you share a log file covering this? Should give me the needed clues as to why it's still happening on your device. @jesmrec

felix-schwarz:
> ### (11)
> After the fixes in the ' Open In` feature:
> 
> 1. Select a file and "Open In"
> 2. Select Mail as target
> 
> Current: File is not attached
> Expected: File is attached

Fixed as of https://github.com/owncloud/ios-app/pull/402/commits/185f632bb336a269cbccdcf98f83090d695b96a0

jesmrec:
About (5), uploads problem, i got the following log (level debug, too much info). The log was cleaned just before recovering the connection. The first alerts in screen were about files  ´1 copy 22.txt´ and ´1 copy 25.txt´

[ownCloud_12_Jun_2019_at_15_38_36.logV.txt](https://github.com/owncloud/ios-app/files/3281594/ownCloud_12_Jun_2019_at_15_38_36.logV.txt)

About deletions (6), with `info` level in logging, i got the following error in logs:

`2019-06-12 15:26:44.978000+0200 ownCloud[604:062772] [info] | [APP] Error Optional(Error Domain=OCError Code=32 ""Txt/1 copy 182.txt" is locked (error 32)" (-[OCConnection _handleDeleteItemResult:error:] [OCConnection.m:2169]) UserInfo={NSDebugDescription=-[OCConnection _handleDeleteItemResult:error:] [OCConnection.m:2169], NSLocalizedDescription="Txt/1 copy 182.txt" is locked}) deleting Optional("/Txt/1 copy 182.txt") [DeleteAction.swift:74|FULL]`

Full `info  log:  
[ownCloud_12_Jun_2019_at_15_29_38.log.txt](https://github.com/owncloud/ios-app/files/3281598/ownCloud_12_Jun_2019_at_15_29_38.log.txt)

With debug level (file `1 copy 92.txt` failing in first instance ): 

[ownCloud_12_Jun_2019_at_15_55_07.log.txt](https://github.com/owncloud/ios-app/files/3281635/ownCloud_12_Jun_2019_at_15_55_07.log.txt)

felix-schwarz:
@jesmrec Thanks for the logs! For this request to delete `1 copy 92.txt`:

```
# REQUEST ---------------------------------------------------------
URL:   http://10.40.40.198:33000/remote.php/dav/files/user1/Txt/1%20copy%2092.txt
Error: -
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
DELETE /remote.php/dav/files/user1/Txt/1 copy 92.txt HTTP/1.1
Host: 10.40.40.198:33000
X-Request-ID: 13AD2598-B92A-4207-80D5-8D43271E70FB
Accept-Language: en-gb, en
Authorization: [redacted]
If-Match: "5515cdb832824829f5c898288e4d0402"
```

… this response is returned:

```
# RESPONSE --------------------------------------------------------
Method:     DELETE
URL:        http://10.40.40.198:33000/remote.php/dav/files/user1/Txt/1%20copy%2092.txt
Request-ID: 13AD2598-B92A-4207-80D5-8D43271E70FB
Error:      -
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
412 PRECONDITION FAILED
Content-Type: application/xml; charset=utf-8
Keep-Alive: timeout=5, max=100
Pragma: no-cache
Content-Security-Policy: default-src 'none';
Set-Cookie: ocbubotjeecp=2ia7ogdc6ii6atjtctrvninetv; path=/; HttpOnly, oc_sessionPassphrase=e%2BenDD5HJs%2FjoVWqQcBhwUPNbSOzcAPUFqbKkC%2BgtjggY5ed%2Fe%2Fn00%2Fh3kF7wZ3IWUnguuyuFgBKH1GXgdWOAaY2U81ZRsImMjoXLU0IxYN%2Bd77dIWtzxMRqnxymC0mD; path=/; HttpOnly, ocbubotjeecp=bcul82icv7efr9nf54oea73sn0; path=/; HttpOnly, cookie_test=test; expires=Wed, 12-Jun-2019 14:55:22 GMT; Max-Age=3600
Server: Apache
X-Download-Options: noopen
X-XSS-Protection: 1; mode=block
Expires: Thu, 19 Nov 1981 08:52:00 GMT
X-Permitted-Cross-Domain-Policies: none
Cache-Control: no-store, no-cache, must-revalidate
Date: Wed, 12 Jun 2019 13:55:22 GMT
X-Robots-Tag: none
Content-Length: 297
Connection: Keep-Alive
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN

<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\PreconditionFailed</s:exception>
  <s:message>An If-Match header was specified and the resource did not exist</s:message>
  <s:header>If-Match</s:header>
</d:error>
```

`412 PRECONDITION FAILED` is the server's way to indicate that the target's `ETag` is not identical to the one provided in the original request's `If-Match` header field value, indicating the file has been changed on the server.

The same response is also returned for several other files, (correctly) leading to these alerts.

If the connection is really cut off, there should be no HTTP responses like these.

felix-schwarz:
@jesmrec Regarding (5): I can see one case of `NSURLSession` dropping a request - and that error - unlike connectivity errors - propagating back to the upload action, making it fail.

May need to adjust the code that reschedules certain requests when there are connectivity errors to also included these. Will follow up to this, but need to think about possible implications first.

jesmrec:
About (8):

to illustrate the scenario:

`user1` has several files inside the folder `Noconn`. I select three,  option `Move` and the account `admin` as target. Finally, the selected files keep in user1's root folder

Jun-12-2019 17-10-45

If i perform the same operation by dragging and dropping, behaviour is the same
If the source files are in root folder (not inside another folder), files are copied to the root of the other account

jesmrec:
### (12)

In files app, the pics' thumbnails are missing. Check with several servers and two different devices. Could you take a look?

felix-schwarz:
> About (8):
> 
> to illustrate the scenario:
> 
> `user1` has several files inside the folder `Noconn`. I select three, option `Move` and the account `admin` as target. Finally, the selected files keep in user1's root folder

@jesmrec Thanks for the additional info. Unfortunately this is an unfixable bug in Apple's API definition - and I'll explain why:
- the Files app addresses items by their `NSFileProviderItemIdentifier` (`itemID`), a string value that each File Provider can pick freely. For the root directory, however, the `itemID ` value *has to be* `NSFileProviderRootContainerItemIdentifier`.
- when the Files app wants to move a file within an account - or across different accounts of the same File Provider, it calls [`reparentItemWithIdentifier:toParentItemWithIdentifier:newName:completionHandler:`](https://developer.apple.com/documentation/fileprovider/nsfileproviderextension/2921609-reparentitemwithidentifier?language=objc), to which it provides the `itemID` of the item that should be moved - and the `itemID` of the target folder it should be moved too.
- when the target folder is on a different account, the File Provider can easily detect this because there won't be any item with the `itemID` of that target folder - and can return an error
- the root directory, however, is required to have the same `itemID` in all accounts (`NSFileProviderRootContainerItemIdentifier`). So when you move a file from account A to the root of account B, [`reparentItemWithIdentifier:toParentItemWithIdentifier:newName:completionHandler:`](https://developer.apple.com/documentation/fileprovider/nsfileproviderextension/2921609-reparentitemwithidentifier?language=objc) will be called with the `itemID` of the item to move and `NSFileProviderRootContainerItemIdentifier` for the target directory.
- `NSFileProviderRootContainerItemIdentifier`, of course, is a valid `itemID` inside every account, and identifies the root directory.
- in effect, there's no way for the File Provider to distinguish if a user wants to move a file to the account's root directory - or the root directory of another account, because what it gets passed in both cases is just `NSFileProviderRootContainerItemIdentifier` as the `itemID` of the target folder

You'll get a proper error when trying to move a file to a non-root folder of another account. But for moving to the root folder of another account within the same File Provider, this is technically not possible.

jesmrec:
### (13)

Small glitch:

1. Tap on a file to download. Don't choose a small one or will be difficult to perform next steps.
2. Once download has started, cancel it

Current: `Downloading <file>` notification in bottom bar does not disappear. With a pull down, it disappears
Expected: After cancelling, the notification disappears

Jun-13-2019 12-47-08

jesmrec:
Will mark (8) as wont fix. Thanks for the explanation @felix-schwarz

jesmrec:
(13) is not easily reproducible... i reproduce it with an specific server, but with a different one is more complicated. Of course not a blocker, if you can not reproduce, it will be addressed separately.

felix-schwarz:
@jesmrec Regarding (13), I couldn't reproduce it, but tried to find a vector that could trigger it - and found one. Please check if the issue can still be reproduced with https://github.com/owncloud/ios-app/pull/402/commits/ee575de367e36de560ab5c257f4571ac93e588ba

felix-schwarz:
> @jesmrec Regarding (5): I can see one case of `NSURLSession` dropping a request - and that error - unlike connectivity errors - propagating back to the upload action, making it fail.
> 
> May need to adjust the code that reschedules certain requests when there are connectivity errors to also included these. Will follow up to this, but need to think about possible implications first.

As of https://github.com/owncloud/ios-app/pull/402/commits/478535e9342a8f68d446e412d511a91b2482df33, requests dropped by the `NSURLSession` are now rescheduled/retried, too.

felix-schwarz:
@jesmrec Re (12): I made some changes so that thumbnail requests by the File Provider no longer wait for connectivity if there is none (as of https://github.com/owncloud/ios-app/pull/402/commits/3cc734f0d26a4e6b11a3834ab37545f32813dcac)

Also added more debug output, so if you can still reproduce the issue, I'd be very interested at looking at a log file of the incident.

jesmrec:
### (14)

This localizable string:

https://github.com/owncloud/ios-app/blob/fe9442d1ccf3a871184bcd2a178e749a480e4ae0/ownCloud/Resources/en.lproj/Localizable.strings#L204

contains `ownCloud`, it should be parameterized.

Also contains `ownCloud` the themes name (ownCloud Dark, classic and light). New parameters should be set there?

jesmrec:
(5) -> I still reproduce it. But now, it is only one file (before, the alerts showed different files). This is the log i got. The involved file in ' 1 copy 11.txt`

[ownCloud_14_Jun_2019_at_11_05_57.log.txt](https://github.com/owncloud/ios-app/files/3289917/ownCloud_14_Jun_2019_at_11_05_57.log.txt)

(12) -> Still reproducible. Logs from the app are enough (will be there Files App stuff)?

(13) -> More difficult to reproduce as before, but i could reproduce sometimes. This would be an extreme corner case and the current fix is enough for me.

felix-schwarz:
@jrecio Thanks for the log file for (5)! Should no longer occur as of 7dbf49008f4a7731333e5834380ee5adff746b4f.

jesmrec:
(5) is fixed

hosy:
@jesmrec (14) is ready to test!

jesmrec:
Regarding (12), i got a log files with the following steps:

1. Open Files App
2. Open an account with images and browse to the folder 
3. Wait for the thumbnails...

[ownCloud_17_Jun_2019_at_10_50_46.log.txt](https://github.com/owncloud/ios-app/files/3296039/ownCloud_17_Jun_2019_at_10_50_46.log.txt)

jesmrec:
### (15)

Log files are called `ownCloudApp.log` 

https://github.com/owncloud/ios-sdk/blob/19696380ddbd9447887cec1b3c041e3cdc2ccf6d/ownCloudSDK/Logging/Writers/OCLogFileWriter.m#L53

Should be change depending on the name of the app? 

CC @michaelstingl

jesmrec:
(14) is fixed @hosy

hosy:
## 16

If you select `Browse Log Files` and choose a log file by selecting the table row, nothing is happening. You have to exactly touch the accessory view for activate the sharing action. On a real device this touch area is very small.

Expected result:
Sharing action should also be activated by selecting the table row.

hosy:
## (17)
Public Links does not show password text field in edit mode, if a password is set

hosy:
## (18)
If a file will be overwritten, before it was synced, there is an error message with a format string inside.

Simulator Screen Shot - iPhone SE - 2019-06-18 at 12 05 03

jesmrec:
### (19)

1. User1 shares with User2 a non-empty folder, granting sharing permissions
2. User2 browses the shared folder to share with User3 any item inside

Current: Items are not allowed to be re-shared, only the parent folder is.
Expected: Items can be shared individually, because the parent folder is granted "Share" permission

mneuwert:
### (20)

Glitches when creating new folder:

1. Create new folder
2. As soon as placeholder folder icon appears, tap it
3. New view is opened and it shows first the sub-folder with same name and then message "Folder removed"
4. Go back to the parent folder and the new folder is visible twice.

out

mneuwert:
@jesmrec Fixed (20) with 70bd66cdea5c6845a2ff5025d495d340e98aac00 commit

felix-schwarz (approved):
Approved without review to allow merging.

hosy (approved):

Issue #395, opened 2019-05-21
Implement media player using AVKit


mneuwert:
Right now media files are opened using `WKWebView` which facilitates some rudimentary media playback. It would be however much nice to have a proper media player used to play audio and video files and even showing album artwork as requested in #374.

This can be easily done with `AVPlayerViewController`:
https://developer.apple.com/documentation/avkit/avplayerviewcontroller

hosy:
@mneuwert would be cool to have this. This would also have the advantage to directly stream the file, if the media file is locally not available!?

michaelstingl:
Yeah, [MP4 Streaming](https://rigor.com/blog/optimizing-mp4-video-for-fast-streaming) should be supported if the media files has the correct format.

mneuwert:
@hosy @michaelstingl yeah, should be possible!

jesmrec:
Done in #429

PR #429, opened 2019-06-30
Media player implemented using AVKit

mneuwert:
## Description
Instead of using webView, AVPlayerViewController is from now on used to play media content (audio and video)

## Related Issue
#395 

## Motivation and Context
We want to use AVKit for media playing to provide better UI and also to be able to stream media content if it is not available locally.

## Types of changes

## Checklist:


ownCloud/Client/Viewer/Media/MediaDisplayViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+
+	override func viewDidLoad() {
+		super.viewDidLoad()
+		self.requiresLocalItemCopy = false
+	}
+
+	override func renderSpecificView() {
+		if source != nil {
+
+			let asset = AVURLAsset(url: source!, options: self.httpAuthHeaders != nil ? ["AVURLAssetHTTPHeaderFieldsKey" : self.httpAuthHeaders!] : nil )
```
hosy: if you use `if let source = source` before, you can remove force unwrapping here

mneuwert: @hosy fixed in 74ec9adc270a144663aa045e6bc8ddded8583f1b

hosy:
@mneuwert is it possible to move the player controls above the safe area, because it is not really possible to move the player position?

IMG_9384

mneuwert:
> @mneuwert is it possible to move the player controls above the safe area, because it is not really possible to move the player position?

@hosy unfortunatelly, you have no control over how controls are positioned within `AVPlayerViewController` view

hosy:
> > @mneuwert is it possible to move the player controls above the safe area, because it is not really possible to move the player position?
> 
> @hosy unfortunatelly, you have no control over how controls are positioned within `AVPlayerViewController` view

We need a other solution, because slider control is not usable.
Possible Solutions:
1. Add a toolbar on the bottom of the view
Con: wasting space

2. Add custom player controls and hide the default controls
Pro: we can use our own design
Con: more effort needed

I would like to go with option 2. What do you think @mneuwert ?

mneuwert:
> > > @mneuwert is it possible to move the player controls above the safe area, because it is not really possible to move the player position?
> > 
> > 
> > @hosy unfortunatelly, you have no control over how controls are positioned within `AVPlayerViewController` view
> 
> We need a other solution, because slider control is not usable.
> Possible Solutions:
> 
> 1. Add a toolbar on the bottom of the view
>    Con: wasting space
> 2. Add custom player controls and hide the default controls
>    Pro: we can use our own design
>    Con: more effort needed
> 
> I would like to go with option 2. What do you think @mneuwert ?

@hosy I will play with it and check what's possible

mneuwert:
@hosy I've pushed a fix which makes sure that safe area is considered and player controls are not covered. Can you please check if this is what you have expected?

jesmrec:
### (1)

While a video is streamed, options over the video are available and can lead to weird behaviours. For example, if you rename the video while it is being streamed, an error happens and streaming stops (the video URL changed with the name, so the error makes sense but is ugly for users). If the video is moved, the streaming is also stopped. 

Different options:

1. Disabling the three-dot-button, if feasible. During streaming, file can not be handled.
2. Disabling those operations that modify the source file (move and rename). Other operations do not change the playing file and can remain as enabled.

jesmrec:
### (2)

With the streaming, the only way to download (locally) a media file is using the "Open In" option.  We can improve this in a separate issue, but imho the media files should be downloadable as any other file. 

Options:

- Add a `Download` option in the Card (for every file, or only for media files)
- Add a new Setting to choose: "Download or stream media files" in Settings.
- Both of them

what do you think? maybe out of scope here, just for clarify.

jesmrec:
Also, just for clarifying, artwork is mentioned here: https://github.com/owncloud/ios-app/issues/395#issue-446569652, but in audio streaming, i do not see it. Is it implemented here?

jesmrec:
### (3)

1. Stream a video
2. From other client or web UI, rename it

Current: Video is played twice (two background audios)
Expected: Different options, related also with (1): forbid to rename, only one streaming, or stop the streaming

mneuwert:
> Also, just for clarifying, artwork is mentioned here: [#395 (comment)](https://github.com/owncloud/ios-app/issues/395#issue-446569652), but in audio streaming, i do not see it. Is it implemented here?

@jrecio Artwork stuff was not discussed in scope of this change.

mneuwert:
### Re: (2) 
- Right now you can still download media files by tapping on the cloud icon.
- We could add an option in settings. Easily doable.

mneuwert:
### Re: (1) and (3)
These are fixed with the commit 92befda5f44de3406d70e22b669087cac1f0bdfc. The only thing is that after renaming, the video is paused and user has to hit playback button again. But I don't know if it is really an issue.

jesmrec:
Regarding (1) and (3):

It is OK for me to stop the streaming after renaming. 

Regarding (2):

I am not able to download the file by using the cloud icon. If i tap in the cloud, the streaming is triggered (small clicking area?). About the Setting, i will open a separate issue to discuss the details, if you agree.

mneuwert:
@jesmrec added a setting which controls how media files are handled (default: files are downloaded like just any other type of file). User can enable media file streaming in settings menu.

jesmrec:
### (4)

Installed from scratch (streaming disabled), tap on a media file. 

Current: File is downloaded but not played, showing with an error. Cloud mark is there, so file was not really downloaded

Jul-24-2019 12-05-55

NOTE: not all files show that behaviour. Let me know if you need the files i reproduced with.

mneuwert:
### Re: (4)
Can't reproduce with both files you provided.

jesmrec:
It is not really easy to reproduce afterwards. I checked with other containers and i was not able. Will set as [NOT REPRODUCIBLE]

jesmrec:
Approved on my side

Issue #426, opened 2019-06-30
'version.hide' => true in config.php results in error "This server runs an unsupported version (  )…"


michaelstingl:
```
/**
 * While hardening an ownCloud instance hiding the version information in status.php
 * can be a legitimate step. Please consult the documentation before enabling this.
 */
'version.hide' => false,
```
Source: https://github.com/owncloud/core/blob/v10.2.0/config/config.sample.php#L91-L95

### Steps to reproduce

1. 
```
docker run -it --rm \
  --name oc1020 \
  -p 10208:8080 \
  -e OWNCLOUD_VERSION_HIDE=true \
owncloud/server:10.2.0
```
2.
```
$ curl http://localhost:10208/status.php | json_pp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   123  100   123    0     0   5356      0 --:--:-- --:--:-- --:--:--  5590
{
   "installed" : true,
   "versionstring" : "",
   "maintenance" : false,
   "needsDbUpgrade" : false,
   "productname" : "",
   "version" : "",
   "edition" : ""
}
```
3. Add new account with server URL `http://localhost:10208`

### Expected behaviour

ownCloud app should connect

### Actual behaviour

<img width="834" alt="Screenshot 2019-06-30 at 15 31 41" src="https://user-images.githubusercontent.com/214010/60397370-c5364a00-9b4c-11e9-8399-0f08c84abe03.png">

### Client
**iOS version:**
13.0

**ownCloud app version:**
1.0.1 from App Store

**Device model:**
iPad Pro (10.5-inch)

### Server configuration

```
docker run -it --rm \
  --name oc1020 \
  -p 10208:8080 \
  -e OWNCLOUD_VERSION_HIDE=true \
owncloud/server:10.2.0
```

jesmrec:
Fixed!

PR #432, opened 2019-07-01
[feature/version-hide] Add support for version.hide server setting

felix-schwarz:
## Description
Adds support for `version.hide` server setting.

## Related Issue
#426

## Types of changes

### DIFF
--- a/ownCloud/Bookmarks/BookmarkViewController.swift
+++ b/ownCloud/Bookmarks/BookmarkViewController.swift
@@ -509,7 +509,7 @@ class BookmarkViewController: StaticTableViewController {
 
 			let connection = OCConnection(bookmark: bookmark)
 
-			connection.connect { [weak self] (error, _) in
+			connection.connect { [weak self] (error, issue) in
 				if let weakSelf = self {
 					if error == nil {
 						bookmark.displayName = connection.loggedInUser?.displayName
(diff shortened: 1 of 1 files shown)


ownCloud/Bookmarks/BookmarkViewController.swift
```diff
@@ -536,7 +536,26 @@ class BookmarkViewController: StaticTableViewController {
 					} else {
 						OnMainThread {
 							hudCompletion({
-								weakSelf.presentingViewController?.dismiss(animated: true, completion: nil)
+								if issue != nil {
+									self?.bookmark?.authenticationData = nil
+
+									let issuesViewController = ConnectionIssueViewController(displayIssues: issue?.prepareForDisplay(), completion: { [weak self] (response) in
```
hosy: `[weak self]` is here not needed, because you already using `weakself` from above

hosy (changes requested):

ownCloud/Bookmarks/BookmarkViewController.swift
felix-schwarz: That `[weak self]` is to make the reference to `self` weak in the `ConnectionIssueViewController` completion block. `weakSelf` above is actually a strong reference to self derived from a previous `[weak self]`. I'll change the naming to make this less confusing.

felix-schwarz: Changed the naming in the latest commit.

hosy (approved):

jesmrec:
Approved, now accounts are acceded even when the version string in `status.php` is hidden

jesmrec (approved):

Issue #443, opened 2019-07-08
Add a "Create Folder" button in ClientDirectoryPickerViewController


hosy:
### Steps to reproduce
1. Move or Copy an item in the App
2. `ClientDirectoryPickerViewController` will appear
3. This view should have to option to create a new folder in the selected directory

### Expected behaviour
No option to add a new folder

### Actual behaviour
Button on the left top to create a new folder

### Client
**iOS version:**
12.3

**ownCloud app version:**
1.0.2 (125)

jesmrec:
Let's QA this

jesmrec:
### (1)

I am able to share content from the Safari browser, for example a PDF file. But, in Photos app, the ownCloud app does not appear to share with. do i miss something? 

Safari:
![Screenshot 2019-07-24 at 16 35 14](https://user-images.githubusercontent.com/14894746/61802603-0d394b80-ae31-11e9-9f7b-dfe47a10737e.png)

Photos app:

![Jul-24-2019 16-37-07](https://user-images.githubusercontent.com/14894746/61802762-538eaa80-ae31-11e9-99b2-cfe9f15f8105.gif)

of course, both screenshots were taken with the same device with the ownCloud app installed.

jesmrec:
only as a tip, out of scope of this issue: If a user has two accounts in the same server (f.ex., an user account and an admin account) with the same full name, accounts can not be distinguished:

![Screenshot 2019-07-24 at 16 56 59](https://user-images.githubusercontent.com/14894746/61804355-14158d80-ae34-11e9-91ec-d1ceb9dc13ab.png)

this is out of scope because the same issue happens in the account list. i will open a separate issue about that.

Issue #76, opened 2018-06-11
Receive a file Shared from other app


javiergonzper:
As a user, I want the possibility to send a file from an app to ownCloud iOS app to be uploaded to my ownCloud Sever.

### Technical info:
We need to implemente the `Share extension` technical info at the [Apple Documentation](https://developer.apple.com/library/archive/documentation/General/Conceptual/ExtensibilityPG/Share.html)

### AC:

- [ ] Show the ownCloud app on the list of files to share a supported file
- The supported files should be:
     - [ ] Documents (pdf, doc, txt, ppt...)
     - [ ] Images
     - [ ] Videos
- To prevent delays and other issues the limits on the share should be:
     - [ ] Documents: 3
     - [ ] Images: 20
     - [ ] Videos: 3
- [ ] If the passcode is active ask it
- [ ] If the biometrical unlock is active ask it
- [ ] If it is not possible lock the rotation, the passcode should be support landscape and portrait
- [ ] Give to the user the possibility to navigate in order to select the folder where upload the files
- [ ] Give to the user the possibility to create a new folder where upload the files
- [ ] QA

### Open to discussion:
The UI depends of this because we have here 2 options:
- Upload the file from the "Share" UI. We have to design an upload view to see the progress and cancel the upload
- Send the file to ownCloud in order to be uploaded later. This is the option implemented on the old iOS app.

michaelstingl:
@javiergonzper In the old app, it's pretty confusing. In some apps, I have 2 different options:

| Voice Memos | Photos | *.PDF in Safari |
|--- |---|---|
![img_3489](https://user-images.githubusercontent.com/214010/41231078-2b271ad2-6d82-11e8-8688-aa4b272e62e2.png) | ![img_3789](https://user-images.githubusercontent.com/214010/41231079-2b47938e-6d82-11e8-98f8-b46465da68a7.png) | ![img_3787](https://user-images.githubusercontent.com/214010/41231080-2b705b2a-6d82-11e8-8800-69e3db9ea748.png) |

javiergonzper:
@michaelstingl are different things to do more or less the same. 

- Share extension is the one that appear just like "ownCloud"
- Open In ownCloud it is the one that appear like "Copy to ownCloud"

The "Open in ownCloud" is the old way to upload files to ownCloud from other app. Was implemented before the "Extensions" exist on iOS.
If I remember correctly it was on iOS6 and extensions appear on iOS7.

[...]

PR #445, opened 2019-07-08
[feature/share-sheet] Import files from share sheet / Create Folder in Directorypicker

hosy:
## Description
User can import files using the iOS share sheet. all file types a…

## Related Issue
#76 
#443 

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Import/ImportFilesController.swift
+++ b/ownCloud/Import/ImportFilesController.swift
@@ -94,7 +94,8 @@ extension ImportFilesController {
 												at: targetDirectory,
 												from: url,
 												isSecurityScoped: false,
-												options: [OCCoreOption.importByCopying : true],
+												options: [OCCoreOption.importByCopying : true,
+														  OCCoreOption.automaticConflictResolutionNameStyle : OCCoreDuplicateNameStyle.bracketed.rawValue],
 												placeholderCompletionHandler: { (error, item) in
 													if error != nil {
 														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path)), error: \(error?.localizedDescription ?? "" )")
--- a/ownCloud/Import/ImportFilesController.swift
+++ b/ownCloud/Import/ImportFilesController.swift
@@ -23,6 +23,8 @@ class ImportFilesController: NSObject {
 
 	// MARK: - Instance variables
 	var url: URL
+	var localCopyContainerURL: URL?
+	var localCopyURL: URL?
 	var fileIsLocalCopy: Bool
 
 	// MARK: - Init & Deinit
@@ -44,7 +46,7 @@ extension ImportFilesController {
 			if let appGroupURL = OCAppIdentity.shared.appGroupContainerURL {
 				let fileManager = FileManager.default
 
-				var inboxUrl = URL(fileURLWithPath: appGroupURL.appendingPathComponent("File-Import").path)
+				var inboxUrl = appGroupURL.appendingPathComponent("File-Import")
 				if !fileManager.fileExists(atPath: inboxUrl.path) {
 					do {
 						try fileManager.createDirectory(at: inboxUrl, withIntermediateDirectories: false, attributes: nil)
@@ -64,11 +66,13 @@ extension ImportFilesController {
 						return false
 					}
 				}
+				self.localCopyContainerURL = inboxUrl
 
 				inboxUrl = inboxUrl.appendingPathComponent(url.lastPathComponent)
 				do {
 					try fileManager.copyItem(at: url, to: inboxUrl)
 					self.url = inboxUrl
+					self.localCopyURL = inboxUrl
 				} catch let error as NSError {
 					Log.debug("Error creating directory \(inboxUrl) \(error.localizedDescription)")
 					return false
@@ -85,9 +89,9 @@ extension ImportFilesController {
 
 			if bookmarks.count > 1 {
 				let moreViewController = self.cardViewController(for: url)
-				if let window = UIApplication.shared.delegate?.window {
-					let viewController = window!.rootViewController
-					if let navCon = viewController as? UINavigationController, let viewController = navCon.visibleViewController {
+				if let delegateWindow = UIApplication.shared.delegate?.window, let window = delegateWindow {
+					let viewController = window.rootViewController
+					if let navigationController = viewController as? UINavigationController, let viewController = navigationController.visibleViewController {
 						OnMainThread {
 							viewController.present(asCard: moreViewController, animated: true)
 						}
@@ -137,29 +141,32 @@ extension ImportFilesController {
 
 	func importFile(url : URL, to targetDirectory : OCItem, bookmark: OCBookmark, core : OCCore?) {
 		let name = url.lastPathComponent
-		if let progress = core?.importFileNamed(name,
-												at: targetDirectory,
-												from: url,
-												isSecurityScoped: false,
-												options: [OCCoreOption.importByCopying : true,
-														  OCCoreOption.automaticConflictResolutionNameStyle : OCCoreDuplicateNameStyle.bracketed.rawValue],
-												placeholderCompletionHandler: { (error, item) in
-													if error != nil {
-														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path)), error: \(error?.localizedDescription ?? "" )")
-													}
-													// Return OCCore
-													OCCoreManager.shared.returnCore(for: bookmark, completionHandler: nil)
-		},
-												resultHandler: { (error, _ core, _ item, _) in
-													if error != nil {
-														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path)), error: \(error?.localizedDescription ?? "" )")
-													} else {
-														Log.debug("Success uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path))")
+		if core?.importFileNamed(name,
+					 at: targetDirectory,
+					 from: url,
+					 isSecurityScoped: false,
+					 options: [OCCoreOption.importByCopying : true,
+						   OCCoreOption.automaticConflictResolutionNameStyle : OCCoreDuplicateNameStyle.bracketed.rawValue],
+					 placeholderCompletionHandler: { (error, item) in
+						if error != nil {
+							Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path)), error: \(error?.localizedDescription ?? "" )")
+						}
 
-														self.removeLocalCopy()
-													}
-		}) {
-		} else {
+						OnBackgroundQueue(after: 2) {
+							// Return OCCore after 2 seconds, giving the core a chance to schedule the upload with a NSURLSession
+							OCCoreManager.shared.returnCore(for: bookmark, completionHandler: nil)
+						}
+					 },
+					 resultHandler: { (error, _ core, _ item, _) in
+						if error != nil {
+							Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path)), error: \(error?.localizedDescription ?? "" )")
+						} else {
+							Log.debug("Success uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path))")
+
+							self.removeLocalCopy()
+						}
+					}
+		) == nil {
 			Log.debug("Error setting up upload of \(Log.mask(name)) to \(Log.mask(targetDirectory.path))")
 		}
 	}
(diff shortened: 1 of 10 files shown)


hosy:
Would be great if you can give me some feedback, which version you like. This screen will be shown, if `Open In` was triggered from an other app to import a file into the ownCloud app and if the user has more than one account configured.
The user will see this view, after the user will kicked into our app.

In my opinion the CardView is more custom and gives more informations but the AlertView is more clean and more a view to decide something. What do you think?

| CardView                                                                                                                                                         | AlertView                                                                                                                                                        |
|------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Simulator Screen Shot - iPhone X - 2019-07-09 at 22 35 40 | Simulator Screen Shot - iPhone X - 2019-07-09 at 22 58 24 |

michaelstingl:
Alert view works for me. I‘d tend to the simple solution, then iterate based on user feedback.

mneuwert:
i would vote for the card view but if alert is simpler then we could use it and then go for a new iOS13 style context menu as soon as we start working on iOS13 related enhancements

hosy:
@michaelstingl @mneuwert both solutions are implemented and can be tested via bitrise build

hosy:
Changed the CardView and added a description row and also changed the title:
Simulator Screen Shot - iPhone X - 2019-07-11 at 22 03 15

ownCloud/Import/ImportFilesController.swift
```diff
@@ -0,0 +1,154 @@
...
+			return true
+		}
+
+		return false
+	}
+
+	func importItemWithDirectoryPicker(with url : URL, into bookmark: OCBookmark) {
+		OCCoreManager.shared.requestCore(for: bookmark, setup: { (_, _) in
+		}, completionHandler: { (core, error) in
+			if error == nil {
```
felix-schwarz: Rather than using `core!` a few lines below, please use `if let core = core, error == nil` here for added safety.

ownCloud/Import/ImportFilesController.swift
```diff
@@ -0,0 +1,154 @@
...
+
+						if let targetDirectory = selectedDirectory {
+							self?.importFile(url: url, to: targetDirectory, core: core)
+						}
+					})
+
+					let pickerNavigationController = ThemeNavigationController(rootViewController: directoryPickerViewController)
+
+					if let wd = UIApplication.shared.delegate?.window {
+						let vc = wd!.rootViewController
```
felix-schwarz: Please use `window` instead of `wd` and `viewController` instead of `vc` to match the code style of the rest of the code base.

ownCloud/Import/ImportFilesController.swift
```diff
@@ -0,0 +1,154 @@
...
+						if let navCon = vc as? UINavigationController, let topVC = navCon.visibleViewController {
+							topVC.present(pickerNavigationController, animated: true)
```
felix-schwarz: Please use `navigationController` instead of `navCon` and `topViewController` instead of `topVC` to match the code style of the rest of the code base.

ownCloud/Import/ImportFilesController.swift
```diff
@@ -0,0 +1,154 @@
...
+extension ImportFilesController {
+
+	func accountOrImportUI() -> Bool {
+		let bookmarks : [OCBookmark] = OCBookmarkManager.shared.bookmarks as [OCBookmark]
+
+		if bookmarks.count > 1 {
+			let moreViewController = self.cardViewController(for: url)
+			if let wd = UIApplication.shared.delegate?.window {
+				let vc = wd!.rootViewController
+				if let navCon = vc as? UINavigationController, let topVC = navCon.visibleViewController {
```
felix-schwarz: Please use `window` instead of `wd`, `viewController` instead of `vc`, `navigationController` instead of `navCon` and `topViewController` instead of `topVC` to match the code style of the rest of the code base.

ownCloud/Import/ImportFilesController.swift
```diff
@@ -0,0 +1,154 @@
...
+
+	func importItemWithDirectoryPicker(with url : URL, into bookmark: OCBookmark) {
+		OCCoreManager.shared.requestCore(for: bookmark, setup: { (_, _) in
+		}, completionHandler: { (core, error) in
+			if error == nil {
+				OnMainThread {
+					let directoryPickerViewController = ClientDirectoryPickerViewController(core: core!, path: "/", selectButtonTitle: "Save here".localized, avoidConflictsWith: [], choiceHandler: { [weak self] (selectedDirectory) in
+
+						if let targetDirectory = selectedDirectory {
+							self?.importFile(url: url, to: targetDirectory, core: core)
```
felix-schwarz: `importFile` is never called here because there's `[weak self]` for this block a few lines above and the `ImportFilesController` is released before this block is run (testing with a single account setup, YMMV).

Please make `self` non-weak here.

ownCloud/Import/ImportFilesController.swift
```diff
@@ -0,0 +1,154 @@
...
+			self.importItemWithDirectoryPicker(with: url, into: bookmark)
+
+			return true
+		}
+
+		return false
+	}
+
+	func importItemWithDirectoryPicker(with url : URL, into bookmark: OCBookmark) {
+		OCCoreManager.shared.requestCore(for: bookmark, setup: { (_, _) in
```
felix-schwarz: The `requestCore` for the `OCCore` needs to be balanced with a `returnCore` in another place.

The best place for this would be in the `placeholderCompletionHandler` passed to `core.importFileNamed(…)` - with a short delay of 1 or 2 seconds to give `OCCore` a chance to schedule the background `NSURLSessionTask` to upload the file.

Putting it in the `completionHandler` passed to `core.importFileNamed(…)`, on the other hand, would guarantee that the core stays open until the upload has finished. However, if the user doesn't have connectivity, that could keep the core open indefinitely.

ownCloud/Import/ImportFilesController.swift
```diff
@@ -0,0 +1,154 @@
...
+		},
+												resultHandler: { (error, _ core, _ item, _) in
+													if error != nil {
+														Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path)), error: \(error?.localizedDescription ?? "" )")
+													} else {
+														Log.debug("Success uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path))")
+
+														let fileManager = FileManager.default
+														do {
+															try fileManager.removeItem(at: url)
```
felix-schwarz: If `true` was passed for [`UIApplicationOpenURLOptionsOpenInPlaceKey`](https://developer.apple.com/documentation/uikit/uiapplicationopenurloptionsopeninplacekey?language=objc), this would delete the original file. Please make sure to honor the value passed for [`UIApplicationOpenURLOptionsOpenInPlaceKey`](https://developer.apple.com/documentation/uikit/uiapplicationopenurloptionsopeninplacekey?language=objc) before deleting the file (possibly by using the approach I outlined in the review comment in `AppDelegate.swift`).

ownCloud/Resources/en.lproj/Localizable.strings
```diff
@@ -406,3 +406,7 @@
 "Delete Offline Copies" = "Delete Offline Copies";
 "Manage" = "Manage";
 "Storage" = "Storage";
+
+/* Import File */
+"Save File" = "Save File";
+"Select an account where to import the file and choose a destination directory." = "Select an account where to import the file and choose a destination directory.";
```
felix-schwarz: Just an alternative suggestion: `Choose an account and directory to import the file into.`

hosy: Used one of your description

ownCloud/Import/ImportFilesController.swift
hosy: Please can you share a code example?

felix-schwarz: Sure!

```swift
		if let progress = core?.importFileNamed(name,
							at: targetDirectory,
							from: url,
							isSecurityScoped: false,
							options: [OCCoreOption.importByCopying : true,
								  OCCoreOption.automaticConflictResolutionNameStyle : OCCoreDuplicateNameStyle.bracketed.rawValue],
							placeholderCompletionHandler: { (error, item) in
								if error != nil {
									Log.debug("Error uploading \(Log.mask(name)) to \(Log.mask(targetDirectory.path)), error: \(error?.localizedDescription ?? "" )")
								}

								// Return OCCore
								OCCoreManager.shared.returnCore(for: bookmark, completionHandler: nil)
							},
							resultHandler: { (error, _ core, _ item, _) in
…
```

jesmrec:
### (1)

I am able to share content from the Safari browser, for example a PDF file. But, in Photos app, the ownCloud app does not appear to share with. do i miss something? 

Safari:
Screenshot 2019-07-24 at 16 35 14

Photos app:

Jul-24-2019 16-37-07

of course, both screenshots were taken with the same device with the ownCloud app installed.

jesmrec:
only as a tip, out of scope of this issue: If a user has two accounts in the same server (f.ex., an user account and an admin account) with the same full name, accounts can not be distinguished:

Screenshot 2019-07-24 at 16 56 59

this is out of scope because the same issue happens in the account list. i will open a separate issue about that.

jesmrec:
### (2)

1. Select several files to share from an external app
2. Select an account to upload to

Current: only one file is uploaded
Expected: all files are uploaded

Jul-24-2019 17-28-39

michaelstingl:
> ### (1)
> I am able to share content from the Safari browser, for example a PDF file, Pages App, Files App... but, in Photos app, the ownCloud app does not appear to share with. do i miss something?

Yeah, I think iOS behaves different in some apps. We'd need to add a FAQ section to the docs and open a new issue for "Share Sheet v2"

mneuwert:
### Re (2)
This is a limitation of opening URLs in the app, the `AppDelegate.application(app, url, options)` method which is called in this case will get a single file URL even if multiple are selected. I guess we'll need to implement full blown sharing extension to get rid of this annoyance.

mneuwert:
### Re (1)
I tried to extend `LSItemContentTypes` key in Info.plist with more file types (public.image etc) but it had no effect on Photos app. Probably it is Apple's intention to not allow opening image url directly in another app.

jesmrec:
(1) :then, this is a [WONT FIX]

(2):

> I guess we'll need to implement full blown sharing extension to get rid of this annoyance.

inside the current PR? or separately. In case this will be addressed in another issue, we should take in account the limitation (CC @michaelstingl)

michaelstingl:
> inside the current PR? or separately. In case this will be addressed in another issue, we should take in account the limitation (CC @michaelstingl)

No, new PR. I'll give the Share Sheet v1 a try. Probably it's better than no share sheet. Then work on Share Sheet v2 …

michaelstingl:
While we still use the limited v1-Share-Sheet, we could help users to discover the "Save to Files" entry in the Share Sheet…

jesmrec:
### (3)

The icon to share with contains a different name in different apps, f. ex:

- ownCloud: "Copy in ownCloud"
- iCloud: "Open to ownCloud"

i guess it is not relevant, but it took my attention.

jesmrec:
### (4)

Having problems to share form Files App. In terms of UI, after sharing from Files App, the following log:

```
2019-08-01 08:16:41.012000+0200 ownCloud[284.006252] [ERRO] | [CORE, CommandLocalImport] Local creation for item <OCItem: 0x10a24f2a0, type: 0, name: Term Paper.pages, path: /Term Paper.pages, size: 0 bytes, MIME-Type: (null), Last modified: (null), Last used: (null) fileID: _placeholder_8E3A6DB1-21F9-41F0-9C19-D7C908C114E4, eTag: _placeholder_, parentID: 00002598ocfizvgliqbl, localID: 911A74292B8D412CB2D9B2BA1ADAB554, parentLocalID: D972724BF5454EAF9969FFE7A37CE767> from file:///private/var/mobile/Library/Mobile%20Documents/com~apple~Pages/Documents/Term%20Paper.pages to file:///private/var/mobile/Containers/Shared/AppGroup/B2AD5B71-5A1F-4A41-9B2F-2CDB4579DBEA/File%20Provider%20Storage/886C0A1F-576F-48E6-925C-E0CDAE8379ED/911A74292B8D412CB2D9B2BA1ADAB554/Term%20Paper.pages failed in move phase with error: Error Domain=NSCocoaErrorDomain Code=257 "The file “Term Paper.pages” couldn’t be opened because you don’t have permission to view it." UserInfo={NSFilePath=/private/var/mobile/Library/Mobile Documents/com~apple~Pages/Documents/Term Paper.pages, NSUnderlyingError=0x10a227ae0 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}} [OCCore+CommandLocalImport.m:139|FULL]
2019-08-01 08:16:41.021000+0200 ownCloud[284.006252] [dbug] | [APP] Error uploading Term Paper.pages to /, error: The file “Term Paper.pages” couldn’t be opened because you don’t have permission to view it. [ImportFilesController.swift:149|FULL]
2019-08-01 08:16:41.027000+0200 ownCloud[284.006252] [dbug] | [APP] Error uploading Term Paper.pages to /, error: The file “Term Paper.pages” couldn’t be opened because you don’t have permission to view it. [ImportFilesController.swift:159|FULL]
2019-08-01 08:16:41.028000+0200 ownCloud[284.006252] [dbug] | [APP] Error setting up upload of Term Paper.pages to / [ImportFilesController.swift:167|FULL]
```

felix-schwarz:
@jesmrec #471 fixes finding (4), along with two other issues (described in the PR).

settermjd:
I can't document this functionality, as it is not working in version 1.1 of the app, which I have for testing.

michaelstingl:
> I can't document this functionality, as it is not working in version 1.1 of the app, which I have for testing.

Please open a new issue and describe the problem.

Issue #444, opened 2019-07-08
Log files with version inside


cdamken:
People send me Log files and there is no line where it says which version is e testing.

It will be AMAZING if you add one line in the log files that says the version.

@michaelstingl @hosy

michaelstingl:
Cool idea. This is it how it looks like in the desktop log:

```
$ /Applications/owncloud.app/Contents/MacOS/owncloud --logfile - --logdebug
m_crashReporterChar:  /Applications/owncloud.app/Contents/MacOS/owncloud_crash_reporter
m_crashReporterWChar: /Applications/owncloud.app/Contents/MacOS/owncloud_crash_reporter
07-08 15:38:32:245 [ info gui.application ]:	"################## ownCloud locale:[en_DE] ui_lang:[] version:[2.6.0alpha1 (build 11615)] os:[macOS Mojave (10.14)]"
07-08 15:38:32:245 [ info gui.application ]:	Using "en_DE" translation
07-08 15:38:32:246 [ info gui.application ]:	Adding extra plugin search path: "/Applications/owncloud.app/Contents/MacOS/../PlugIns"
07-08 15:38:32:247 [ info gui.application ]:	VFS suffix plugin is available
07-08 15:38:32:290 [ info gui.socketapi ]:	server started, listening at  "4AP2STM4H5.com.owncloud.desktopclient.socketApi"
07-08 15:38:32:290 [ info gui.folder.manager ]:	setting remote poll timer interval to 30000 msec
07-08 15:38:32:900 [ debug gui.activity ]	[ OCC::ActivitySettings::setNotificationRefreshInterval ]:	Starting Notification refresh timer with  300  sec interval
07-08 15:38:32:901 [ info gui.application ]:	Tray menu workarounds: noabouttoshow: true fakedoubleclick: false showhide: false manualvisibility: true
07-08 15:38:32:914 [ info gui.folder.manager ]:	Setup folders from settings file
07-08 15:38:32:915 [ info plugins ]:	Created VFS instance from plugin "owncloudsync_vfs_suffix"
07-08 15:38:32:915 [ debug gui.folder ]	[ OCC::Folder::checkLocalPath ]:	Checked local path ok
07-08 15:38:32:915 [ info sync.configfile ]:	Adding system ignore list to csync: "/Applications/owncloud.app/Contents/Resources/sync-exclude.lst"
[...]

PR #446, opened 2019-07-08
[feature/log-info] Log device, version and locale info

felix-schwarz:
## Description
- Log device, version and locale information at the beginning of every log file using new SDK protocol
- Show SDK commit hash in Settings

## Example
A line like this is included near the top of every log file:

`2019-07-08 23:08:38.667000+0200 ownCloud[1253:538893] [info] | [LogIntro] Host: com.owncloud.ios-app 1.0.2 (125) #ea6c365; SDK: 1.0 (1) #49664d1; OS: iOS 12.3.1; Device: iPhone (iPhone8,1); Localizations: [de-DE, de]; Log options: level=Debug, destinations=["writer.stderr", "writer.file"], options=["option.log-requests-and-responses"], maskPrivateData=false [OCLogFileWriter.m:103|FULL]`

## Related Issue
#444 

## Types of changes

## Bugs & improvements

- [X] (1) Reset files do not create new headers https://github.com/owncloud/ios-app/pull/446#issuecomment-514127207
- [ ] (2) Header missing after switching log level to warning/error https://github.com/owncloud/ios-app/pull/446#issuecomment-514246941

### DIFF
--- a/ownCloud/Resources/en.lproj/Localizable.strings
+++ b/ownCloud/Resources/en.lproj/Localizable.strings
@@ -166,7 +166,7 @@
 "You need to configure an email account first to be able to send emails." = "You need to configure an email account first to be able to send emails.";
 "Do you want to open the following URL?" = "Do you want to open the following URL?";
 
-"%@ %@ version %@ build %@ (%@)" = "%@ %@ version %@ build %@ (%@)";
+"%@ %@ version %@ build %@ (app: %@, sdk: %@)" = "%@ %@ version %@ build %@ (app: %@, sdk: %@)";
 "beta" = "beta";
 "release" = "release";
 
--- a/ownCloud/Settings/MoreSettingsSection.swift
+++ b/ownCloud/Settings/MoreSettingsSection.swift
@@ -41,8 +41,8 @@ class MoreSettingsSection: SettingsSection {
 			buildType = "beta".localized
 		}
 
-		let localizedFooter = "%@ %@ version %@ build %@ (%@)".localized
-		let footerTitle = String(format: localizedFooter, OCAppIdentity.shared.appName!, buildType, VendorServices.shared.appVersion, VendorServices.shared.appBuildNumber, VendorServices.shared.lastGitCommit)
+		let localizedFooter = "%@ %@ version %@ build %@ (app: %@, sdk: %@)".localized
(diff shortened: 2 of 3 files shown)


hosy (approved):

jesmrec:
Checks performed:

- [X] Show log level in header
- [X] Show correct destinations in header
- [X] Show correct options in header
- [X] Show correct commit hashes
- [X] Show correct device info and versions
- [ ] Reset file -> by changing options, shown the previous ones

jesmrec:
### (1)

1. Create a log file with different options (f.ex., everything enabled: HTTP request and responses, standard error output, debug level, and log file)
2. Fill the log with some operations
3. Change options (f. ex, disable http requests and responses)
4. Reset log file by swiping 

Current: in the log header, Log options are set with the options in 1. 
Expected:  in the log header, Log options are set with the options in 3.

Not sure if that is an unexpected behaviour. Are the headers re-calculated when the log file is deleted individually?

felix-schwarz:
@jesmrec The log header is cached internally and therefore not updated. Will change that.

felix-schwarz:
@jesmrec The latest `milestone/1.1` commit should address this. Good catch!

jesmrec:
(1) -> Fixed

jesmrec:
### (2)

Check these steps out, similar to (1):

1. Set Debug level 
2. Delete the log file by swiping
3. Open the log file to check -> correct header 👍 

Then, if you switch the log level to...
 - ... info, and delete the log file again -> correct header
 - ... warning, and delete the log file again -> header is not there 
 - ... error, and delete the log file again -> header is not there  

That means, the new header is not there if the log file is reset after switching the log level to Warn|Error

felix-schwarz:
@jesmrec The header is currently logged with level `info`, so it's not included if limited only to `warning` and `error`. It would need to be logged at the `error` level to always be logged. I could also look into changing the code so it gets logged by the FileWriter itself, without passing through the Logger itself.

However, is it really desirable or necessary to log this info at anything other than the `debug` level? Because anything above that would likely be too incomplete for us to really be useful, and the user then be asked to create a new log at the debug level and send that in.

jesmrec:
I get your idea, so that issue is like a feature and not a bug. Feel free to improve it. From my side this is approved then.

Issue #454, opened 2019-07-16
Autocomplete starts at 4 characters


pmaier1:
- Version 1.0.2

**Actual behavior**
The user/group sharing autocomplete starts giving results after the user entered 4 characters, supposedly independent of the server settings.

**Expected behavior**
The user/group sharing autocomplete starts giving results after the user enters the configured number of characters on the server (default = 2).

See here for the config.php setting: https://github.com/owncloud/core/blob/master/config/config.sample.php#L302

hosy:
@pmaier1 for search the capabilities are used and just tested it with demo.owncloud.com.
Search will start after I entered the 3 letter. What setting do we have for demo.owncloud.com?

pmaier1:
> What setting do we have for demo.owncloud.com?

I don't know. Tried with my personal instance, that has the default value (2) and search started after I entered the 4th letter.

pmaier1:
> I don't know.

Well, actually you can find that pretty easily :D the WebUI autocomplete starts searching when you enter 2 characters.

pmaier1:
Just tried with Desktop client. That one starts searching after you enter one character. Looks like the API doesn't respect that setting or so :(

michaelstingl:
Just check settings of demo.owncloud.com with admin:admin

pmaier1:
> Just check settings of demo.owncloud.com with admin:admin

I did here ^^: https://github.com/owncloud/ios-app/issues/454#issuecomment-511873653

Presumably server issue: https://github.com/owncloud/enterprise/issues/3438

hosy:
@pmaier1 @michaelstingl 
thank you for your help. Just fixed the problem.

jesmrec:
Fixed in #455

PR #455, opened 2019-07-16
[fix/sharing-search] Fixed min length for searching sharing users

hosy:
## Description
Fixed min length for searching a user or group in sharing view.

## Related Issue
#454 

## How Has This Been Tested?
- Share File
- Search for a user or group
- Count letters, when activity indicator will appear for searching
- Check count result with server capability `user.search_min_length`

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/Sharing/GroupSharingTableViewController.swift
+++ b/ownCloud/Client/Sharing/GroupSharingTableViewController.swift
@@ -347,7 +347,7 @@ class GroupSharingTableViewController: SharingTableViewController, UISearchResul
 
 	func updateSearchResults(for searchController: UISearchController) {
 		guard let text = searchController.searchBar.text else { return }
-		if text.count >= core?.connection.capabilities?.sharingSearchMinLength?.intValue ?? 2 {
+		if text.count >= core?.connection.capabilities?.sharingSearchMinLength?.intValue ?? OCCapabilities.defaultSharingSearchMinLength {
 			if let recipients = recipientSearchController?.recipients, recipients.count > 0,
 				recipientSearchController?.searchTerm == text,
 				searchResultsSection == nil {
--- a/ownCloud/Client/Sharing/GroupSharingTableViewController.swift
+++ b/ownCloud/Client/Sharing/GroupSharingTableViewController.swift
@@ -347,7 +347,7 @@ class GroupSharingTableViewController: SharingTableViewController, UISearchResul
 
 	func updateSearchResults(for searchController: UISearchController) {
 		guard let text = searchController.searchBar.text else { return }
-		if text.count > core?.connection.capabilities?.sharingSearchMinLength?.intValue ?? 1 {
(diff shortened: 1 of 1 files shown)


ownCloud/Client/Sharing/GroupSharingTableViewController.swift
```diff
@@ -347,7 +347,7 @@ class GroupSharingTableViewController: SharingTableViewController, UISearchResul
 
 	func updateSearchResults(for searchController: UISearchController) {
 		guard let text = searchController.searchBar.text else { return }
-		if text.count > core?.connection.capabilities?.sharingSearchMinLength?.intValue ?? 1 {
+		if text.count >= core?.connection.capabilities?.sharingSearchMinLength?.intValue ?? 2 {
```
mneuwert: @hosy I think constant (e.g. defaultSharingSearchMinLength) would make sense instead of plain value 2

jesmrec:
Checks performed:

- [X] 1 character -> a request or every typed character
- [X] 3 character -> first request with the third character
- [X] 3 character: type 2, delete 1, type 1 -> no requests sent
- [X] 3 character: type 2, delete 1, type 2 -> request sent
- [ ] 3 character: type 3, delete 3, type 3 -> request sent -> needed 4

jesmrec:
### (1)

With: `'user.search_min_length' => 3`

1. Type 3 characters: request sent
2. Delete the characters by using the delete key of the keyboard (not the (x) in the edit field)
3. Type 3 characters

Current: No request send. Needed 4 characters
Expected: Request should be sent after typing the third character

Jul-23-2019 10-05-26

mneuwert:
### Re (1)
Actually the condition in the code is numberOfCharacters >= 2, so after typing 2nd character the request is sent. If the requirement is to send the search request after 3 characters, constant in `OCCapabilities` has to be adjusted. For me, following your test procedure, the request was consistently sent after 2nd typed character.

jesmrec:
This is the expected behaviour (i tested with), as far as i know:

- The number of characters to send  the first request is led by the capability (set in server's `config.php`). 

- It should not be hardcoded. If the number is 1, every typed character triggers a request. If number is 5, till the 5th character no request will be sent, then every character will send a request with the pattern. 

- If characters are removed from the pattern, a request will be sent every typed character while the number of typed characters is greater than the capability value. If it is lower, no request sent till the capability value is reached again.

let me know everything you think about.

mneuwert:
@jesmrec I found a problem: `OCRecipientSearchController.searchTerm` was only set when number of typed characters has reached the configured minimum. Internally the setter of this property checks if it is not the same string which is being assigned to the ivar. Only if it is different, actual search request would be triggered. So, if you typed always the same character sequences (ABC, ABC, ABC), then the search would be triggered only for the first sequence. I fixed this in 1cbecd4fdbf0d2886aca051596c21f1666f63160 which has required an SDK change as well.

jesmrec:
@mneuwert good catch, and fixed as well as (1)

Approved on my side

Issue #135, opened 2018-10-26
Available offline (folders)


jesmrec:
_As an user, i want to set a folder as available offline, so all the content inside will be downloaded and synced_

Following the line of https://github.com/owncloud/ios-app/issues/134, also folders have the property.

AC's are the same as the ones for files (check link above), but with some specific for folders:

- When a folder is set as av. offline, all the content inside is set as av. offline: files and subfolders. The line is kept recursively with all subfolders inside.
- A file or subfolder inside an av. offline folder, can not be unset as av. offline. Only the parent folder (the one that was set as av. offline) can be unset as av. offline.
- The av. offline operation can be cancelled before finishing. The files inside that were already downloaded will be wiped.
- By pulling down in the list of files, the av. offline folders will be synced. 
- An av. offline folder is av. offline even if it is moved to another location.
- If an av. offline folder is copied or duplicated, the new copy is not av. offline. 
- If any item (file or folder) is moved inside an av. offline folder, it gets the av. offline condition as everything inside the folder
- If any item (file or folder) inside an av. offline folder is moved/copied out of the av .offline folder, will not be av. offline.  

A first step here should be developing the `Download folder` feature. This will not be a feature itself, but will be needed. Should be implemented in a separate issue?

jesmrec:
Done

Issue #134, opened 2018-10-26
Available offline (files)


jesmrec:
_As an user, i want to set files as available offline, so the content will be downloaded and synced_

Available offline (or keep-it-sync') is a property that makes files to be:

- Downloaded and persisted on the device
- Always synced with the latest version available.

User does not have to take care about any of these actions after the moment the file is set as available offline. 

AC:

- Available offline is an option to be set in the three-dot-point menu
- Available offline is available in the preview
- File is marked in the file list as available offline with a gadget. Tipically, we can use:

https://github.com/owncloud/android/blob/master/res/drawable/offline_available_pin.png

- While the file is being synced, it is posible to mark the file with some icon that means such action
- As well as an option to make files available offline, other option to unset av. offline should be available.
- In case of multiselection available, av. offline should be an option to set several files at the time.
- N2H: a shortcut to all av. offline files in the account

jesmrec:
Done

PR #456, opened 2019-07-19
[feature/itempolicy] Item Policy / Available Offline support

felix-schwarz:
## Description
This PR adds Available Offline support throughout the app.

## Related Issue
#134 #135

### Actions

Make available offline | Make unavailable offline | Folder action
-- | -- | --
Bildschirmfoto 2019-07-19 um 09 51 03 | Bildschirmfoto 2019-07-19 um 09 51 58 | Bildschirmfoto 2019-07-19 um 09 53 20

### Badges

Regular | Available offline
-- | --
Bildschirmfoto 2019-07-19 um 09 54 44 | Bildschirmfoto 2019-07-30 um 22 24 12

### Quick Access View
TOC | No Available Offline | 
-- | --
Bildschirmfoto 2019-07-19 um 09 51 20 | Bildschirmfoto 2019-07-19 um 09 51 33

With one offline folder | Offline folder removal | All available offline items 
-- | -- | --
Bildschirmfoto 2019-07-19 um 09 51 17 | Bildschirmfoto 2019-07-19 um 09 51 31 | Bildschirmfoto 2019-07-19 um 09 51 24

### Settings
Settings | Options
-- | --
Bildschirmfoto 2019-07-30 um 22 28 00 | Bildschirmfoto 2019-07-30 um 22 28 04

## Types of changes

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.1.0/Available%20Offline.md

Bugs & improvements

- [X] (1) Crash propagated. Branches update needed https://github.com/owncloud/ios-app/pull/456#issuecomment-518722832 [FIXED]
- [X] (2) Downloaded files inside an av. offline folder https://github.com/owncloud/ios-app/pull/456#issuecomment-518730781
- [X] (3) Av. offline option in menu https://github.com/owncloud/ios-app/pull/456#issuecomment-518739202 [WONT FIX HERE]
- [X] (4) Multiselection in av. offline https://github.com/owncloud/ios-app/pull/456#issuecomment-518739986  [WONT FIX HERE]
- [X] (5) Moving an av. off line folder, loses the condition https://github.com/owncloud/ios-app/pull/456#issuecomment-518745617 [FIXED]
- [X] (6) Renaming an av. offline folder, makes the content losing the av. offline condition https://github.com/owncloud/ios-app/pull/456#issuecomment-518747805 [FIXED]
- [X] (7) Cancel download of av. offline file https://github.com/owncloud/ios-app/pull/456#issuecomment-518750368 [WONT FIX]
- [X] (8) Av. offline inside other av. offline https://github.com/owncloud/ios-app/pull/456#issuecomment-518752393 [WONT FIX]
- [X] (9) Wording in quick access https://github.com/owncloud/ios-app/pull/456#issuecomment-519033345  [DONE]
- [X] (10) View of "Delete unused copies" https://github.com/owncloud/ios-app/pull/456#issuecomment-519120637  [DONE]
- [X] (11) File is wiped even if it was opened during the time https://github.com/owncloud/ios-app/pull/456#issuecomment-519129105 [FIXED]
- [x] (12) Files must be downloaded every time the local copies are deleted https://github.com/owncloud/ios-app/pull/456#issuecomment-519134703 [FIXED]
- [X] (13) Content inside an av. offline subfolder is not marked https://github.com/owncloud/ios-app/pull/456#issuecomment-525170936 [FIXED]


ownCloud/Client/Actions/Actions+Extensions/MakeUnavailableOfflineAction.swift
```diff
@@ -0,0 +1,77 @@
...
+ *
+ */
+
+import UIKit
+import ownCloudSDK
+
+class MakeUnavailableOfflineAction: Action {
+	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.makeUnavailableOffline") }
+	override class var category : ActionCategory? { return .normal }
+	override class var name : String? { return "Make unavailable offline".localized }
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Actions/Actions+Extensions/MakeAvailableOfflineAction.swift
```diff
@@ -0,0 +1,70 @@
...
+ *
+ */
+
+import UIKit
+import ownCloudSDK
+
+class MakeAvailableOfflineAction: Action {
+	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.makeAvailableOffline") }
+	override class var category : ActionCategory? { return .normal }
+	override class var name : String? { return "Make available offline".localized }
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Actions/Actions+Extensions/CreateFolderAction.swift
```diff
@@ -22,7 +22,7 @@ class CreateFolderAction : Action {
 	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.createFolder") }
 	override class var category : ActionCategory? { return .normal }
 	override class var name : String? { return "Create Folder".localized }
-	override class var locations : [OCExtensionLocationIdentifier]? { return [ .plusButton ] }
+	override class var locations : [OCExtensionLocationIdentifier]? { return [ .folderAction ] }
```
hosy: Please check capitalization of "Create Folder". I only found "Create folder" in localizable.strings file

ownCloud/Client/Library/Item Policies/ItemPolicyCell.swift
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	required init?(coder aDecoder: NSCoder) {
+		fatalError("init(coder:) has not been implemented")
+	}
+
+	// MARK: - Item policy item
+	override func titleLabelString(for item: OCItem?) -> String {
+		if (itemResolutionPath as NSString?)?.isRootPath == true {
+			return "Root folder".localized
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Library/Item Policies/ItemPolicyCell.swift
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		if let item = item {
+			return super.titleLabelString(for: item)
+		} else {
+			return "\((itemResolutionPath as NSString?)?.lastPathComponent ?? "") \("(no match)".localized)"
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Library/Item Policies/ItemPolicyCell.swift
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+
+	override func detailLabelString(for item: OCItem?) -> String {
+		if itemPolicy?.localID != nil, let itemPath = item?.path {
+			return "\("at".localized) \(itemPath)"
+		} else if let itemPolicyPath = itemPolicy?.path as NSString?, itemPolicyPath.length > 0 {
+			return "\("at".localized) \(itemPolicyPath)"
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Library/Item Policies/ItemPolicyTableViewController.swift
```diff
@@ -0,0 +1,225 @@
...
+				return false
+			}) ?? []
+		})
+	}
+
+	var itemPolicies : [OCItemPolicy] = [] {
+		didSet {
+			OnMainThread {
+				if self.itemPolicies.count == 0 {
+					self.messageView?.message(show: true, imageName: "icon-available-offline", title: "Available Offline".localized, message: "No items have been selected for offline availability.".localized)
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Library/Item Policies/ItemPolicyTableViewController.swift
```diff
@@ -0,0 +1,225 @@
...
+			}
+		} else {
+			return 0
+		}
+	}
+
+	override func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+		if let section = ItemPolicySectionIndex(rawValue: section) {
+			switch section {
+				case .all: 	return "Overview".localized
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Library/Item Policies/ItemPolicyTableViewController.swift
```diff
@@ -0,0 +1,225 @@
...
+		}
+	}
+
+	override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+		if let section = ItemPolicySectionIndex(rawValue: indexPath.section) {
+			switch section {
+				case .all:
+					let cell = tableView.dequeueReusableCell(withIdentifier: "metaCell", for: indexPath) as? ThemeTableViewCell
+
+					cell?.textLabel?.text = "All Files".localized
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Library/LibraryTableViewController.swift
```diff
@@ -377,6 +379,15 @@ class LibraryTableViewController: StaticTableViewController {
 				})
 			})
 
+			addCollectionRow(to: section, title: "Available Offline".localized, image: UIImage(named: "cloud-available-offline")!, query: nil, actionHandler: { [weak self] (completion) in
```
hosy: Localized string is missing in localizable.strings file

ownCloud/Client/Actions/ClientItemResolvingCell.swift
```diff
@@ -0,0 +1,78 @@
...
+				self.itemResolutionLocalID = nil
+				self.itemTracker = core?.trackItem(atPath: atPath, trackingHandler: { (error, item, isInitial) in
+					if error == nil, let item = item, isInitial {
+						OnMainThread {
+							self.item = item
+							self.itemTracker = nil // unless isInitial is removed from above "if", that's it - we can stop tracking as any updates won't be used anyway
+						}
+					} else if (error != nil) || (item == nil) {
+						OnMainThread {
+							self.resolutionFailed(error: error)
```
hosy: This function does not contain code. Remove this line and the function or should this function have any code inside?

ownCloud/Settings/StorageSettingsSection.swift
```diff
@@ -0,0 +1,133 @@
...
+
+		localCopyExpirationViewController.navigationItem.title = "Storage".localized
+
+		if let availableStyles = ThemeStyle.availableStyles {
+			var themeIdentifiersByName : [[String:Any]] = []
+
+			for style in availableStyles {
+				themeIdentifiersByName.append([style.localizedName : style.identifier ])
+			}
+
```
hosy: Are you sure this `ThemeStyle.availableStyles` `if`-block is necessary? Maybe used from other SettingsSection?

felix-schwarz: No, absolutely not. 😂 Thanks, will remove 🤓

ownCloud/Client/Actions/Actions+Extensions/CreateFolderAction.swift
felix-schwarz: Good catch! Fixed.

ownCloud/Client/Actions/ClientItemResolvingCell.swift
felix-schwarz: It's just a subclassing point for now, with no default functionality.

In case we have an item cell that's resolving its item from a path in the future (by subclassing ClientItemResolvingCell) and we want to handle the case that it can't be resolved, that'd be where that case would be handled.

felix-schwarz:
@hosy Fixed the findings that I marked as resolved with the latest commit.

jesmrec:
### (1) 

Crash reported here (https://github.com/owncloud/ios-app/pull/466#issuecomment-517651499) also happening in the current branch, by propagation. Please, update the current/parent branch with `master`, where the fix currently is.

jesmrec:
### (2)

1. Browse inside a folder with several files
2. Download manually some files
3. Browse back 
4. Make the folder of 1. available offline

Current: Downloaded files in 2. do not have the av. offline badge. Only the non-downloaded ones are marked as av. offline
Expected: The content inside an av. offline folder is av. offline, regardless it was downloaded before or not

jesmrec:
### (3) (suggestion/discussion)

About the option in the menu:

Screenshot 2019-08-06 at 18 13 32

Currently, the option makes av. offline everything in the current folder. The action performed is the same as making av. offline the container folder. 

I'd use that option to make available offline all the items in the current folder, forgetting the parent folder. I mean, the same as setting av. offline one by one all the items in the screen.

jesmrec:
### (4) (improvement)

Is there any choice to use the av. offline option inside the multiselection? Currently, five actions available, maybe a sixth one would overload the bottom bar. On the other side, is useful to select several items at the time to make them av. offline. What do you think? 

@hosy @michaelstingl @mneuwert

jesmrec:
### (5)

This behaviour was not specified here: https://github.com/owncloud/ios-app/issues/135. Will add moving items' rules.

Steps:

1. Set a folder as av. offline
2. Move the folder to other location

Current: folder loses the av. offline 
Expected: An av. offline folder is av. offline wherever it is placed

jesmrec:
### (6)

1. Set a folder as av. offline 
2. Rename the folder

Current: folder keeps the av. offline badge, but everything inside lose it.
Expected: Folder and its content is av. offline after renaming.

jesmrec:
### (7)

1. Set a huge file as av. offline. It starts to download
2. Cancel the download

Current: Download starts again. No way to stop it, only waiting till ends 
Expected: Set a wat to mark as unavailable offline during the transfer (in the swipe?). Here, ideas are welcome.

jesmrec:
### (8)

1. Make a file inside a folder as av. offline
2. Browse back and make the folder av. offline

Current: file in step 1. is av. offline individually, not as part of the parent folder. In can be set as unavailable offline, regardless the parent is available offline.
Expected: file is av. offline as part of the av. offline parent folder. So, it can not be set unavailable offline.

jesmrec:
### (9) improvement

Open Quick Access -> Av. offline.

By swiping any item, the option `Remove` is available to set as unavailable offline. This is confusing, since `Remove` means "get rid of this".  I'd use a different wording:

- Unavailable offline (maybe too long)
- Unset
- Disable swiping in the list, so no options available there (maybe too strict)
- ... (other ideas)

felix-schwarz:
@jesmrec 
**(1)** I've updated the `develop` and `feature/itempolicy` SDK branches with the latest changes from `master`. Should be resolved now.

**(2)** That's currently by design. Thereby downloads retain their behaviour depending on how they were last downloaded. So if a file is downloaded due to user interaction, it'll stay around until it has expired. And if a file is downloaded via available offline, its lifetime is managed by its requested available offline status.

That ensures that if a user downloads a file, then switches its folder to available offline and back to unavailable offline, the user-requested file will still be around.

It'd not be hard to change that, though, and switch over existing, user-initiated downloads to ones managed by available offline. 

**(3) + (4)** I think having this action in this place act on the current folder is the right thing to do. This is currently also the only way to take the entire account available offline.

If the action only applied to all items currently in view, it'd not be possible to take the entire account available offline and newly added items and directories would require the user to make each of them available offline.

Also, what option would be presented if there already are available offline items in the currently shown folder? Both Make Available Offline and Make Unavailable Offline actions would need to be shown then.

I think a better way to support the use case of making all items in a folder available offline is through adding multi-selection support as proposed by you in (4).

**(5)** Yes, currently available offline *folders* are tracked by path (removing the rule when the folder name changes or the folder is removed), and available offline *files* are tracked by `LocalID` (so they are tracked across renames and moves).

**(6)** Good catch. That's clearly a bug.

**(7)** I don't see a good way to handle this otherwise. If a user marks a folder as available offline, the app is supposed to make the contents of that folder available offline, incl. retrying downloads that failed or were cancelled. If a user doesn't want a large file available offline, they'd need to pick the other items they want available offline rather than the whole folder.

Technically, available offline support already supports rules that could be used to exclude single files or all items that match certain criteria (f.ex. file size larger than X). It'd require a different, more advanced UI to expose those capabilities to users though.

**(8)** Available Offline is stored per item, so an item that's already been made Available Offline inside a folder that has been made Available Offline afterwards will retain that individual Available Offline status, incl. when the enclosing folder is made Unavailable Offline again or when the item is moved to another folder that's not Available Offline.

I see this as really useful functionality as it allows to have a subset of items and folders always Available Offline, and selectively take their enclosing folders Available Offline temporarily.

An example would be a sales representative that typically just needs `/Contracts/Template.doc` available offline, to f.ex. start penning new contracts during commute, but needs the whole `/Contracts` folder available offline for quick access when going to an event for clients where connectivity is really bad or non-existent.

That'd work by making `/Contracts` Available Offline. If the sales representative subsequently takes the folder Unavailable Offline again, `/Contracts/Template.doc` would still remain on the device without extra steps.

The SDK already has functionality to spot and report technically redundant Available Offline entries (like `/Contracts/Template.doc` when taking `/Contracts` available offline). However, I figured there's no real benefit to bringing up alerts for this, and let users choose - or even silently removing those redundant Available Offline entries. It'd just interrupt the user's work and (worst case) take away flexibility to deal with situations like the one described above. If a user really has the desire to remove redundancies, they can always remove individual entries through the table found under Quick Access.

---

What I'd change at this point:
**(2)** Switch over existing, user-initiated downloads to ones managed by available offline.
**(5)** Look into adding support for moved folders.
**(6)** Fix that bug.

jesmrec:
At this point , with commit `1477457`

(1)  -> Fixed

(2) ->

(3) + (4) ->

>  I think having this action in this place act on the current folder is the right thing to do. This is currently also the only way to take the entire account available offline.

So, another way is keeping the option only in root folder (if feasible)

> Also, what option would be presented if there already are available offline items in the currently shown folder? Both Make Available Offline and Make Unavailable Offline actions would need to be shown then.

following the approach in other clients, both. It is a good solution, though.

> I think a better way to support the use case of making all items in a folder available offline is through adding multi-selection support as proposed by you in (4).

maybe iterate and create (thinking in a preliminar design) a new issue to develop separately. CC @hosy @michaelstingl 

(5) -> does it mean that the reported behaviour is the expected one?

(6) -> Fixed for files. Still reproducible for folders 

(7) -> will get the current approach. Even, if we set an option to "make unavailable offline" as in (3), will fix the problem.

(8) -> Thanks for the explanation. I've notice that unsetting an av. offline file inside an av. offline folder, the file keeps being av. offline as part of the folder. Such behaviour is OK from my side. 

Keep executing tests

jesmrec:
### (10) (improvement)

Regarding the "Delete unused copies" in Settings.

In terms of design i'd go for a different approach. Currently, we have a list of 14 options from "never" to "after 1 year". Maybe is more flexible and less complex form users to set a selector with two fields:

- Number of units 
- Kind of unit (minute, hour, day, month, year).

what do you think?

jesmrec:
### (11)

1. Set "Delete unusual local copies" to 1 minute
2. Download a file
3. Open the file several times

Current: After 1 minute of downloading, the file is locally deleted even when it was opened several times.

Is that one the expected?

jesmrec:
### (12)

1. Set a huge file as av. offline
2. In the bookmark list -> manage -> delete local copies

Current: the huge file must be downloaded again (time and data plan wasted). In case you are offline during step 2. , file will not be available.

Expected: Available offline content must be pinned to the device. The only way to be able to remove the local copy is making it unavailable offline.

jesmrec:
### (12)

1. Set a huge file as av. offline
2. In the bookmark list ->swipe ->  manage -> delete local copies

Current: the huge file must be downloaded again (time and data plan wasted). In case you are offline during step 2. , file will not be available.

Expected: Available offline content must be pinned to the device. The only way to be able to remove the local copy is making it unavailable offline. If such behaviour is the expected for the Setting "Delete unused copies", the option in the "Manage" section must be clarified as "Remove everything, no matter the status"

felix-schwarz:
Thanks for the feedback @jesmrec!

As of https://github.com/owncloud/ios-app/pull/456/commits/700e7825cae08a50c672cc23e319c5e213000760:
- addressed findings 2, 5 and 6
- there's a known issue with 5 whereby the app may trigger a removal and re-download of items in moved available offline folders after the move. This needs to be addressed in a different place in the core and is something I'll take care of.

Regarding the new issues:
- **(10)** Thinking about putting a slider there. Or a picker, like is used in Calendar when entering dates.
- **(11)** Good catch. I assume that's a side effect of the newer viewer-related changes that don't always go through `download`, which updates that last used date. Will take care of this.
- **(12)** I can see good cases both for deleting and not deleting available offline files in that case. Maybe we should ask the user whether available offline files should also be removed.

jesmrec:
At this point

(2), (5) and (6) fixed.

About the other stuff:

(3), (4) -> will check how is the best choice if multiselection is implemented. I would move multiselection for av. offline to the next milestone, after the first implementation is accepted by users. Is this the expected for the first implementation @michaelstingl @hosy ?

(9), (10), (11): @felix-schwarz on it.

(12) -> i will think a bit about, because it is not a good idea to mess this up with many different options. Will share with you my ideas about when i polish them.

jesmrec:
About (12), local copies removal. With current implementation, users have two options:

- In bookmark list, remove the local copies inside one specific account. Everything is removed, no matter if it is av. offline or not.
- In Settings, remove local copies through all accounts once the time has passed. An. offline stuff is not removed.

Some thoughts about:

- As described in (12), av. offline files might not be available in offline. This is conflicting.
- As a matter of internal storage, maybe is not important for the user, from which account is the stuff deleted. Because of this, both options seem duplicated because they do the same. 
- Av. offline concept should be simple: those files and folders set as av. offline are pinned to the device till the user sets them as unavailable offline. Automatic wipe or removal of local copies should never remove local copies. Otherwise, they'd be regular downloaded files.

felix-schwarz:
> - there's a known issue with (5) whereby the app may trigger a removal and re-download of items in moved available offline folders after the move. This needs to be addressed in a different place in the core and is something I'll take care of.

That known issue is fixed as of https://github.com/owncloud/ios-app/pull/456/commits/9ade2506f16ba6ec95d660e1102e3783788bbca5. @jesmrec

jesmrec:
(5) -> fixed, included the issue with the extra download after moving.

felix-schwarz:
As of https://github.com/owncloud/ios-app/pull/456/commits/aad26748852ba68ccc1e086976747b88af03f6d1 (10) and (11) is fixed.

felix-schwarz:
As of https://github.com/owncloud/ios-app/pull/456/commits/ca2774d8d0e4e9625002342f3ee62428366ea025 (9) is addressed.

felix-schwarz:
As of https://github.com/owncloud/ios-app/pull/456/commits/536a5f0028258c6a11a1bad2d0f28a5588391525 (12) is addressed as well.

felix-schwarz:
@jesmrec (9), (10), (11) and (12) are fixed now.

Regarding (3) + (4): I'd like to address these by including them in multi selection eventually, but the current multi selection toolbar is already packed with 5 items. Not sure tap targets won't become too small if we try to add more there. This may need further discussion, so I suggest to break these out into an issue.

jesmrec:
(9), (10), (11) -> fixed.
(3) , (4) -> will open a separate issue

About (12):

When the switch "Include av. offline files" is off, works fine. Av. offline content stays at the device, and downloaded content is wiped out. 

When the switch is on:

- Content is initially wiped ("Offline files use" indicator moves to 0 bytes), but av. offline content in the list keeps the mark and then,  "Offline files use" goes to previous value because everything av. offline is downloaded again (expected behaviour?). If content is automatically re-downloaded, which is here the use case? 

- The switch is always off every time you enter to "Manage". If you switch it on, then close the view and later enter again, it is switched off. Must it be switched on manually every time you want to wipe? shouldn't it be persisted as a setting?

felix-schwarz:
> (3) , (4) -> will open a separate issue (#485)

Thanks!

> About (12):
> 
> When the switch "Include av. offline files" is off, works fine. Av. offline content stays at the device, and downloaded content is wiped out.
> 
> When the switch is on:
> 
> * Content is initially wiped ("Offline files use" indicator moves to 0 bytes), but av. offline content in the list keeps the mark and then,  "Offline files use" goes to previous value because everything av. offline is downloaded again (expected behaviour?).

The expected behaviour is that:
* when turning this option on, the Available Offline files are removed from the device
* when logging into the account next time, Available Offline downloads the files again

> If content is automatically re-downloaded, which is here the use case?

It's for situations where a user needs to free up space on the device temporarily and bandwidth is not a concern.

Without this option, users would have to 
1) log into the account
2) manually remove Available Offline items
3) *perform whatever other action they need additional storage for*
4) go back to the account
5) re-add all Available Offline items

> * The switch is off every time you enter to "Manage". If you switch it on, then close the view and later enter again, it is switched off. Must it be switched on manually every time you want to wipe? shouldn't it be persisted as a setting?

Not saving this was a conscious design decision:
1) wiping Available Offline files is considered an exception, not something people do frequently
2) using the switch, the user has to confirm they know the files will be re-downloaded
3) not persisting the setting ensures that the user can't accidentally delete Available Offline files after using the option previously

jesmrec:
about the re-downloading caused by the switch: i got your idea. But, if we are going to remove files that should be available even offline, we should (at least) warn the user. 

Actually, an alert is shown, letting users know that everything will be downloaded again in the following connection. But, if you are offline, nothing will be downloaded and there will be available offline content that is not available (the original problem). If the mentioned one is going to be the expected behaviour, i'd warn the user explicitly about this issue ("if you do that, your files will not be available till you recover the connection", or similar).

Adding such short clarification, or modifying a bit the current alert will be enough. My suggestion:

`Files and folders marked as available offline will be re-downloaded next time you log into your account. If the connection is lost, files marked as available offline will not be available during the offline status. They will be  downloaded again once the connection is back"

felix-schwarz:
@jesmrec Agreed! I tried to write it in less words as `Files and folders marked as Available Offline will become unavailable. They will be re-downloaded next time you log into your account (connectivity required)`. What do you think?

jesmrec:
For me it's ok the short version

felix-schwarz:
@jesmrec As of https://github.com/owncloud/ios-app/pull/456/commits/ecfcd0c53203bea01934a59b92943c8b94e6da86:
- the message text is now the extended version
- the app now takes advantage of the SDKs new support for pausing available offline downloads, to pause them after removing local copies of available offline files, until the user logs into the account in the app again

jesmrec:
I performed these steps to check the behaviour:

1. Set as av. offline a big file
2. While the download is running, remove local copies including av. offline
3. Perform some operations in other bookmarks
4. Open the bookmark of file in step 1.

What i see: ongoing download was paused (not removed), and resumes.

felix-schwarz:
@jesmrec Sorry for my imprecise wording.

What I wanted to say is not that it pauses the downloads themselves, but it deactivates the mechanism that schedules new downloads until the bookmark is opened in the app again.

Ongoing downloads of files technically live outside the respective bookmarks, in the shared HTTP Pipeline, until they are finished. They are therefore not covered by the "Remove local copies" functionality.

To verify the mechanism is paused, you can perform these steps:
1) Make f.ex. the "Photos" folder available offline and wait for downloads to finish
2) Go to Account > Manage, activate "Include Available Offline files", confirm and tap "Remove local copies"
3) Go to the Files app and open the account there. If the Available Offline mechanism wasn't paused, that would trigger a re-download of all AO files.

Since the File Provider may be active before and after tapping "Remove local copies", and could immediately restart downloads in the background, the AO mechanism only resumes operation when logging in via the app. Otherwise, it'd not be reliably possible to implement the option to remove local copies of Available Offline files.

jesmrec:
Thanks for the clarification. Checked the steps, worked fine.

jesmrec:
### (13)

1. Set a folder as av. offline (all its content becomes av. offline)
2. Create a folder inside the av. offline folder (the created folder is now av. offline)
3. Upload some pics inside the subfolder in 2.

Current: the content is not marked as av. offline (no badge next to the file name)
Expected: the content is marked as av. offline

jesmrec:
Everything fixed and correct. Great job!!

PR #459, opened 2019-07-22
[fix/auth] Fix issues in handling and improve handling of auth errors

felix-schwarz:
## Description
- Authentication error produce only a single alert now
- Added special handling and messages for
	- Account disabled server side
	- Authentication data missing for bookmark (f.ex. after a device restore from an unencrypted backup)
- Fix retain cycle leading to zombie `OCCore` when hitting `Edit`

## Authentication flow
If a `401 Unauthorized` error is returned, the following flow takes place in the SDK:
- the `AuthenticationMethod` root class creates an `AuthorizationFailed` error and - if the response is parsable as a XML SabreException - also includes error messages in the error object
- the `AuthenticationMethodOAuth2` subclass tries to directly resolve the error by requesting a new `auth_token` using the `refresh_token`
	- if a new `auth_token` could be retrieved, tells the HTTP Pipeline to retry the request that got the `401 Unauthorized` response
	- if no new `auth_token` could be retrieved, passes on the `AuthorizationFailed` error, which then gets sent to the App for handling

The App then handles the `AuthorizationFailed` error as follows:
- if the SabreException XML contained the error message `User disabled`, brings up an alert to tell the user that the account has been disabled, with the single choice being `Continue offline`
- else the app presents an alert informing the user that authentication failed, providing the user with the choice to either `Ignore` the error (and continue offline) or `Edit` the connection, which ends the session and brings up the account edit UI, where the user can provide new credentials or - if OAuth2 was used - re-authenticate with the server to receive a new `auth_token` / `refresh_token` token pair

## Related Issue
https://github.com/owncloud/enterprise/issues/3406

## Screenshots
Auth data missing | Account disabled
-- | --
Bildschirmfoto 2019-07-22 um 10 02 00 | Bildschirmfoto 2019-07-22 um 09 38 55

## Types of changes

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.1.0/ImprovedOAuth2.md

- [X] (1) Error after removing user in basic auth https://github.com/owncloud/ios-app/pull/459#issuecomment-515374025 [FIXED]


jesmrec:
### (1) improvement - error handling

Check these steps out:

1. Login (basic) with an user
2. In web UI, remove the user
3. In app, select "Sing In" in the dialog
4. Save

Current:

Screenshot 2019-07-26 at 11 10 42

Expected: Friendly message, depending on which information can be caught from exception

iPhoneX v12, commit `1042a18`, server 10.2.1

felix-schwarz:
@jesmrec https://github.com/owncloud/ios-app/pull/459/commits/b2a6c608021863c68df348de0b03706101f3da46 should fix (1)

jesmrec:
(1) fixed

approved on my side

Issue #386, opened 2019-05-17
Background task manager


felix-schwarz:
We have a couple of background features on our roadmap like f.ex.:
- background photo upload
- background available offline sync
- background file tree update

What they all share is the need to run code periodically while the app is in the background - and/or based on app state change events.

## Architecture

Since the above list will certainly grow in the future, I think it'd be great to take a modular approach to this, possibly structured like this:

- `BackgroundTaskExtension`s, based on the `OCExtension` mechanism:
    - leverage the `location` and `features` properties to find matching extensions
    - the `location.identifier` part could represent certain time events (like `launch`, `periodic`, `didBecomeBackgrounded`, `didComeToFront`, …)
    - the `features` part could be used to define restrictions (like `lowBattery`, `wifiAvailable`, `backgroundOnly`, …) on when the action the extension represents should or shouldn't be run
    - `provideObjectForContext` returns `BackgroundTaskAction` instances
    - `BackgroundTaskExtensionContext` (an `OCExtensionContext` subclass) could provide additional information for `BackgroundTaskExtension`s to make decisions on whether they want their action be run at a certain time
- `BackgroundTaskAction` encapsulates an action to run (f.ex. a block). It provides a method to indicate completion (possibly calling a completionHandler underneath), so it can run asynchronously. To the extent that result values are needed (I think this is the case for background tasks), they could be implemented as properties that the `BackgroundTaskManager` would then access on task completion. Main benefit of having these as properties rather than completion-method parameters would be: new ones can be added without breaking existing code + providing default values.
- a `BackgroundTaskManager` that ties all of this together
    - provides methods for the respective time events (like `launch`, `periodic`, `didBecomeBackgrounded`, `didComeToFront`, …)
    - creates a `BackgroundTaskExtensionContext` describing the respective event and then uses that to query `OCExtensionManager` for matching `BackgroundTaskExtension`s
    - asks the `BackgroundTaskExtension`s to provide `BackgroundTaskAction`s using the same context, then takes care of running these actions

## Modularity / Custom features for clients

Every background (or timed) feature could then be encapsulated into its own `BackgroundTaskExtension` and no worries be spent on adding calls to every such feature's implementation in the various places.

Like with the `OCExtension`-based `Action`s system we already have in the app, this would be especially useful when creating special features for clients that shouldn't find their way into the main/public code base.

felix-schwarz:
[...]

Issue #116, opened 2018-09-25
Instant uploads 


jesmrec:
_As user, i want that the content taken for the camera is upload directly to a selected folder, so i do not have to take care of uploading them manually_

Mock already defined and available in [WIP] in the app

![screen shot 2018-09-25 at 10 54 36](https://user-images.githubusercontent.com/14894746/46003801-9ec11b00-c0b1-11e8-98d2-28a383654dd0.png)

AC:

- Pics and videos handled separately
- Under 4G only if option "only wifi" is not selected. 
- Target folder selected. If it does not exist (because it was removed before or other reasons), it is created again and the content uploaded into. Folder displayed below the option name
- Only is uploaded the content taken while the option is enabled (videos/pics separately). If it is disabled, no content upload anymore
- Possibility to upload in background, needed location services in such case. If not, content will be uploaded when the app comes to foreground. Option is displayed when (at least) one of pics/videos is enabled.

michaelstingl:
> Please consider to implement a default sync profile when you have more than one sync profiles present.

@mmattel Could you elaborate? What are the characteristics of such a sync profile? How/when to change them?

mmattel:
Imagine having accounts on three different ownclouds. Lets say one private and two business.
Now you of course want to define which account will be the one defined as target for you automated picture upload. Independent of which account you currently have activated, the pictures should only go into the defined one and not the active one. Not waiting that the defined one gets activated, upload in the background to the defined while working with the activated !

jesmrec:
you mean, a default account to upload to, not the active one. This needs to think about. 

I'd go for different stages to be developed separately/incrementally

1. instant uploads to the selected folder
2. instant uploads with only wifi
3. instant uploads with background support
4. option to set a default account or the active one

more ideas welcome!

PR #484, opened 2019-08-15
Feature/task scheduling

mneuwert:
## Description
- First version of task scheduler which takes into account app states (backgrounded, foregrounded background fetch) and several events (WiFi available, low power, external power connected, photo library has changed etc). Those contribute to the execution context. Tasks are scheduled matching this context.
- Currently 2 different tasks are introduced:

1) Instant photo upload triggered when photo library has new assets and user is connected to WLAN network.
2) Bookmark update task triggered by background fetch event scheduled by iOS.

### Details regarding instant photo upload
- New settings allowing to enable / disable photo / video upload separately
- For instant media upload account and upload path has to be selected
- Changes in the photo library are detected when app goes into foreground (no signification location change detection magic as in some other apps to make it work in background). However as soon as `NSURLSession` instances are created uploads will continue in the background.
- Currently if user removes the folder he specified for instant upload nothing will happen and the the upload task will just silently exit.
- When user first activates instant upload, timestamp of the activation is stored and compared against modification date of the assets to be uploaded. As soon as one asset is successfully uploaded, its modification timestamp will be used to update instant upload activation timestamp to prevent that assets will be uploaded twice and to remove the need to keep track of uploaded items in a kind of database.
- Media conversion settings are taken into account by instant upload feature.

## Related Issue
#386 
#116 

## Motivation and Context
Achieve feature parity with an old app and provide a generic mechanisms allowing to schedule automated tasks based on defined conditions.

## Types of changes

## Checklist:

## QA

Test plan [WIP]: 

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.1.0/Instant%20Uploads.md

Bugs & improvements:

- [X] (1) Stucked in Account view https://github.com/owncloud/ios-app/pull/484#issuecomment-522001983 [FIXED]
- [X] (2) Pics are uploaded more than one time https://github.com/owncloud/ios-app/pull/484#issuecomment-522562887 [FIXED]
- [X] (3) Settings makes crashes after removing the latest account with i.u. enabled https://github.com/owncloud/ios-app/pull/484#issuecomment-523357946  [FIXED]
- [X] (4) Account or upload path can be empty https://github.com/owncloud/ios-app/pull/484#issuecomment-523416947 [FIXED]
- [X] (5) Account and items can not be deleted https://github.com/owncloud/ios-app/pull/484#issuecomment-523418902 [FIXED]
- [X] (6) Videos not uploaded https://github.com/owncloud/ios-app/pull/484#issuecomment-523424309 [FIXED]
- [x] (7) Images not uploaded in background https://github.com/owncloud/ios-app/pull/484#issuecomment-523499215 [FIXED]
- [X] (8) Re-attaching an account, does not set the option for account and path https://github.com/owncloud/ios-app/pull/484#issuecomment-524219916 [FIXED]


jesmrec:
### (1)

1. Install the app form scratch
2. Before attaching an account, tap on Settings
3. Enable instant uploads
4. Tap on "Account"

Current: App stucked there, no way to come back. Only killing
Expected: Some way or breadcrumb to cancel or go back to Settings

jesmrec:
### (2)

1. Set the instant uploads (account + path)
2. Take two pics 
3. Delete those pics
4. Take other two pics

Current: Four pics (2 pics of step 2. and 2 pics of step 4.) are uploaded
Expected: Only pics in step 4. are uploaded, since pics in step 2. had already been uploaded.

mneuwert:
### Re: (2)
@jesmrec I tried to reproduce the same scenario in simulator and it works as expected for me. But there I didn't use camera (since not possible). I imported pictures from mac into simulator's photo library and by issuing touch <image.jpg> on command line to make sure that all pictures have different modification date. I will check later with the real camera on physical iPhone.

ownCloud/Settings/MediaUploadSettingsSection.swift
```diff
@@ -44,12 +54,85 @@ extension UserDefaults {
...
+		set {
+			self.set(newValue, forKey: MediaUploadKeys.InstantUploadPhotosAfterDateKey.rawValue)
+		}
+
+		get {
+			return self.value(forKey: MediaUploadKeys.InstantUploadPhotosAfterDateKey.rawValue) as? Date
+		}
+	}
+
+	public var instantUploaVideosAfter: Date? {
```
felix-schwarz: This should be `instantUploadVideosAfter`.

ownCloud/Settings/MediaUploadSettingsSection.swift
```diff
@@ -17,14 +17,24 @@
...
 
 	enum MediaUploadKeys : String {
 		case ConvertHEICtoJPEGKey = "convert-heic-to-jpeg"
 		case ConvertVideosToMP4Key = "convert-videos-to-mp4"
+		case InstantUploadPhotosKey = "instant-upload-photos"
+		case InstantUploadVideosKey = "instant-upload-videos"
+		case InstantUploadBookmarkUUIDKey = "instant-upload-bookmark-uuid"
+		case InstantUploadPathKey = "instant-upload-path"
+		case InstantUploadPhotosAfterDateKey = "instant-upload-photos-after-date"
+		case InstatnUploadVideosAfterDateKey = "instant-upload-videos-after-date"
```
felix-schwarz: This should be `InstantUploadVideosAfterDateKey`.

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,149 @@
...
+											if finished {
+												self.completed()
+											}
+										})
+									}
+								}
+
+								// Upload new videos
+								if let uploadVideosAfter = userDefaults.instantUploaVideosAfter {
+									if let photoFetchResult = self.fetchAssetsFromCameraRoll(.videos, createdAfter: uploadVideosAfter) {
```
felix-schwarz: The code for videos and images looks very similar. Maybe, it could be shared between both?

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,149 @@
...
+										var assets = [PHAsset]()
+										photoFetchResult.enumerateObjects({ (asset, _, _) in
+											assets.append(asset)
+										})
+										MediaUploadQueue.shared.uploadAssets(assets, with: core, at: rootItem, assetUploadCompletion: { (asset, finished) in
+											if let asset = asset {
+												userDefaults.instantUploadPhotosAfter = asset.modificationDate
+											}
+											if finished {
+												self.completed()
```
felix-schwarz: Seeing the code calls `self.completed()` for both possible uploads - photos and videos - this could potentially be called twice.

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,149 @@
...
+								}
+
+								// Upload new videos
+								if let uploadVideosAfter = userDefaults.instantUploaVideosAfter {
+									if let photoFetchResult = self.fetchAssetsFromCameraRoll(.videos, createdAfter: uploadVideosAfter) {
+										var assets = [PHAsset]()
+										photoFetchResult.enumerateObjects({ (asset, _, _) in
+											assets.append(asset)
+										})
+										MediaUploadQueue.shared.uploadAssets(assets, with: core, at: rootItem, assetUploadCompletion: { (asset, finished) in
```
felix-schwarz: I remember there were issues with high memory usage back when upload of multiple media items was implemented. As far as I can see from the code, the way `MediaUploadQueue.shared.uploadAssets` is used here, videos and photos could end up being uploaded / processed concurrently.

Would it be an option to first determine new photos and videos, then put them all into one array - and then submit them for upload? That would prevent concurrency at this point.

mneuwert:
@jesmrec Latest commit 727b7536a94bacaf2291ccd01e61a221f72974ef makes use of `PHAsset.creationDate` to determine which assets to upload. With this change (2) can't be reproduced anymore.

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,149 @@
...
+
+		guard userDefaults.instantUploadPhotos == true || userDefaults.instantUploadVideos == true else { return }
+
+		guard let bookmarkUUID = userDefaults.instantUploadBookmarkUUID else { return }
+
+		guard let path = userDefaults.instantUploadPath else { return }
+
+		if let bookmark = OCBookmarkManager.shared.bookmark(for: bookmarkUUID) {
+
+			OCCoreManager.shared.requestCore(for: bookmark, setup:nil, completionHandler: { (core, _) in
```
felix-schwarz: The `request` to `OCCoreManager` for a core is not balanced with a `returnCore`. Please return the core to the `OCCoreManager` when the task is done.

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,149 @@
...
+				if let core = core {
+					core.delegate = self
+
+					let pathQuery = OCQuery(forPath: path)
+					pathQuery.changesAvailableNotificationHandler = { query in
+						if query.state == .idle {
+							core.stop(query)
```
felix-schwarz: Starting a query will trigger a full `PROPFIND` for the directory, which could cause quite some traffic if there are already many files in that directory.

You can avoid that overhead by using `-[OCCore trackItemAtPath:(OCPath)path trackingHandler:(void(^)(NSError *error, OCItem *item, BOOL isInitial))trackingHandler]`  instead. If there already is an `OCItem` for that path, it'll return it from cache immediately, without making a potentially [CPU/memory/time]-expensive `PROPFIND`.

ownCloud/Client/Tasks/ScheduledTaskManager.swift
```diff
@@ -0,0 +1,243 @@
...
+			}
+
+			let context = OCExtensionContext(location: location, requirements: requirements, preferences: nil)
+
+			// Find a task to run
+			if let matches = try? OCExtensionManager.shared.provideExtensions(for: context) {
+				for match in matches {
+					if let task = match.extension.provideObject(for: context) as? ScheduledTaskAction {
+						if self.state == .backgroundFetch {
+							task.backgroundFetchCompletion = fetchCompletion
```
felix-schwarz: With `task.backgroundFetchCompletion` called in `task.completion()`, the `fetchCompletion` block will be called multiple times (and possibly prematurely) if there's more than one extension running.

AFAIK the `fetchCompletion` block provided by iOS should only called once - after all tasks have finished.

ownCloud/Client/Tasks/ScheduledTaskExtension.swift
```diff
@@ -0,0 +1,119 @@
...
+				self.backgroundFetchCompletion = nil
+			}
+
+			if self.backgroundTaskIdentifier != .invalid {
+				UIApplication.shared.endBackgroundTask(self.backgroundTaskIdentifier)
+			}
+		}
+	}
+
+	func allowedToRun() -> Bool {
```
felix-schwarz: This could probably be turned into a property via `var allowedToRun : Bool {`. Right now, it doesn't seem to be used anywhere (but probably should, to return from background fetch in time).

ownCloud/Settings/MediaUploadSettingsSection.swift
```diff
@@ -72,6 +155,180 @@ class MediaUploadSettingsSection: SettingsSection {
 
 		self.add(row: convertPhotosSwitchRow!)
 		self.add(row: convertVideosSwitchRow!)
+
+		// Instant upload requires at least one configured account
+		OCBookmarkManager.shared.loadBookmarks()
```
felix-schwarz: Not sure if this `loadBookmarks()` call is needed. They should already have been loaded at that point.

ownCloud/Settings/MediaUploadSettingsSection.swift
```diff
@@ -72,6 +155,180 @@ class MediaUploadSettingsSection: SettingsSection {
...
+				}, title: "Upload Path".localized, value: "", accessoryType: .disclosureIndicator, identifier: MediaUploadSettingsSection.pathSelectionRowIdentifier)
+
+			self.add(row: instantUploadPhotosRow!)
+			self.add(row: instantUploadVideosRow!)
+		}
+
+		updateDynamicUI()
+	}
+
+	private func getSelectedBookmark() -> OCBookmark? {
```
felix-schwarz: Not needed for now, but would have been cool to have a kind of standardized/generic picker for an account + path combination that could also be used in other places.

ownCloud/Client/Tasks/ScheduledTaskManager.swift
```diff
@@ -0,0 +1,243 @@
...
+	// MARK: - Background fetching
+
+	func backgroundFetch(completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+		self.state = .backgroundFetch
+		scheduleTasks(fetchCompletion: completionHandler)
+	}
+
+	// MARK: - Private methods
+
+	private func scheduleTasks(fetchCompletion:((UIBackgroundFetchResult) -> Void)? = nil, completion:((_ scheduledTaskCount:Int)->Void)? = nil) {
```
felix-schwarz: There's currently no protection against:
- a task getting scheduled more than once at the same time (which definitely *can* make sense, but should be opt-in from the Task's side)
- the `state` changing *during* scheduling (could pose a problem for Tasks that rely on being called for every event they sign up for)

That led to a new photo being uploaded twice when I added an image to the Photo Library via the ownCloud app's "Open In" action. The photo upload task was apparently run twice in parallel.

ownCloud/PhotKit Extensions/MeidaUploadQueue.swift
```diff
@@ -0,0 +1,76 @@
+//
+//  MeidaUploadQueue.swift
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz this must have been a late night coding session:) Fixed.

ownCloud/Client/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,111 @@
+//
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz Fixed

ownCloud/AppDelegate.swift
```diff
@@ -53,6 +53,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
 		FileProviderInterfaceManager.shared.updateDomainsFromBookmarks()
 
+		_ = ScheduledTaskManager.shared
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/AppDelegate.swift
```diff
@@ -84,6 +89,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 			UIView.setAnimationsEnabled(enableUIAnimations)
 		}
 
+		UIApplication.shared.setMinimumBackgroundFetchInterval(
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Settings/MediaUploadSettingsSection.swift
mneuwert: @felix-schwarz I think, that doesn't happen necessarily if the first thing user does after opening the app is navigating to settings menu

jesmrec:
### (3)

1. Install the app from scratch and attach an account
2. In settings, enable the instant uploads (no matter the set up, only enable it)
3. Remove the account created in 1.
4. Tap on Settings (bottom right corner)

Current: app crashes.

`2019-08-21 10:37:42.265000+0200 ownCloud[359:000000] [info] | [OS] Fatal error: Unexpectedly found nil while unwrapping an Optional value`

Expected: no crash

iOS12

jesmrec:
@mneuwert 

`For instant media upload account and upload path has to be selected`

Does this mean, that in case the user does not select both account and path, the uploads will not be performed, even if the setting is enabled?

mneuwert:
> @mneuwert
> 
> `For instant media upload account and upload path has to be selected`
> 
> Does this mean, that in case the user does not select both account and path, the uploads will not be performed, even if the setting is enabled?

Yes, user has to select account / path, otherwise app doesn't know where to upload.. But may be we should change the UI in the way that setting account / path is enforced, otherwise the switch shall be disabled if e.g. user cancels the selection process.

jesmrec:
> Yes, user has to select account / path, otherwise app doesn't know where to upload.. But may be we should change the UI in the way that setting account / path is enforced, otherwise the switch shall be disabled if e.g. user cancels the selection process.

yes. i was thinking in that kind of solution. The problem is when the target account is deleted or the target folder is renamed/moved/deleted, then the i.u. are still enabled but not i.u. are done.  

Posible solutions:

- If the target account is removed from the app, disable instant uploads and reset values. Warn the user with an alert to select another target.
- If the target folder is moved/renamed, is it feasible to change in settings as well? this one should be the most graceful fix, but seems to be the most complex (i guess)
- If the target folder is deleted, same behaviour as deleted account (disable i.u. and warn the user to select a different target)

what do you think?

jesmrec:
### (4)

1. Install from scratch
2. Create account
3. Open an account
4. Go back to bookmark page and try to delete it. 

Current: account not deleted, when you try to open it, error "account locked" is shown and new list entry added each time you tap it.

Expected: Account deleted

jesmrec:
### (6)

1. Install the app from scratch
2. Add an account
3. Enable instant uploads for Videos, and keep instant upload for Images disabled
4. Select a correct account and folder to upload
5. Take an image and a video

Current: image is uploaded, video is not uploaded. In fact, videos are never uploaded, only images.
Expected: backwards. image not uploaded, video uploaded.

felix-schwarz:
@jesmrec Fixed the duplication of bookmarks part of (5) in https://github.com/owncloud/ios-app/pull/484/commits/e3f899eb38735ea4b5ede11ef0afce9b38a5fd44.

jesmrec:
Regarding (5), problem is reproducible when setting the account as i.u. account:

Aug-21-2019 16-32-29

jesmrec:
### (7)

1. Set up the i.u. feature correctly
2. Take a huge amount of pics (200, f. ex)
3. open the app -> pics are detected start to upload
4. Once everything is detected, minimize the app and let the device
5. After 5 minutes, open the app again

Current: Many pics seems to be stucked
Expected: everything uploaded in background.

ownCloud/Client/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,111 @@
...
+import Foundation
+import ownCloudSDK
+
+class BackgroundFetchUpdateTaskAction : ScheduledTaskAction, OCCoreDelegate {
+
+	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.background_update") }
+	override class var locations : [OCExtensionLocationIdentifier]? { return [.appBackgroundFetch] }
+	override class var features : [String : Any]? { return [ FeatureKeys.runOnWifi : true] }
+
+	var cores = [OCCore]()
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,111 @@
...
+		OCBookmarkManager.shared.loadBookmarks()
+
+		for bookmark in OCBookmarkManager.shared.bookmarks {
+
+			// Request cores for the bookmarks and add them to the list
+			let waitForCoreSemaphore = DispatchSemaphore(value: 0)
+			OCCoreManager.shared.requestCore(for: bookmark, setup:nil, completionHandler: { (core, _) in
+				if let core = core {
+					core.delegate = self
+					self.cores.append(core)
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,111 @@
...
+						Log.error("fetchUpdates() for \(core) returned with error \(error!)")
+					}
+
+					if self.cores.count == coresUpdated {
+						if successfulUpdates == coresUpdated {
+							self.result = .success(nil)
+						} else {
+							self.result = .failure(lastError)
+						}
+						self.completed()
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,111 @@
...
+			completed()
+		}
+	}
+
+	func core(_ core: OCCore, handleError error: Error?, issue: OCIssue?) {
+		if let error = error {
+			self.result = .failure(error)
+			Log.error("Error \(String(describing: error))")
+		}
+		completed()
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz I addressed this with latest changes but still if not all bookmarks could be updated, the task will return an error. However all cores now shall be updated.

ownCloud/Settings/MediaUploadSettingsSection.swift
mneuwert: @felix-schwarz fixed

ownCloud/Settings/MediaUploadSettingsSection.swift
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
mneuwert: @felix-schwarz sure, makes sense. Fixed.

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/ScheduledTaskExtension.swift
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,111 @@
...
+		super.run(background: background)
+
+		// Iterate through bookmarks
+		OCBookmarkManager.shared.loadBookmarks()
+
+		for bookmark in OCBookmarkManager.shared.bookmarks {
+
+			// Request cores for the bookmarks and add them to the list
+			let waitForCoreSemaphore = DispatchSemaphore(value: 0)
+			OCCoreManager.shared.requestCore(for: bookmark, setup:nil, completionHandler: { (core, _) in
```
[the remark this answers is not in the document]
felix-schwarz: While implementing the `feature/itempolicy` part of this, I also ran into `lockedBookmarks` and realized both this and `lockedBookmarks` really should be dealt with on the SDK level, so that the File Provider doesn't 
- try to use a bookmark that's in the process of getting erased
- re-download Available Offline files when it shouldn't

Working on that now.

mneuwert:
@jesmrec (3) + (5) + (6) shall be fixed with commit eb2a8d3ff9313273874966eab2f44d123ce0df45

ownCloud/Client/Tasks/ScheduledTaskManager.swift
mneuwert: @felix-schwarz fixed by moving the logic calling OS fetch completion block to the ScheduledTaskManager itself which is called when completion handlers of all background-fetch tasks have been called.

ownCloud/Client/Actions/Actions+Extensions/UploadFileAction.swift
```diff
@@ -63,10 +63,26 @@ extension UploadFileAction : UIDocumentPickerDelegate {
...
+					self.completed(with: NSError(ocError: .internal))
+					return
+				}
 			}
 		}
 
 		self.completed()
 	}
+
+	private func upload(itemURL: URL, to rootItem: OCItem, name: String) -> Bool {
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Tasks/InstantMediaUploadTaskExtension.swift
mneuwert: @felix-schwarz fixed

mneuwert:
@jesmrec Added some improvements for (4) in terms of bookmark / path selection for instant upload. Still handling for account / folder removed TBD though.

jesmrec:
### (7)

1. Install the app from scratch
2. Add an account
3. Enable i.u. (with account and path)
4. Remove the account of 2.
5. Add another account to the app
6. Open Settings

Current: Instant uploads can be enabled, but account and path are missed
Expected:  Instant uploads can be enabled, with account and path

mneuwert:
### Re (7)
@jesmrec in commit 9d20c1adc4454d86067790c82e7f596581c41016 which I just pushed I changed it, so that IU configuration is reset if the bookmark is not present any longer.

jesmrec:
regarding (8). It is fixed but in the case of a moved folder:

1. Install the app from scratch
2. Add an account
3. Enable i.u. (with account and path)
4. Move the upload folder to another account
5. Go to settings

Current: Account/Upload setting is missing
Expected: Following the approach of deletion and rename, the feature should be disabled.

On the other hand, if the target account or folder are removed, user should be warned that the feature will be off from such moment.

jesmrec:
At this point, i'm having several issues in the instant upload flow. They are reproducible but not in an easy way, i will try to detail as much as posible:

The list of files you will see in the gifs is the pre-set for instant uploads, with the path `/Instant Uploads`: 

1. Instant Uploads get sometimes stucked. `Instant Uploads` is the folder to upload. You take a pic and the pic is queued forever till the app is killed and reopened. Easier to reproduce from a scratch instalation. Will logs help? maybe this is not an instant uploads specific issue, but with manual uploads i did not reproduce in `milestone/1.1` branch.

ezgif com-resize

2. Deletion of instant upload folder is stucked. It is required to kill the app to finish the action. After reopening, the Instant upload settings are disabled, but the user is not warned

ezgif com-resize (1)

3. Moving target folder. After taking an image, nothing happens but the feature is disabled.

ezgif com-resize (2)

4. Removing an account. If the account for the instant uploads is removed, feature is disabled in Settings, but user is not warned. If an image is taken after the account deletion, nothing happens.

Aug-28-2019 12-09-38

Will give more details in case background issues are still there.

On the other hand, any clue about how to test this:

> First version of task scheduler which takes into account app states (backgrounded, foregrounded background fetch) and several events (WiFi available, low power, external power connected, photo library has changed etc). Those contribute to the execution context. Tasks are scheduled matching this context.

ownCloud/Settings/MediaUploadSettingsSection.swift
```diff
@@ -72,6 +170,194 @@ class MediaUploadSettingsSection: SettingsSection {
...
+		self.remove(rowWithIdentifier: MediaUploadSettingsSection.bookmarkAndPathSelectionRowIdentifier)
+
+		if let bookmark = getSelectedBookmark(), let path = self.userDefaults.instantUploadPath {
+
+			OCCoreManager.shared.requestCore(for: bookmark, setup: { (_, _) in },
+											 completionHandler: { (core, error) in
+												if let core = core, error == nil {
+													core.fetchUpdates(completionHandler: { (_, _) in
+														self.uploadPathTracking = core.trackItem(atPath: path, trackingHandler: { (_, pathItem, _) in
+															if pathItem != nil {
```
felix-schwarz: Please consider the possibility that the `trackingHandler` may be called several times. That doesn't seem to be the case right now. If you only want to handle the first value, the `isInitial` parameter passed to the `trackingHandler` can help with that.

ownCloud/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,203 @@
...
+						})
+					}
+
+					core.fetchUpdates(completionHandler: { (_, _) in
+						self.uploadDirectoryTracking = core.trackItem(atPath: path, trackingHandler: { (error, item, _) in
+
+							if error != nil {
+								Log.error("Error \(String(describing: error))")
+							}
+
```
felix-schwarz: Please consider the possibility that the `trackingHandler` may be called several times. That doesn't seem to be the case right now. If you only want to handle the first value, the `isInitial` parameter passed to the `trackingHandler` can help with that.

ownCloud/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,203 @@
...
+					assets.append(asset)
+				})
+			}
+		}
+
+		// Perform actual upload operation
+		if assets.count > 0 {
+			uploadGroup.enter()
+			self.upload(assets: assets, with: core, at: item, completion: { () in
+				uploadGroup.leave()
```
felix-schwarz: Not sure if a `DispatchGroup` is needed here. Instead, the `completion`-block here could `OnMainThread`-call this method's `completion`-block directly.

mneuwert: @felix-schwarz Indeed, fixed

ownCloud/Tasks/InstantMediaUploadTaskExtension.swift
mneuwert: @felix-schwarz fixed

ownCloud/Settings/MediaUploadSettingsSection.swift
mneuwert: @felix-schwarz fixed

ownCloud/PhotoKit Extensions/MediaUploadQueue.swift
```diff
@@ -0,0 +1,76 @@
...
+									assetUploadCompletion?(asset, false)
+								}
+								uploadGroup.leave()
+							}, progressHandler: { (progress) in
+								progressHandler?(progress)
+							})
+						}
+
+						// Avoid submitting to many jobs simultaneously to reduce memory pressure
+						_ = uploadGroup.wait(timeout: .now() + 1.5)
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz Removed the timeout

ownCloud/PhotoKit Extensions/MediaUploadQueue.swift
```diff
@@ -0,0 +1,76 @@
...
+class MediaUploadQueue {
+	private let uploadSerialQueue = DispatchQueue(label: "com.owncloud.upload.queue", target: DispatchQueue.global(qos: .background))
+
+	static let shared = MediaUploadQueue()
+
+	func uploadAssets(_ assets:[PHAsset], with core:OCCore, at rootItem:OCItem, progressHandler:((Progress) -> Void)? = nil, assetUploadCompletion:((_ asset:PHAsset?, _ finished:Bool) -> Void)? = nil ) {
+
+		let queue = DispatchQueue.global(qos: .userInitiated)
+
+		queue.async {
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/PhotoKit Extensions/MediaUploadQueue.swift
```diff
@@ -0,0 +1,76 @@
...
+import ownCloudSDK
+import Photos
+import MobileCoreServices
+
+class MediaUploadQueue {
+	private let uploadSerialQueue = DispatchQueue(label: "com.owncloud.upload.queue", target: DispatchQueue.global(qos: .background))
+
+	static let shared = MediaUploadQueue()
+
+	func uploadAssets(_ assets:[PHAsset], with core:OCCore, at rootItem:OCItem, progressHandler:((Progress) -> Void)? = nil, assetUploadCompletion:((_ asset:PHAsset?, _ finished:Bool) -> Void)? = nil ) {
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Actions/Actions+Extensions/UploadBaseAction.swift
```diff
@@ -40,36 +37,16 @@ class UploadBaseAction: Action {
...
-									completionHandler?(false, item)
-								} else {
-									Log.debug("Success uploading \(Log.mask(name)) to \(Log.mask(rootItem.path))")
-									completionHandler?(true, item)
-								}
-							}
-		) {
+	internal func upload(itemURL: URL, to rootItem: OCItem, name: String) -> Bool {
+
+		guard let core = self.core else { return false }
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Tools/URL+Extensions.swift
```diff
@@ -0,0 +1,41 @@
...
+//  Copyright © 2019 ownCloud GmbH. All rights reserved.
+//
+
+import Foundation
+import ownCloudSDK
+
+typealias UploadHandler = (OCItem?, Error?) -> Void
+
+extension URL {
+	func upload(with core:OCCore, at rootItem:OCItem, importByCopy:Bool = false, placeholderHandler:UploadHandler? = nil, completionHandler:UploadHandler? = nil) -> Progress? {
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,203 @@
...
+					core.delegate = self
+
+					func finalize() {
+						OCCoreManager.shared.returnCore(for: bookmark, completionHandler: {
+							self.completed()
+						})
+					}
+
+					core.fetchUpdates(completionHandler: { (_, _) in
+						self.uploadDirectoryTracking = core.trackItem(atPath: path, trackingHandler: { (error, item, _) in
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Tasks/InstantMediaUploadTaskExtension.swift
```diff
@@ -0,0 +1,203 @@
...
+
+		guard let bookmarkUUID = userDefaults.instantUploadBookmarkUUID else { return }
+
+		guard let path = userDefaults.instantUploadPath else { return }
+
+		if let bookmark = OCBookmarkManager.shared.bookmark(for: bookmarkUUID) {
+
+			OCCoreManager.shared.requestCore(for: bookmark, setup:nil, completionHandler: { (core, _) in
+				if let core = core {
+					core.delegate = self
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,93 @@
...
+
+		// Iterate through bookmarks
+		OCBookmarkManager.shared.loadBookmarks()
+
+		for bookmark in OCBookmarkManager.shared.bookmarks {
+
+			// Request cores for the bookmarks and add them to the list
+			coreUpdateGroup.enter()
+			OCCoreManager.shared.requestCore(for: bookmark, setup:nil, completionHandler: { (core, _) in
+				if let core = core {
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Tasks/BackgroundFetchUpdateTaskAction.swift
```diff
@@ -0,0 +1,93 @@
...
+					core.delegate = self
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

ownCloud/Client/Actions/Actions+Extensions/UploadMediaAction.swift
```diff
@@ -19,274 +19,60 @@
...
+		if let viewController = self.context.viewController {
+			let photoAlbumViewController = PhotoAlbumTableViewController()
+			photoAlbumViewController.selectionCallback = { (assets) in
 
-	private func exportVideoAsset(_ asset:AVAsset, targetURL:URL, type:AVFileType, completion:@escaping (_ success:Bool) -> Void) {
-		if asset.isExportable {
+				self.completed()
 
-			let preset = AVAssetExportPresetHighestQuality
+				guard let core = self.core else { return }
```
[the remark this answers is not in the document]
mneuwert: @felix-schwarz fixed

jesmrec:
After the latest commits `4771555` in the app and `747d90e` in the sdk, working fine with background included.

PR #489, opened 2019-08-29
[feature/darkmode] iOS 13 dark mode support

felix-schwarz:
## Description
iOS 13 dark mode support:
- detects dark / light mode changes and applies a different ThemeStyle as needed
- added ThemeStyle.darkStyleIdentifier so ThemeStyles can indicate their dark counter part (or their lack thereof, in case of the dark theme)
- add setting "Follow system light / dark appearance" to Settings > Theme

## Known UI issues in iOS 13
- the status bar background color is not always correct. Need to investigate if this is something we need to/can address or if it's an iOS 13 beta issue.

## Types of changes

____

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.2.0/DarkMode.md

Bugs & improvements:

- [x] (1) New structure n theming option https://github.com/owncloud/ios-app/pull/489#issuecomment- [FIXED]
- [X] (2) Color of multiselection and disclosures  https://github.com/owncloud/ios-app/pull/489#issuecomment-538292539 [FIXED]
- [X] (3) Theme after upgrading https://github.com/owncloud/ios-app/pull/489#issuecomment-538868796 [FIXED]


ownCloud/Client/Actions/MoreViewHeader.swift
```diff
@@ -25,17 +25,19 @@ class MoreViewHeader: UIView {
...
 
 	var showFavoriteButton: Bool
+	var adaptBackgroundColor : Bool
 
 	var item: OCItem
 	weak var core: OCCore?
 	var url: URL?
 
-	init(for item: OCItem, with core: OCCore, favorite: Bool = true) {
+	init(for item: OCItem, with core: OCCore, favorite: Bool = true, adaptBackgroundColor adaptBgCol: Bool = false) {
```
mneuwert: @felix-schwarz I think `_ adaptBackgroundColor: Boo`l would be nicer

ownCloud/Client/SortBar.swift
```diff
@@ -240,10 +245,16 @@ class SortBar: UIView, Themeable, UIPopoverPresentationControllerDelegate {
...
 	@objc private func presentSortButtonOptions(_ sender : UIButton) {
 		let tableViewController = SortMethodTableViewController()
-		tableViewController.modalPresentationStyle = UIModalPresentationStyle.popover
+		tableViewController.modalPresentationStyle = .popover
 		tableViewController.sortBarDelegate = self.delegate
 		tableViewController.sortBar = self
 
+		if #available(iOS 13, *) {
+			tableViewController.tableView.contentInsetAdjustmentBehavior = .never
+			tableViewController.tableView.contentInset = UIEdgeInsets(top: 13, left: 0, bottom: 0, right: 0)
```
mneuwert: @felix-schwarz move margin constant into separate definition?

ownCloud/Theming/ThemeCollection.swift
```diff
@@ -73,9 +73,29 @@ enum ThemeCollectionStyle : String, CaseIterable {
 	}
 }
 
+enum ThemeCollectionInterfaceStyle : String, CaseIterable {
+	case dark
+	case light
+	case unspecified
+
+	@available(iOS 12.0, *)
```
mneuwert: @felix-schwarz did you mean iOS 13.0?

ownCloud/Client/Actions/MoreViewHeader.swift
felix-schwarz: I agree `adaptBgCol` is ugly, but I'd like to keep the label. Addressed it by removing `adaptBgCol`.

ownCloud/Client/SortBar.swift
felix-schwarz: Added a local constant and added a comment to describe what this fixes.

ownCloud/Theming/ThemeCollection.swift
felix-schwarz: It's there mostly because of iOS 13, yeah, but since `UIUserInterfaceStyle` is available since iOS 12, that's what I use here.

ownCloud/Bookmarks/Issues/Issues Subclasses/ConnectionIssueViewController.swift
```diff
@@ -163,15 +163,13 @@ extension ConnectionIssueViewController: UITableViewDataSource {
...
-		cell.textLabel?.attributedText = NSAttributedString(string: issue.localizedTitle ?? "", attributes: [
-			.foregroundColor : color,
-			.font : UIFont.systemFont(ofSize: 18, weight: .semibold)
-			])
-
-		cell.detailTextLabel?.attributedText = NSAttributedString(string: issue.localizedDescription ?? "", attributes: [
-			.foregroundColor : UIColor(hex: 0x4F4F4F),
-			.font : UIFont.systemFont(ofSize: 15, weight: .regular)
-			])
+		cell.textLabel?.font = UIFont.systemFont(ofSize: 18, weight: .semibold)
```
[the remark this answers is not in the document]
felix-schwarz: In the scope of this PR, this is just a translation of the previously used attributed text to `UILabel` properties. I agree this should be addressed, but given this is far from the only place where fixed font sizes are used (Xcode returns 43 matches for `systemFont(ofSize:`), it should probably be dealt with in a separate PR that addresses this across the entire code base.

hosy:
@felix-schwarz does it make sense to hide the UI for selecting the theme type, if `Follow system light / dark appearance` is enabled? 
I would put this the switch `Follow system light / dark appearance` on the top of the view and if it is enabled, don't show the theme type selection, if it is disabled show it.

jesmrec:
Agreed. Also, in ' Settings, a value like `System`, so the following one is not consistent:

Screenshot 2019-10-04 at 10 05 33

You have the dark theme, but there is set as `Light`

jesmrec:
In which mode affects the light/dark mode of the device to the dark/light mode if the app.

Just an example:

Light mode in the app, the multiselection circles are different in the app depending the mode you select for the device. In case of device's dark mode, they are almost invisible. Other example is the disclosure indicators in the Settings view. Is this something we can control?

To illustrate:

|Light in device|Dark in device|
| :-: | :-: |
| Screenshot 2019-10-04 at 10 48 14 | Screenshot 2019-10-04 at 10 48 44 |

Same happens in the other way: Dark mode in the app, and light in device.

michaelstingl:
I like the theme switcher from [Outbank](https://apps.apple.com/app/id1094254051):



Du we really need our Light-Light theme? Make the Classic theme the Light theme?

hosy:
> In which mode affects the light/dark mode of the device to the dark/light mode if the app?
> 
> Just an example:
> 
> Light mode in the app, the multiselection circles are different in the app depending the mode you select for the device. In case of device's dark mode, they are almost invisible. Other example is the disclosure indicators in the Settings view. Is this something we can control?
> 
> To illustrate:
> 
> Light in device	Dark in device
> Screenshot 2019-10-04 at 10 48 14	Screenshot 2019-10-04 at 10 48 44
> Same happens in the other way: Dark mode in the app, and light in device.

@jesmrec multi-selection circles and disclosure indicator colors cannot be changed!

jesmrec:
### (3)

Install the app from scratch in iOS13

Current: default value for the theme is `System` 
Suggested: I'd set one of the app values (Classic/Light/Dark) as default, and the `Follow system` disabled. 

This is because the upgrading scenario: 

1. You have app 1.1.0 with dark theme. 
2. You have your iPhone in light theme 
3. Upgrade to 1.1.1

Result: `Follow system` is enabled, but the app is dark.

hosy:
> ### (3)
> Install the app from scratch in iOS13
> 
> Current: default value for the theme is `System`
> Suggested: I'd set one of the app values (Classic/Light/Dark) as default, and the `Follow system` disabled.
> 
> This is because the upgrading scenario:
> 
> 1. You have app 1.1.0 with dark theme.
> 2. You have your iPhone in light theme
> 3. Upgrade to 1.1.1
> 
> Result: `Follow system` is enabled, but the app is dark.

@jesmrec default value was fixed, please try again

felix-schwarz:
> ### (1)
> @felix-schwarz does it make sense to hide the UI for selecting the theme type, if `Follow system light / dark appearance` is enabled?
> I would put this the switch `Follow system light / dark appearance` on the top of the view and if it is enabled, don't show the theme type selection, if it is disabled show it.

No, theme selection should still be possible even when enabling `Follow system light / dark appearance`. Only that way can a user select which theme they want to use for the light appearance (`Classic`, `Light` - or anything else we might add in the future).

> Du we really need our Light-Light theme? Make the Classic theme the Light theme?

A matter of personal taste I guess, but I really don't like the Classic theme and prefer the Light one instead. I can imagine others might feel the same and appreciate the choice.

> You have the dark theme, but there is set as `Light`

Each Theme currently specifies if it is "light" or "dark", with light themes also providing their dark counterpart. So your selection between these themes:

System Regular | System Dark mode
-- | --
Light | Light
Classic | Classic
Dark | Dark

… becomes a pair when enabling `Follow system light / dark appearance`:

System Regular | System Dark mode
-- | --
Light | Dark
Classic | Dark
Dark | Dark

Putting `System` in the Settings line in question would not distinguish between `Light` and `Classic`. I'd suggest simply adding `(auto)` when automatic switching is enabled. Like `Light (auto)`, `Classic (auto)`, etc.

jesmrec:
Summing up with your thoughts @felix-schwarz 

This would be the testing matrix (tell me if i'm wrong)

System | App | Result
-- | -- | --
**Not following (same as before)**
Light | Light | Light
Light | Classic | Classic 
Light | Dark | Dark 
Dark | Light | Light
Dark | Classic | Classic
Dark | Dark | Dark
**Following (auto)**
Light | Light | Light
Light | Classic | Light - Classic
Light | Dark | Light 
Dark | Light | Dark with light variant
Dark | Classic | Dark with classic variant
Dark | Dark | Dark

felix-schwarz:
@jesmrec That's mostly correct. Only one thing is different: if someone selects the Dark theme, they'll always get the Dark theme. (fixed in your table and made it bold to highlight the change)

jesmrec:
Thanks a lot!!

hosy:
@felix-schwarz thank you for clarification, but I think there is a new problem:
Sample:
When the user selecting 
- dark theme, 
- "Follow System" is enabled, 
- System is in light mode, 
but no dark theme appears. 
Maybe we should only show the themes, which are available in the current system appearance mode, if "Follow System" is enabled.

hosy:
@felix-schwarz @jesmrec 
Summary of my idea:
If "Follow system mode" is enabled, I would only show the themes, which are relevant for the current system mode.

hosy:
@jesmrec did you check the new commit?

jesmrec:
Latest implementation is OK from my side in iOS13.

Approved.

PR #546, opened 2019-10-29
[fix/2ndLevelSort] Sort alphabetically if sort attribute is equal

felix-schwarz:
## Description
Fix alphabetically random order when sorting by anything but alphabetically: now items that share the same sort attribute value are subsequently also sorted alphabetically.

## Related Issue
App Store User Review

## Types of changes


jesmrec:
QA checklist:

- [X] By name: same behaviour as before
- [X] By type: files with same type are sorted by name
- [X] By size: files with same size are sorted by name
- [X] By date: files with same date are sorted by name
- [X] By shared: files shared/no shared are sorted by name

everything OK. Approved

PR #547, opened 2019-10-29
[fix/fp-offline-browsing] Allow offline browsing of folders in the File Provider

felix-schwarz:
## Description
Allows browsing of folders in the File Provider while in Airplane mode, with Files app no longer bringing up the "You're not connected to the Internet" error.

## Types of changes


jesmrec:
Now, content can be browsed in Files App even when connection is off.

Approved.

Issue #374, opened 2019-05-07
[Feature Request] Show cover/album art when playing music


mmattel:
This FR references issue https://github.com/owncloud/ios/issues/935
As you can see, there is already a PR for it, see https://github.com/owncloud/ios/pull/1105
Written by https://github.com/andrevz
@jesmrec 

I am wondering if that could be reused / implemented in the new IOS app...

jesmrec:
We will take it in account and check in which part of the roadmap fits better. Thanks for opening!!!

mmattel:
Hi @jesmrec, I would be happy if that would find its way into the roadmap, not seen so far in it...

jesmrec:
ok @mmattel, we will discuss about it 

CC @hosy @michaelstingl

Issue #59, opened 2018-06-03
[FR] Autoplay media files


michaelstingl:
Feedback from @michaelchadwick :

> I reeeaaallly want a way for the iOS client to be able to just play multiple mp3 files in a row.

(Source: [Tweet](https://twitter.com/mjchadwick/status/1002061007823368192))

I like the idea and thought about a "Autoplay" feature for all kinds of media file types:

![image](https://user-images.githubusercontent.com/214010/40887320-76c1bfc4-6747-11e8-91ed-6a91ba8a8331.png)

I used Figma.com for this quick sketch. Here you can fork it to propose own ideas:
https://www.figma.com/file/PgTb3ADoPgv4uQVuet0xYxAy

michaelchadwick:
@michaelstingl I like this idea!

So, if you listen to song1.mp3 in a folder, if "Include Audio Files" is selected, it will just automagically play song2.mp3 (downloading it first, if needed)? Will it loop back to song1.mp3 when song2.mp3 is done?

felixboehm:
Little research on this, as I find it annoying to start every mp3 in a golder manually.
From AVPlayer docs:

> AVPlayer is for playing a single media asset at a time. You can reuse the player instance to play additional media assets using its replaceCurrentItem(with:) method, but it manages the playback of only a single media asset at a time. The framework also provides a subclass of AVPlayer, called AVQueuePlayer, you use to create and manage the queuing of media assets played sequentially.

A queue would be nice, yet a simple solution could make use of

var actionAtItemEnd: AVPlayer.ActionAtItemEnd
The action to perform when the current player item has finished playing
https://developer.apple.com/documentation/avfoundation/avplayer/actionatitemend
can be set to advance, but I can‘t see where it takes the next item form

If the item end action can be handled it is possible to advance via
func replaceCurrentItem(with: AVPlayerItem?)
Replaces the current player item with a new player item

michaelstingl:
@mneuwert Could you provide a first estimation?

mneuwert:
@michaelstingl I implemented possible quick solution and pushed it on branch `feature/continue_media_playback`. I think it is quite simple and is fitting current preview architecture. Using `AVQueuePlayer` etc. would need some refactoring which I think would be needed anyway at some point for the preview sub-system.

PR #566, opened 2019-11-28
Autoplay Media / Show album artwork in the media player view

mneuwert:
## Description
- If there are further media files in the current folder they will enqueued for playing after current item has finished
- If media file contains artwork meta data it will be shown as overlay in the player

## Related Issue
#59 
#374 

## Motivation and Context
Improved UX concerning media playback

## How Has This Been Tested?
Tested with different mp3 files

## Types of changes

## Checklist:

## QA

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.2.0/Autoplay%20%26%20Artwork.md

### DIFF
--- a/external/libzip/libzip.xcodeproj/project.pbxproj
+++ b/external/libzip/libzip.xcodeproj/project.pbxproj
@@ -467,6 +467,7 @@
 			hasScannedForEncodings = 0;
 			knownRegions = (
 				en,
+				Base,
 			);
 			mainGroup = DCE93FD921FCA42B000E14F2;
 			productRefGroup = DCE93FE421FCA42C000E14F2 /* Products */;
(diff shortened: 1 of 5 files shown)


hosy:
@mneuwert it does not play the next item, if the app is in background. Furthermore do not get player controls in lock screen.

mneuwert:
@hosy strange thing is.. it works in simulator but not on the device. However since `UIKit` is involved, I think current implementation would be hard to get working in the background. The way to go would be to refactor the preview sub-system and allow `DisplayViewController` instances to handle multiple items and then also to use `AVQueuePlayer`. Regarding lock screen controls, I will continue to analyse what is going wrong

mneuwert:
@hosy Fixed the issue with background audio continuity and added even more features to the now playing view in the lock screen:
- Pause / play
- Skip 10 seconds backwards / forwards
- Show artwork, title, artist name

hosy:
@mneuwert great improvement. Thank you!

jesmrec:
Approved.

PR #574, opened 2019-12-10
[fix/fp-word] Fix Word FileProvider issue

felix-schwarz:
## Description
Fixes error and zero-sized documents when saving a document from Word.

## Related Issue
https://github.com/owncloud/enterprise/issues/3642

## Types of changes

### DIFF
--- a/ownCloud File Provider/FileProviderExtension.m
+++ b/ownCloud File Provider/FileProviderExtension.m
@@ -26,6 +26,9 @@
 #import "NSError+MessageResolution.h"
 
 @interface FileProviderExtension ()
+{
+	NSFileCoordinator *_fileCoordinator;
+}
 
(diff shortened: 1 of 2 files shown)


jesmrec:
- Creating the doc file from scratch and savinf in oC via FP -> OK
- Editing an existing doc from oC via FP -> OK
- Creating the doc from scratch with no device connection -> file is uploaded but appears with the spinner forever. File is really uploaded (you can see it in web UI), but in app seems to be in some intermediate status (you can not delete it, for example)

ezgif com-optimize (2)

I will add to the release list, feel free to fix here or in the branch

felix-schwarz:
@jesmrec Good find! Should be fixed with the latest commit.

jesmrec:
fixed. This can be merge into release branch. Will perform more tests there.

PR #579, opened 2019-12-19
[feature/awake] Keep device awake

felix-schwarz:
## Description
Keeps the device awake while the `Status` tab is selected and there are activities pending. Allows to let potentially long-running activities (like the initial scan of a large instance) complete.

## Motivation and Context
- User feedback like this: https://twitter.com/hector_gen/status/1207582465524740097

## Types of changes


jesmrec:
Building the current branch returns a fail:

Screenshot 2020-01-21 at 13 00 16

in files `updateAfterSingleFinishedUpload()` and `MediaUploadQueue`

Commit `9bf55ae`
Xcode 11.3  (11C29)

CC @felix-schwarz

jesmrec:
I do not get exactly what is really fixing this. Look at the following behaviour:

1. Added an account with a huge amount of files and recursive directories
2. In Status tab, it starts to sync

From here, i perform several actions:

- If let the device in foreground: it keeps syncing till it ends. It takes long since there are a lot of items to sync
- If i switch to another app and come back to oC several minutes later. It keeps syncing.
- If i switch to another account and come back to the first account. It starts to sync from scratch.
- If i kill the app and reopen: it starts to sync from scratch

The twitter report complains about the continuous syncing in which the app is, when the account is pretty populated. I do not notice differences with such tests.

All these behaviours are the same i noticed in `master`, so  i'm missing something. Please, clarify how to test this one.

hosy:
@jesmrec the device will prevented of sleeping and sync process can performed. But we cannot prevent that user put the app into the background. At the moment we don't have an other solution for this but it could help.
@felix-schwarz maybe its makes sense to show a notification to the user to keep the app in foreground until first sync is finished?

felix-schwarz:
@jesmrec It gives users an opportunity to leave the app open so that the initial discovery of server contents can complete. Without this feature to keep the device awake, the device will fall asleep - and the app will not be able to continue.

When putting the app into the background, logging out or killing the app, discovery shouldn't start from scratch. Instead, it should continue discovery where it left off. Since there's no way to retrieve a complete list of directories without visiting a directory, the list of directories to scan will continue to grow until the last directory has been discovered. Once that's the case, the number of remaining directories to scan will remain stable.

jesmrec:
Ok, after testing this:

- Device keep awake when Status tab is displayed and syncing -> OK
- Switching to other app and back to oC, sync is not reset -> OK

> killing the app, discovery shouldn't start from scratch. 

if i kill the app, the discovery starts from scratch:

Jan-24-2020 11-58-41

same happens if you switch to other account inside the app (is this one the expected behaviour in such case?)

jesmrec:
Discovering refers to the second number, so it does not stop after killing.

- Device keep awake when Status tab is displayed and syncing -> OK
- Switching to other app and back to oC, sync is not reset -> OK
- Killing the app, discovery shouldn't start from scratch -> OK

Approved on my side.

Issue #550, opened 2019-10-30
[RELEASE] 1.2


jesmrec:
### Git & Code

- [x] [GIT] Create branch `release/[major].[minor].[patch]` (freeze the code)
- [x] [DEV] Update `APP_SHORT_VERSION` `[major].[minor].[patch]` in [ownCloud.xcodeproj/project.pbxproj](https://github.com/owncloud/ios-app/blob/master/ownCloud.xcodeproj/project.pbxproj)
- [x] [TRFX] Update translations from transifex branch.
- [x] [DIS] Update [changelog](https://github.com/owncloud/ios-app/blob/master/CHANGELOG.md)
- [x] [QA] Design Test plan
- [x] [QA] Regression Test plan
- [ ] [DOC] Update user manual with the new functionalities
- [ ] [DOC] Update owncloud.org/download version numbers (notify #wordpress)
- [x] [GIT] Merge branch `release/[major].[minor].[patch]` in master
- [x] [GIT] Create tag and sign it `[major].[minor].[patch]`
- [ ] [GIT] Add the new release on [GitHub ios-app](https://github.com/owncloud/ios-app/releases)

If it is required to update the iOS-SDK version:

- [x] [GIT] Create branch library `release/[major].[minor].[patch]`(freeze the code) 
- [ ] [mail] inform #marketing about the new release.
- [ ] [DIS] Update README.md (version number, third party, supported versions of iOS, Xcode)
- [ ] [DIS] Update [changelog](https://github.com/owncloud/ios-sdk/blob/master/CHANGELOG.md)
- [x] [GIT] Merge branch `release/[major].[minor].[patch]` in `master`
- [x] [GIT] Create tag and sign it `[major].[minor].[patch]`
- [x] [GIT] Add the new release on [GitHub ios-sdk](https://github.com/owncloud/ios-sdk/releases)

If it is required to update third party:

- [x] [DIS] Update THIRD_PARTY.txt

## App Store

- [x] [DIS] App Store Connect: Create a new version following the `[major].[minor].[patch]`
- [x] [DIS] App Store Connect: Trigger Fastlane screenshots generation and upload
- [x] [DIS] Upload the binary to the App Store
- [x] [DIS] App Store Connect: Trigger release (manually)
- [x] [DIS] App Store Connect: Decide reset of iOS summary rating (Default: keep)
- [x] [DIS] App Store Connect: Update description if necessary (coordinated with #marketing)
[...]

PR #580, opened 2019-12-19
Fix for table view updates in the activity screen

mneuwert:
## Description
- Fixes jumping of progress view
- Updating only changed parts of the activity list instead of reloading complete tableview 

## Related Issue
#550 Issue Nr. 12

## Types of changes

## Checklist:


mneuwert:
@felix-schwarz you probably meant `ClientActivityCell.swift` with the first bullet point? 

As for the second point I also just found that code could be probably simplified by looking at the notification `userInfo` dictionary instead of trying to figure out what activities have been added / removed.

I didn't find any official information in the documentation that special care should be taken when chaining `performBatchUpdates` calls. Documentation only says that to ensure consistency model updates and `deleteRow` / `insertRow` calls shall happen inside the update closure.

Generally, I do agree that this change could have some impact and should be tested thoroughly before releasing it.

Issue #541, opened 2019-10-23
PDF Annotations in iOS 13.x


mneuwert:
Inspired by @hosy's implementation for his iOS13 _occon 2019_ talk.

Use `PKCanvasView` from `PencilKit` to annotate PDF document pages.

Implementation used for the demo can be found here bcbcb8cd88143253d47de8365577ad024b647a54

PR #607, opened 2020-01-23
[feature/quicklook-editing] Markup or Edit Files (on iOS 13)

hosy:
## Description
Implements a new action for PDF-, Image- and Video files to markup or edit these on iOS 13. Opens  item in a new QLPreviewController.

### PDFs and Images
It is possible to markup these file types on iPhone or on iPad with a pencil.

### Videos
Video files can be rotated or cropped

### Saving
User can decide if the original file should be updated or if a copy should created

## Related Issue
#541 

## Motivation and Context
In-App editing files

## How Has This Been Tested?
- tap on more button for the following file types: PDF, Image, Video
- tap on `Markup` (for PDF and Images) or `Crop and Rotate` for Videos

## Types of changes

## Checklist:

## QA

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.3.0/Markup.md

BUGS & IMPROVEMENTS:

- [x] (1) Options menu in landscape, overlaps the markup menu https://github.com/owncloud/ios-app/pull/607#issuecomment-579862811 [FIXED]
- [X] (2) `Save` wording https://github.com/owncloud/ios-app/pull/607#issuecomment-583329360 [FIXED]
- [X] (3) Save a new copy with no changes https://github.com/owncloud/ios-app/pull/607#issuecomment-583349313 [FIXED]
- [X] (4) Lacking some options https://github.com/owncloud/ios-app/pull/607#issuecomment-584071148 [WONT FIX]
- [x] (5) pictures do not fit the screen https://github.com/owncloud/ios-app/pull/607#issuecomment-584081152  [WONT FIX]


ownCloud/Resources/en.lproj/Localizable.strings
```diff
@@ -513,3 +513,11 @@
 "Previous" = "Previous";
 "Favorite" = "Favorite";
 "Cut" = "Cut";
+
+/* Markup */
+"How should this file be saved?" = "How should this file be saved?";
+"Overwrite originale" = "Overwrite originale";
```
mneuwert: **original**e

ownCloud/Client/Actions/Actions+Extensions/MarkupAction.swift
```diff
@@ -0,0 +1,112 @@
...
+
+				let appName = OCAppIdentity.shared.appName ?? "ownCloud"
+				let alertController = ThemedAlertController(with: "Cannot connect to ".localized + appName, message: appName + " couldn't download file(s)".localized, okLabel: "OK".localized, action: nil)
+
+				hostViewController?.present(alertController, animated: true)
+			} else {
+				guard let files = files, files.count > 0, let viewController = hostViewController else { return }
+				if let fileURL = files.first?.url, let item = self.context.items.first {
+					// Make sure self is around until interactionControllerDispatchGroup.leave() is called by the documentInteractionControllerDidDismissOptionsMenu delegate method implementation
+					self.interactionControllerDispatchGroup = DispatchGroup()
```
mneuwert: @hosy not quite convinced about the use of dispatch group here. Not really needed I guess..

ownCloud/Client/Actions/Actions+Extensions/MarkupAction.swift
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		return nil
+	}
+}
+
+@available(iOS 13.0, *)
+extension MarkupAction : EditDocumentViewControllerDelegate {
+	func editDocumentViewControllerDidDismiss(_ controller: EditDocumentViewController) {
+		interactionControllerDispatchGroup?.leave() // We're done! Trigger notify block and then release last reference to self.
```
mneuwert: @hosy you could just move lines 80-81 in here instead of using dispatch group notify, right?

ownCloud/Client/Actions/Actions+Extensions/MarkupAction.swift
```diff
@@ -0,0 +1,112 @@
...
+	override class var keyModifierFlags: UIKeyModifierFlags? { return [.command] }
+	override class var locations : [OCExtensionLocationIdentifier]? { return [.moreItem, .moreFolder, .keyboardShortcut] }
+
+	var interactionControllerDispatchGroup : DispatchGroup?
+	var interactionController : EditDocumentViewController?
+
+	// MARK: - Extension matching
+	override class func applicablePosition(forContext: ActionContext) -> ActionPosition {
+		let supportedMimeTypes = ["image", "pdf"]
+		if forContext.items.contains(where: {$0.type == .collection}) {
```
mneuwert: @hosy could be simplified I think:
- return .none by default at the end
- check if single item is passed AND has supported mimeType -> return .middle

ownCloud/Client/Actions/Actions+Extensions/MediaEditingAction.swift
```diff
@@ -0,0 +1,55 @@
...
+*
+* For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
+* You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+*
+*/
+
+import ownCloudSDK
+
+@available(iOS 13.0, *)
+class MediaEditingAction : MarkupAction {
```
mneuwert: @hosy Seems strange to me when `MediaEditingAction` is derived from `MarkupAction`. May be the base class should be something like DocumentEditingAction which btw. could have supportedMimeTypes property and could also have generic implementation of `applicablePosition(forContext:)` which is duplicated here in both actions

ownCloud/Client/Actions/EditDocumentViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+	func editDocumentViewControllerDidDismiss(_ controller: EditDocumentViewController)
+}
+
+@available(iOS 13.0, *)
+class EditDocumentViewController: QLPreviewController, Themeable {
+
+	weak var core: OCCore?
+	var item: OCItem
+	var editDelegte: EditDocumentViewControllerDelegate?
+	var handleSaving: QLPreviewItemEditingMode?
```
mneuwert: @hosy `handleSaving` sounds like boolean flag, may be just call it `editingMode` or `savingMode`?

ownCloud/Client/Actions/EditDocumentViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+		}))
+
+		self.present(alertController, animated: true, completion: nil)
+	}
+
+	func saveModifiedContents(at url: URL, handleSaving: QLPreviewItemEditingMode) {
+		switch handleSaving {
+		case .createCopy:
+			if let core = core, let parentItem = item.parentItem(from: core) {
+				self.core?.importFileNamed(item.name, at: parentItem, from: url, isSecurityScoped: false, options: [ .automaticConflictResolutionNameStyle : OCCoreDuplicateNameStyle.bracketed.rawValue ], placeholderCompletionHandler: nil, resultHandler: nil)
```
mneuwert: @hosy Are completion handlers of `OCCore` methods not used intentionally, because core would anyway create an issue which is going to be displayed to the user? May be at least adding some logging would make sense?

jesmrec:
### (1) (UI glitch)

Options menu in landscape of iPhoneX, overlaps. No way to see the last option

Screen Shot 2020-01-29 at 18 11 20

hosy:
@jesmrec unfortunately there is no way to dismiss the editing options. The only solution is, to use a shorter title.

hosy:
@jesmrec I can try `Saving file` as title

jesmrec:
Yes, or removing the `Cancel` option and cancelling by discarding changes.

hosy:
@jesmrec `Cancel` is removed and you can check it again

jesmrec:
i will ask for your opinion:

the option `Save` on the top left can be confusing for users. I mean:

1. Open a PDF file to markup
2. Markup 
3. You regret the changes and want to discard everything

Current:

You only see a `Save` button. So, you are afraid that tapping on `Save` will save the changes you want to discard (of course, you have never used the feature before and you don't know what is hidden behind the `Save` button). 

Should we change the word `Save` for other one? 
Should we set the `Discard changes` out of `Save` menu?
Should we let as is?

@michaelstingl @felix-schwarz @mneuwert

hosy:
@jesmrec I can change it back to `Done`. The problem is, I cannot see, if the document was changed. Only if the view will disappear, I get the callback. Maybe `Done` is a better wording and the user can select `Discard Changes` from here

jesmrec:
### (3)

1. Select `Markup` for a PDF
2. With no modification, select `Save` -> `Save  as a copy`

Current: Markup vire is dismissed and nothing happens

Expected: 

- Save copy of the file, even when it is identical to the original
- Warn the user 
- Let the user decide if wants a copy, or dismiss the changes.
- Let it be.

hosy:
@jesmrec (2) -> fixed with `Done` button 
(3) fixed -> copy of the file, even when it is identical to the original

jesmrec:
(2) and (3) fixed

jesmrec:
### (4) 

Some differences in view between iPad and iPhone:

| **iPhone** | **iPad**|
| :-:|:-:|
|Screen Shot 2020-02-10 at 12 02 20|Screen Shot 2020-02-10 at 12 02 03|

1.  Icons for Undo/Redo on the top bar are visible in iPhone, but in iPad

2. "Share with" button is on the top bar for iPad (accesible), and on the bottom on iPhone (overlapped with the markup bar). No place in the top bar but...

3. In iPad is visible a 3-dot-button on the bottom right with the option `Auto-minimise`, not in iPhone (i suspect this is something not-feature-related)

hosy:
@jesmrec (4) this options cannot be configured by us. We do not have an API for this and we can only use it, as it is. Maybe Apple will give as an option in a later iOS version.

jesmrec:
### (5)

Check the differences between the preview (tapping on a file to download) and the markup view (selecting markup option) for a picture:

| **Opening** | **Marking up** |
| :-: | :-: |
|Screen Shot 2020-02-10 at 12 31 29|Screen Shot 2020-02-10 at 12 31 36|

Should the markup view fit the whole screen as well as the preview?

jesmrec:
Approved

PR #609, opened 2020-01-31
[iOS] Open private link in app

mneuwert:
## Description
Support for opening private links using both approaches:
- Custom URL scheme (currently configured owncloud://..)
- Support for associated domains
- For now test domain is configured. This configuration shall be removed and later configured at build time if required.

## Related Issue
n.a.

## Motivation and Context
More convenient way to open shared private links rather than using Safari

## Types of changes

## Checklist:

## QA

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.4.0/Universal%20Link.md

Bugs & improvements:

- [x] (1) First time file is not previewed https://github.com/owncloud/ios-app/pull/609#issuecomment-613281511 [FIXED]
- [X] (2) Latest browsed path is opened, instead of the link https://github.com/owncloud/ios-app/pull/609#issuecomment-613328024 [FIXED]
- [X] (3) With the card opened, the links are not opened https://github.com/owncloud/ios-app/pull/609#issuecomment-613370509 [FIXED]
- [X] (4) Folder created externally is not discovered at time, when the link is clicked https://github.com/owncloud/ios-app/pull/609#issuecomment-613477172 [FIXED]
- [X] (5) With Passcode/Face ID links are not opened https://github.com/owncloud/ios-app/pull/609#issuecomment-613478421 [FIXED]
- [x] (6) Expected behaviour in case of no connectivity https://github.com/owncloud/ios-app/pull/609#issuecomment-613545073 [FIXED]
- [X] (7) Expected error handling https://github.com/owncloud/ios-app/pull/609#issuecomment-613868353 [FIXED]
- [x] (8) Links to hidden files https://github.com/owncloud/ios-app/pull/609#issuecomment-613868680
- [x] (9) After opening link, spinner is there forever https://github.com/owncloud/ios-app/pull/609#issuecomment-613875500 [WONT FIX]
- [X] (10) Late link opening https://github.com/owncloud/ios-app/pull/609#issuecomment-623454796 [FIXED]
- [X] (11) non friendly error message https://github.com/owncloud/ios-app/pull/609#issuecomment-623455811 [FIXED]
- [X] (12) Empty list after clicking on a non.discovered item https://github.com/owncloud/ios-app/pull/609#issuecomment-623455811 [FIXED]
- [X] (13) Incorrect error depending on the status of the device or the existence of the link https://github.com/owncloud/ios-app/pull/609#issuecomment-623455811 [FIXED]
- [X] (14) Incorrect error when opening an inexistent item in the correct account https://github.com/owncloud/ios-app/pull/609#issuecomment-627171762 [FIXED] 
- [X] (15) Some kind of files are not previewed https://github.com/owncloud/ios-app/pull/609#issuecomment-627332229 [FIXED]


ownCloud/Tools/URL+Extensions.swift
```diff
@@ -44,4 +44,53 @@ extension URL {
 
 		return progress
 	}
+    
+    func privateLinkItemID() -> String? {
+        
+        if self.pathComponents.count > 2 {
+            if self.pathComponents[self.pathComponents.count - 2] == "f" {
```
hosy: Please can you add a comment, why you searching for "f" or give an example.

mneuwert: @hosy done in last commit

ownCloud/Resources/Info.plist
```diff
@@ -35,6 +35,19 @@
...
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Viewer</string>
+			<key>CFBundleURLName</key>
+			<string>com.owncloud.com</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>owncloud</string>
```
hosy: Nice, that we have a url scheme to open a resource from outside. Is it also possible to open a file without the complete https url?
I would like to open a file like that:
using fileID: `owncloud://f/22` 
using localID: `owncloud://lid//76CC18...`

mneuwert: @hosy would be possible, but that's the question if we shall leave this feature in e.g. for AppStore customers? cc: @michaelstingl

michaelstingl: > we should build the app in a way both methods are supported. (magic `owncloud://` links as well as `https://cloud.domain.com/f/12345` in branded app)

(from https://github.com/owncloud/enterprise/issues/3551#issuecomment-537088097)

I hope we can get a server extension that converts private links to magic links when opened on mobile devices. This could boost the install base – that's why I would keep it free. (see https://github.com/owncloud/enterprise/issues/3660)

hosy: I think it would make our app more powerful to open files or folders from outside. You can open a file directly from Shortcuts.app or from other apps.

hosy:
@mneuwert I requested some small changes in CR and I tested your implementation and it works really great. Well done and I want to use it ❤️

hosy:
@jesmrec you can try to open an item like:
owncloud://demo.owncloud.com/f/22

ownCloud/Tools/URL+Extensions.swift
```diff
@@ -44,4 +44,70 @@ extension URL {
...
+				// E.g. if we would like to use app URL scheme (owncloud://) instead of universal link, to make it work with oC SDK, we need to change scheme back to the original bookmark URL scheme
+				if replaceScheme {
+					components?.scheme = bookmark.url?.scheme
+				}
+				
+				if let privateLinkURL = components?.url {
+					group.enter()
+					OCCoreManager.shared.requestCore(for: bookmark, setup: nil) { (core, error) in
+						if core != nil {
+							core?.retrieveItem(forPrivateLink: privateLinkURL, completionHandler: { (error, item) in
```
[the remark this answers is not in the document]
mneuwert: @hosy this comment made me think that actually slightly different change is needed. A check if `foundItem` is `nil` to make sure it is not overwritten if it is already valid. Implemented it in latest commit.

ownCloud/Tools/URL+Extensions.swift
```diff
@@ -44,4 +44,70 @@ extension URL {
 
 		return progress
 	}
+
```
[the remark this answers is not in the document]
mneuwert: @hosy fixed

ownCloud/SceneDelegate.swift
```diff
@@ -66,14 +72,39 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+                        }
+                    }, replaceScheme: true)
+                }
+            } else {
+                ImportFilesController(url: urlContext.url, copyBeforeUsing: urlContext.options.openInPlace).accountUI()
+            }
+        }
+    }
+    
+    func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
```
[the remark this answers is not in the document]
mneuwert: @hosy fixed

jesmrec:
### (1)

1. Open a link over a previewable file (image, pdf...) that was not ever opened before (not downloaded)

Current: oC app opens, file downloads but content is not displayed
Expected: after download the file, file is opened.

iPadAir v13

jesmrec:
### (2)

1. In your account, create a folder structure like `/A/B/C/D/E`... with several subfolders
2. Copy the link to some item in the root path
3. Browse through the structure defined in 1.
4. When you are in some of the folders (f. ex, in `D`), kill the app.
5. Open the link created in 2.

Current: `D` folder is opened, as the latest browsed one
Expected: File or folder pointed by the link is opened.

iPadAir v13
commit `f914f79`

jesmrec:
### (3)

1. In the app, open the card of any item in the list
2. Switch to another app (no killing)
3. Open a link to any item 

Current: Card is dismissed but file/folder is not opened
Expected: Card dismissed and folder opened:

In this gif, the link points to the `Photos` folder:

Apr-14-2020 12-51-11

iPadAir v13
commit `c38c88f`

jesmrec:
### (4)

1. In your oC app, switch to other account
2. In a different client, create a folder and get the private link
3. In your device, paste or write the link and open it (the app in your device has not discovered the new folder yet)

Current: Bookmark list is displayed. It seems that the item does not exist yet in the moment the app is opened, and it browses back to the list of bookmarks. 

Expected: Item is opened after being discovered.

jesmrec:
### (5)

(probably, related with (3))

1. In app, set passcode/face id
2. Open a link

Current: Passcode/face id is asked. Then, the file/folder is not opened
Expected: After unlocking, file/folder is opened

iPadAir v13
commit `c38c88f`

jesmrec:
### (6)

What is the expected behaviour in case of no connectivity? 

With the current implementation, nothing happens when the link is opened:

- App is open, with no further feedback
- If the file is already downloaded, it is not opened (it could)
- If the file is not downloaded, it is not opened in preview with the "download" button (it could)
- If it is a folder, it is not browsed (it could)
- If the file is not in root folder, it is not browsed to the path (it could)

this is my pov, since it is not specified the expected behaviour, i have thought in the implementation done in the legacy app.

I'm open to other ideas to scope the behaviour, just in case you think the current one should be improved.

jesmrec:
### (7)

Following the same line as (6), i have some questions regarding the error handling

If the linked item does not exist in the account anymore, the app is opened with no feedback. Same, if the account does not exist in the app.

Is it the expected behaviour?

jesmrec:
### (8)

another question: should links to hidden files be opened? i suppose yes, just for confirmation.

jesmrec:
### (9)

The following behaviour happens sometimes:

1. Open a link to a non-downloaded file

Current: file is downloaded and opened. When browsing back to the list of files, the file has the spinner forever:

Apr-15-2020 09-41-20

I have caught some logs:

[ownCloud_15_Apr_2020_at_09_40_42.log.txt](https://github.com/owncloud/ios-app/files/4479755/ownCloud_15_Apr_2020_at_09_40_42.log.txt)

iPhoneXR v13
commit `c38c88f`

mneuwert:
@jesmrec I did some some testing and couldn't reproduce most of the stuff (1,2,5) in iOS simulator (I tested with iPhone Xs and iPad Air simulators on iOS 12.4 and iOS13.4). Will continue on the device though. 

Regarding (4): Do you mean that you have accounts A and B and then have B as active account in the app and try open a link created in A? 

Regarding (6): I checked and it seems that we need internet connection in order to resolve the link to the actual path. Have seen that the legacy app uses oc:fileid property and tries to match it with the id in the link among cached items. I don't know if this is a safe approach and if we can add support for it in the SDK. I will check with @felix-schwarz on this topic. Otherwise we could present and alert with a warning in case there is no internet connectivity available (explanation that private links can't be resolved when offline).

jesmrec:
> @jesmrec I did some some testing and couldn't reproduce most of the stuff (1,2,5) in iOS simulator (I tested with iPhone Xs and iPad Air simulators on iOS 12.4 and iOS13.4). Will continue on the device though.

All the test i did were with real devices. If you do not finally reproduce, i will do another look. (2) was reproducible easily, and my feeling is that (5) is related.

> Regarding (4): Do you mean that you have accounts A and B and then have B as active account in the app and try open a link created in A?

i'm sorry, the first step in the description is not correct. I have updated the steps in (4). This is, having the app closed, create a folder in other client or web and then, open a link to such folder in the device.

mneuwert:
### Re: (4)
@jesmrec I checked and found following, in this case` -[OCCore retrieveItemForPrivateLink:completionHandler:]` returns valid item and later on, I am re-using existing code to actually navigate to this item which calls `ClientRootViewController.createFiliListStack(for itemLocalID:String) `which calls` -[OCCore retrieveItemFromDatabaseForLocalID:completionHandler:]` which doesn't find an item in the vault database. Not sure if this is correct, need to discuss with @felix-schwarz.

mneuwert:
### Re: (1), (2), (5)
Couldn't reproduce with iPad Pro running iOS 13.3.1. On iPhone I still have issues getting universal links to work at all with oC app. And as previously mentioned I tried simulators as well with the same result.

mneuwert:
### Re (3)
@jesmrec this should be fixed with commit 2c2f778329cd9e12042d14cce1a962294d97b745

mneuwert:
### Re (2)
@jesmrec I found some issues in how UI state restoration is handled and committed 2 fixes improving the behaviour. Also I verified that if the system launches the app correctly (passing the app `NSUserActivity` containing private link URL) it would always take precedence over restoring UI state (SceneDelegate.swift:40). So please re-test eventually with other device. Also please note that UI restoration (going back to last opened item) is only working on iOS13 in oC app since it was added as a part of multi-window implementation. And before iOS13, there was another quite cumbersome mechanism for that which we never supported.

mneuwert:
### Re (4) (6) (7)
@jesmrec Added an error alert shown in case it was not possible for the SDK to resolve a private link to an actual item. In (6) it might be due to lack of internet connection. (4) is under investigation by @felix-schwarz and he is also checking if anything can be improved about (6), e.g. still resolve to cached file even if no server connection possible.

mneuwert:
### Re (8) 
@jesmrec Tested and links to hidden file just work as if it would be any other file / folder

mneuwert:
### Re (9)

@jesmrec After some testing I noticed that this is not related to the private link feature, happens also when you just tap a file in the list.

jesmrec:
Checks with commit `48bedfa`

(1) -> Keep reproducing: Check with this steps, more accurate:

1. Install the app
2. Attach an account and get the link of a file (with no download)
3. Kill the app 
4. Paste the link in an editor (i used "Pages" )
5. Click the link

Current: App is opened. Linked file is neither downloaded nor opened.

If you do a second try, then it works
If you avoid the 3. step and only switch the app with no killing, it works

Let me know if this helps

(2) -> Keep reproducing (tried with two different devices). Did you kill the app after browsing and before clicking the link?

(3) -> fixed

(4) -> Pending to be checked

(will continue...)

mneuwert:
### Re (1) (2)
@jesmrec not sure, may be I missed the step with killing the app... will re-test with killing

mneuwert:
@jesmrec I finally was able to reproduce and fix (1). (2) is not reproducible for me, but it might be due to some fixes (including latest SDK fixes). And those latest changes should have fixed (4), (6) and (7). If (2) can't be reproduced on your side, I consider implementation as finished.

jesmrec:
with commits:

app: `e260de8`
sdk: `12a0d85`

(1) -> fixed

(2) -> keep reproducing, so it seems we are in a deadlock. Any information can help? (logs, any call...). I recorded a GIF. In the GIF, first i get the link of a folder called `FFFF` , then i browse deep through the folders, kill the app, and then, click the link. The latest browsed folder is opened instead of `FFFF`

As i commented above, same steps without killing app (only switching) makes everything work fine.

Apr-22-2020 15-02-32

Tested with two different devices (iPhoneXR and iPadAir)

(3), (4), (5) -> Working fine if the app is not killed. If the app is opened and the link is clicked, everything works. If the app is killed, the link is not opened. Probably, for the same reason as (2) (just wondering)

Caught a couple of GIFS (report 5). Both open the same file, but one with the app open and the other one, killed:

|Switching app| Killing app |
| :-: | :-: |
|Apr-29-2020 13-20-01| Apr-29-2020 13-23-00|

(6) -> Will check all the test cases in the 2nd round. All the checks i did over the fix worked for me.  

(7) -> I do not see the error message when a link to a non-existing item is open. Here you have a GIF. I got a link to an txt file. First, i am able to open the link. Then, i remove the file, and open the link. It should show the error, because the file does not exist anymore.

Apr-29-2020 13-41-34

I thought that the file (and its link) are alive in the trashbin, so i deleted from there with the same result. Any other idea, or something wrong that you notice there? 

(8) -> works for me, then.

(9) -> out of scope.

So, te main problems i'm "suffering" happens when opening links with the app closed... let's check for the best solution.

mneuwert:
@jesmrec I will re-test again trying to follow your test-procedures as close as possible...

mneuwert:
@jesmrec This stuff is difficult to test.. because I think there is lot going on behind the scenes in different parts of the app:

(2)
What I found is that on device if you tap a link to a file (tried with photos) and the app is killed, file Is not downloaded and only preview is shown. Will investigate... But the actual file pointed to by the link is open, so what you initially described: You leave app in folder /A/B/C/D/ and open file /A/file.jpg it is not opened... is not happening for me...

(4)(5)
Seem to work even if app is opened.

(7)
If you kill app, remove folder say in browser, then next time you open link associated with a folder, it would be opened with a "Folder removed" shown in the middle of the screen... Next time, you open it, you'll an error. This error is only shown if SDK doesn't provide valid OCItem for a link... If you try same with a link to a file, first time behaviour is similar to (2), second time error is shown.

jesmrec:
All reports i did were checking instances using the `owncloud://` scheme, not the apple-association-file. Here can be the difference.

jesmrec:
After the 1st round, i will proceed in the following way:

1. Performing a new round over the enterprise instance, since the first one was over a regular server instance.  Here, all the reports above will be checked and maybe new reports will come. 

2. A new round over the regular instance, after all bugfixing is done. All reports will be checked again. The ones above, and the ones after 1.

3. In case the number of reports at this point is big enough, a new round could be needed (i hope not)

in every report, i will add the kind of server it was detected.

jesmrec:
### (10) (improvement)

Server: enterprise instance
Device: iPadAir iOS13.4

After opening a link with the app closed, it takes some time till it is opened. When the file is reached, a message with `Network unavailable` is there during some seconds. Of course, device has connectivity:

May-04-2020 15-07-26
 
Is it concerned a slow server response? Also, after opening the app the list of servers is open and keeps in the screen for some time. An impatient user maybe will click before the process ends.

Any ideas?

jesmrec:
### (11) (improvement)

Server: enterprise instance
Device: iPadAir iOS13.4

Related with the screenshot [here](https://github.com/owncloud/ios-app/pull/609#issuecomment-621411327)

The message to the exception in server is not very friendly. Something to do with it?

Screenshot 2020-05-04 at 14 47 21

jesmrec:
### (12)

Server: enterprise instance
Device: iPadAir iOS13.4

This is not always reproducible but i saw many times:

1. kill the app
2. in web, upload a new file inside an existing folder and take the link
3. in the device, with the app closed, paste the link wherever and click it

Current: App is opened, but it appears the `Account` section instead of the list of files:

(only step 3 in the screenshot)
May-04-2020 15-16-31

mneuwert:
@jesmrec as for (10) I also noticed that sometimes it takes some time until SDK updates the connection status from offline to online.. I asked @felix-schwarz about his opinion on this.

Regarding (11), I think may be it is sufficient to show the text message and remove the error below it. The error in this case comes from the SDK which may be gets through passed from the oC server.

jesmrec:
### (13)

Server: enterprise instance
Device: iPadAir iOS13.4

1. With only an account (enterprise) attached in the device, click an `owncloud://` link

Current: App is opened, but nothing happens

Expected: not sure which one is feasible:
- Open in browser as before, so there is no accounts that matches such scheme
- Open the app and show error

any ideas about?

jesmrec:
Regarding (7) and (11)

Server: enterprise instance
Device: iPadAir iOS13.4

i see the message only if the app was killed and closed

1. Open the root folder
2. Switch to other app when you'll click the link (no kill)
3. Click a link to inexistent item

Current: Switched to oC, nothing happens
Expected: error message.

By killing the app in step 2., works correctly

jesmrec:
(1), (2), (3), (4), (5), (6), (8) and (9) work fine in the enterprise instances.

jesmrec:
> Regarding (11), I think may be it is sufficient to show the text message and remove the error below it. The error in this case comes from the SDK which may be gets through passed from the oC server.

for me it is ok. The text message is enough, and the exception does not give important information

mneuwert:
## Re (12)
I think the issue is here that the existing code (not related to private link) can't reconstruct a full path to the item (which is newly created) traversing it from the corresponding leaf since e.g. the item when retrieved Is missing `parentLocalID`. Need to investigate further...

felix-schwarz:
> ### (10) (improvement)
> Server: enterprise instance
> Device: iPadAir iOS13.4
> 
> After opening a link with the app closed, it takes some time till it is opened. When the file is reached, a message with `Network unavailable` is there during some seconds. Of course, device has connectivity:
> 
> Is it concerned a slow server response?

Yes. The viewer code in the app that presents the message in question is currently only using the `OCCore.connectionStatus` as indicator, which only switches from `offline` to `online` once the core has not only connectivity but is also fully set up and ready for use. So depending on current server and network performance, there can be a delay.

While the viewer code could additionally check `OCCore.connectionStatusSignals` to determine if the connection is really offline, I think a better way to handle this is to make this a high-level abstraction at the SDK level. By adding an additional `OCCoreConnectionStatus` - so `connecting` then joins `offline`, `unavailable` (for maintenance mode) and `online`. I'm working on this right now - and the change will be available in the `develop` branch soon.

felix-schwarz:
> @jesmrec as for (10) I also noticed that sometimes it takes some time until SDK updates the connection status from offline to online.. I asked @felix-schwarz about his opinion on this.
> 
> Regarding (11), I think may be it is sufficient to show the text message and remove the error below it. The error in this case comes from the SDK which may be gets through passed from the oC server.

Yeah, the message there is what was returned by the OC server in the XML response. That `(null)` shouldn't be there, though. I'll look into eliminating that.

Regarding the error message wording, I think it would be good if it differentiated between "couldn't resolve because you're offline" and "couldn't resolve because the server doesn't know the item" rather than leaving it to the user to find out which of the two applies.

mneuwert:
@felix-schwarz Wouldn't it be sufficient to call` OCCore.requestCore(...)` completion handler when the core is either ready to use (all connection signal handlers are in known state etc) or there is an error? To also avoid code like in `ClientRootViewController.swift:155`...

felix-schwarz:
@mneuwert Nope, unfortunately that wouldn't be a solution as connection signal providers don't indicate when they're in a "known state" and should be able to flip their status at any time.

Adding `.connecting` as new status on the other hand is relatively simple by means of adding another signal for this and make it dependant on `OCConnection.state`. Once the new `.connecting` status is available, code like in `ClientRootViewController.swift:155` will no longer be needed.

I've already implemented a new `.connecting` state… only a bit of testing left at this point.

felix-schwarz:
@mneuwert The `develop` SDK now has a new `.connecting` status. I also adapted `feature/auth-migration` to the changes. Please find the changes necessary for `ClientRootViewController` to take advantage of the new status here: https://github.com/owncloud/ios-app/pull/682/commits/b4c6e1681ced564060f038ef9bacc7d680980b67

`DisplayViewController` is not in a good shape, however. I debugged it to the point of seeing that it is receiving KVO-notifications for the `OCCore` changing its state, but it doesn't update its views accordingly.

I think rewriting `DisplayViewController` (and updating its look, like presenting a spinner while the state is `.connecting`)  will likely take less time than trying to find and fix its bugs, but that's just my $.02.

mneuwert:
@felix-schwarz Yes, agree.. `DisplayViewController` cries for rewrite.. I will check what can be done. Thanks..

mneuwert:
@jesmrec please check the latest state of this branch. I committed quite few changes and @felix-schwarz made a very important fix in the SDK. Also UX is improved:
- Progress HUD is shown while link resolution eventually involving server communication is ongoing. 
- Different error messages are shown if e.g. device is offline and item is not cached as opposed to item can't be resolved since it was removed from the server.

Please also not that I refactored the basis for the previews (DisplayViewController). I tested as much as I can but if you also keep an eye on file previews it would be great.

Also, I think there are few points to consider when testing:
- iOS 12.x vs iOS 13.x (especially on iPad) due to introduction of multi-window support. There are different delegates called on iOS13 when private link is invoked.
- iPhone vs. iPad with multiple windows on iOS 13

jesmrec:
Checking the last fixes by using the enterprise server instance with the app commit `f2958a8`:

(10) -> much more improved with a cool performance. I noticed a small glitch after opening the file, the `Connecting` label overlaps the progress bar. I think it is noticeable in the following animation:

May-08-2020 12-15-27

also, sometimes the file is not downloaded and you have to click the button again. But, this should not be a blocker at all.

(11) -> improvement done. Pay attention to (13) below with more related details about error message

(12) -> i did not notice the weird effect anymore. I will have it in my head during the next rounds, just in case. 

(13) -> Error message is displayed when a `owncloud://` link is opened onto an enterprise account (so, the expected behaviour :+1: ). 

But the displayed error message is (current behaviour)

Screenshot 2020-05-08 at 11 55 15

when the expected message is (expected behaviour)

Screenshot 2020-05-08 at 11 56 56

Regarding this, i noticed different error message if the link is opened in the bookmark list as in the file list. Check the following steps and screenshot.

1. Open the bookmark list
2. Switch to the app in which you will click the link (no kill).
3. Click on an inexistent link  -> app shows the bookmark list with the error message about lacking connection
4. Browse to file list
5. Switch to the app in which you will click the link (no kill). and click the same link as 2. -> correct message (link not resolved)

ezgif com-optimize (3)

Anyway, the global current status is pretty optimal with enterprise server. Looking forward to the next round with regular servers and `owncloud://` scheme.

mneuwert:
@jesmrec thanks for feedback, I will test and fix.. Also I am having trouble opening a link to a photo when app was terminated. Checking what's going on.. 

Regarding error messages: those reflect what comes back from SDK. Message is shown when SDK doesn't return an `OCItem` for a link.. and type of error message is depending on current state of `OCCore` connection status.

jesmrec:
> Regarding error messages: those reflect what comes back from SDK. Message is shown when SDK doesn't return an OCItem for a link.. and type of error message is depending on current state of OCCore connection status. So if it is not .online, then the message saying "You are offline etc." is picked.

so, it seems that offline in SDK means "i am not listing the files of any bookmark even if i have connection"

mneuwert:
@jesmrec I suspect that the issue is... when app is terminated, OCCore needs some time to be fully up and running and the retrieve link call triggers server request immediately.. and if no item was found and core is still not fully online, you get this error message. Just to understand it better: do you see error messages for files which exist on the server? Or just for deleted items?

mneuwert:
@jesmrec OK, for (13) I changed the way how proper error message is chosen.. Now the "offline" message is only shown if `OCCore` has determined it's connection status and if it is in fact OFFLINE.

mneuwert:
@jesmrec fixed a glitch you mentioned under (10) and also added small spinner shown under "Connecting..." label

mneuwert:
@jesmrec Fixed another two issues:

- Opening of owncloud://... links on iOS13 when the app is terminated (as a byproduct also fixed importing of files in the same situation)
- Not reliable opening of links prior to iOS13 when the app is terminated

jesmrec:
(10) glitch is fixed
(11) done
(12) i did not see such effect anymore. I will pay attention in the next round

Regarding (13), there are two different scenarios

a) Open an "enterprise" link over the correct account attached to the app, but the linked content does not exist anymore, or the device is offline (or both). All the cases (online + existing, online + non existing, offline + cached, offline + non-cached) displayed correct and accurate messages and tests passed. 👍 

b) Open an `owncloud://` link, but in the app the account does not exist. Regardless the status (online, offline, cached, non-cached, existing, non-existing), the error message always is:

Screenshot 2020-05-08 at 11 55 15

is the SDK checking the account, and as non-existing one, is it determining that it is offline? It is the same (f.ex), if you click on a link and no accounts are attached in the app. You will see the same message. Could it be improved?

mneuwert:
@jesmrec I will check 13 b)

mneuwert:
@jesmrec Pushed a fix for (13b)

jesmrec:
New message makes sense when the account does not exist. 13b fixed

jesmrec:
### (14)

after the fix in 13b:

1. Get the link of an item in the account and paste it wherever
2. Remove the item
3. Open the link

Current: `Link points to an account bookmark not configured in the app`
Expected: `Couldn't resolve the link because the item is not known to the server` 

commit `afb8622`
iPadAir iOS13.4
Server community (`owncloud://` scheme)

mneuwert:
@jesmrec (14) is fixed with the last commit

jesmrec:
(14) fixed

jesmrec:
### (15)

Regarding this:

> Please also not that I refactored the basis for the previews (DisplayViewController) which was quite a mess and especially since now it is also taking into account new connection states of the core, it was important to clean it up a bit. I tested as much as I can but if you also keep an eye on file previews it would be great.

There are several kind of files that are not previewed: txt and all MSOffice extensions (doc, docx, ppt, pptx...). 

It is not required to open private links, with a single tap is reproducible:

May-12-2020 15-06-11

commit `cc774d0`
iPadAir v13.4
iPhone XR v13.4
server community

mneuwert:
@jesmrec From the first look, doesn't seem to be related to private links...

jesmrec:
@mneuwert it does not have anything to do with links, but you suggested to take a look to file previews. The problem is not reproducible in current `master`, so it seems to be introduced in this PR

jesmrec:
(15) fixed

jesmrec:
@mneuwert 

about (15), now files are displayed, but only if they are not downloaded. If the file is already downloaded and you click the link, file is not opened.

You can check it by opening twice the same link: first time, file is downloaded and opened, second time, not posible.

mneuwert:
@jesmrec Please re-test with commit 4c2358104fd1611253cfe2f60395303104fd2a8b

jesmrec:
(15) is totally fixed with the latest commit.

Everything done, feature approved.

PR #616, opened 2020-02-18
[fix/offline-error-412] Fix endless error 412 loop for available offline downloads

felix-schwarz:
## Description
Addresses an issue where an outdated item in a download sync action could overwrite newer ones and lead to endless loops of 412 HTTP errors in combination with available offline.

## Steps to reproduce the fixed issue
- connect to server with iOS device
- turn on Airplane mode
- select an already scanned folder as "Available Offline"
- modify a file inside that folder with another client
- turn off Airplane mode

Previously: iOS app gets stuck in infinite loop of "error 412" issues.
Now: iOS app brings up no error, fixes the issue on its own.

Note: accounts previously stuck on this error will no longer bring up this error, but may retry downloading the file over and over until the database was updated with correct information either by:
- a change in the same directory triggering a rescan of the directory
- the user visiting the directory of the item in the app

## Types of changes


hosy (approved):

jesmrec:
Even when 412 are happening, are handled and fixed silently.

For individual files, the download spinner make retries (small UI glitches) that are fixed after refreshing, as specified in the first message. 

approved

jesmrec (approved):

hosy:
Checked the new code, problem with alert view is fixed, but there are still problems:

- Status view shows for the file a loop with pending and running state
- Never ending progress spinner in file list for the file
- I got an error with "Internal error (0)" after a restart of the app

hosy:
IMG_2041
IMG_2042

jesmrec:
> Status view shows for the file a loop with pending and running state

i reproduced this one, i think this is in the scope of the note above:

```
Note: accounts previously stuck on this error will no longer bring up this error, but may retry downloading the file over and over until the database was updated with correct information either by:

a change in the same directory triggering a rescan of the directory
the user visiting the directory of the item in the app
```

jesmrec:
Check that one: https://github.com/owncloud/ios-app/issues/623

it might be related.

felix-schwarz:
The latest commit should fix the remaining issues. Please give it a try. @hosy @jesmrec

hosy:
@felix-schwarz unfortunately it doesn't fix the problem.
As addition to this, the account does not connect to the server, until I manually log in again.

jesmrec:
I have tested the concerned reports

- Repetition of errors does not happen
- Loops in files does not happen

But, connection with the bookmark is lost and not recovered.

felix-schwarz:
@hosy @jesmrec Thanks for the feedback! Can you provide steps to reproduce the problems you're seeing?

felix-schwarz:
@jesmrec @hosy Please check again with the latest commit. If that doesn't fix the issue, please provide steps to reproduce the problems you're seeing, so I can debug them.

felix-schwarz:
The latest commit now takes many additional edge cases into account. To verify, these steps can be used (covering cases i. to vi.):

1) Place `file.jpg` into `folder`
2) Browse to `folder` in the app
3) Enable Airplane Mode
4) Perform one of the following with another app or in the web UI
    1) remove the file
    2) rename the file
    3) move the file out and into another directory
    4) modify the file
    5) modify the file and rename it
    6) modify the file and move it out to another directory
5) Disable Airplane Mode
6) Give the app time to resume the connection

@hosy @jesmrec

hosy:
@felix-schwarz For me the problem is solved! I got two times the error message again and now the status view is clean and I can open the file. The original file, which was moved, is no longer visible.

jesmrec:
Approved

felix-schwarz:
I also just fixed the merge conflicts.

Issue #622, opened 2020-02-19
[BUG] Shortcuts: Save File not working when already exists


hosy:
### Steps to reproduce
1. Download the following Shortcut: https://www.icloud.com/shortcuts/75704f3703124382962a42c0412e27dd
2. Run it minimum two times
3. Error and original file was not overwritten with the new content

### Expected behaviour
Save file gets new content

### Actual behaviour
Error and new file with conflict file name 
An unknown error occurred 

### Client
**iOS version:**
13.3.2

**ownCloud app version:**
1.3

hosy:
![IMG_0073](https://user-images.githubusercontent.com/736109/74870284-caedd380-5359-11ea-8623-9f4180652d62.jpeg)

jesmrec:
Already done! Shipped in 1.3.1

PR #624, opened 2020-02-21
[fix/shortcut-save-file]

hosy:
## Description
File could to be saved, when trying to save a file with Shortcuts and it already exists

## Related Issue
#622 

## How Has This Been Tested?
Check Issue #622 

## Types of changes

## Checklist:


jesmrec:
The fix directly overwrites the existing file in the same path. This behaviour can be dangerous and lead to data loss. Should it be better to create a new file with the incremental number (same as uploading twice the same file)?

hosy:
@jesmrec yes, this is exactly what it should do, because as I user I want to write data to the same file.
If you do not agree, we can add a parameter `Overwrite if already exists` with `YES/NO`, to prevent overwriting, if it should not. What do you think?

jesmrec:
It sounds good. So, the user can decide to overwrite the latest version or keep it by creating a new one.

hosy:
@jesmrec new parameter was added, please check again

jesmrec:
Works properly when  `Overwrite existing file` is enabled and disabled. And it is good to have that this is disabled by default, so the user has to decide explicitly if the file will be overwritten.

Approved.

Issue #618, opened 2020-02-19
[BUG] Signature in PDF files


jesmrec:
### Steps to reproduce
1. Open a PDF to markup
2. Click on  '+' and select Signature
3. Add a signature, wherever the place in the PDF
4. Save

### Actual behaviour
Open the file after saving, and the signature is not there

### Expected behaviour
Open the file after saving, and the signature is there

### Client
**iOS version:** 13.3

**ownCloud app version:** 1.3.0

**Device model:** iPhoneX

hosy:
Sounds like an iOS bug. What happens, if you sign the document and as an addition draw inside?

jesmrec:
If i only draw, strokes are correctly saved.
If i only sign, signature is not saved.
If i sign and draw together, nothing is saved.

hosy:
I just figured out how a signature will be saved:
- Add a signature to PDF
- Before you tap `Done`, end editing by tapping on the selected Markup button
- Tap on `Done` and save as copy or overwrite original 

For me this is an iOS bug, because the delegate does not return a new file, until the editing was ended. Unfortunately we cannot end editing by code.

jesmrec:
Ok, let's wait then for Apple's fix. We should not forget this in case of complaining users.

Issue #617, opened 2020-02-19
[BUG] Markup feature with HEIC files


jesmrec:
### Steps to reproduce
1. Open an HEIC image
2. Mark it up with signature, magnifier, shapes... 
3. Save it, no matter whether overwriting or saving in new file

### Actual behaviour
Changes are not saved. 

### Expected behaviour
Changes saved

### Client
**iOS version:** 13.3

**ownCloud app version:** 1.3.0

**Device model:** iPhoneX

hosy:
I just figured out how markup for HEIC files will be saved:

- Add a markup to a HEIC file
- Before you tap Done, end editing by tapping on the selected Markup button
- Tap on Done and save as copy or overwrite original
For me this is an iOS bug, because the delegate does not return a new file, until the editing was ended. Unfortunately we cannot end editing by code.

jesmrec:
Ok, let's wait then for Apple's fix. We should not forget this in case of complaining users.

PR #629, opened 2020-02-25
[fix/markup-saving] Fixes markup saving

hosy:
## Description
Fixes saving an edited markup file. When editing mode was not ended by the user, the delegate call with the changed file will be fired after the view was already dismissed.
Now will save this file, when it was not saved before.

## Related Issue
#617 
#618 

## Motivation and Context
Correct saving the edited file

## How Has This Been Tested?

- Markup a file
- Do not end editing mode
- Tap on `Done`
- Select a saving mode
- Changes are saved

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/Actions/EditDocumentViewController.swift
+++ b/ownCloud/Client/Actions/EditDocumentViewController.swift
@@ -28,6 +28,7 @@ class EditDocumentViewController: QLPreviewController, Themeable {
 	var savingMode: QLPreviewItemEditingMode?
 	var itemTracker: OCCoreItemTracking?
 	var modifiedContentsURL: URL?
+	var dismissedViewWithoutSaving: Bool = false
 
 	var source: URL {
 		didSet {
@@ -89,6 +90,8 @@ class EditDocumentViewController: QLPreviewController, Themeable {
 						self.saveModifiedContents(at: modifiedContentsURL, savingMode: savingMode)
 					} else if let savingMode = self.savingMode, savingMode == .createCopy {
 						self.saveModifiedContents(at: self.source, savingMode: savingMode)
+					} else {
+						self.dismissedViewWithoutSaving = true
 					}
 				}
 			}
@@ -98,6 +101,8 @@ class EditDocumentViewController: QLPreviewController, Themeable {
 					self.saveModifiedContents(at: modifiedContentsURL, savingMode: savingMode)
 				} else if let savingMode = self.savingMode, savingMode == .createCopy {
 					self.saveModifiedContents(at: self.source, savingMode: savingMode)
+				} else {
+					self.dismissedViewWithoutSaving = true
 				}
 			}
 		}
@@ -204,5 +209,8 @@ extension EditDocumentViewController: QLPreviewControllerDataSource, QLPreviewCo
 
     func previewController(_ controller: QLPreviewController, didSaveEditedCopyOf previewItem: QLPreviewItem, at modifiedContentsURL: URL) {
 		self.modifiedContentsURL = modifiedContentsURL
+		if self.dismissedViewWithoutSaving, let savingMode = self.savingMode {
+			self.saveModifiedContents(at: modifiedContentsURL, savingMode: savingMode)
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


jesmrec:
> **### How Has This Been Tested?**
> Markup a file
> Do not end editing mode
> Tap on Done
> Select a saving mode
> Changes are saved

does that means, that the expected behaviour is that the changes should not be saved if the user does not get out of markup mode?

hosy:
@jesmrec no, both options are possible. Should fix all problems.

ownCloud/Client/Actions/EditDocumentViewController.swift
```diff
@@ -28,6 +28,7 @@ class EditDocumentViewController: QLPreviewController, Themeable {
 	var savingMode: QLPreviewItemEditingMode?
 	var itemTracker: OCCoreItemTracking?
 	var modifiedContentsURL: URL?
+	var dismissedViewWithoutSaving: Bool = false
```
mneuwert: @hosy may be you could use more descriptive name? E.g. `editingModeLeftWithUnsavedContent`?

hosy: It does not reflect the name in my opinion, because it is important, that view was dismissed without unsaved content.

jesmrec:
With commit `3b9be76`

About  [#617](https://github.com/owncloud/ios-app/issues/617)

If after signing/adding strokes/adding shapes to the HEIC file the changes (signature, magnifier or shapes) are saved, but misplaced or missed (depending on where exactly did you set them). 

| Original HEIC | Markup mode | Saved file |
| :-: | :-: | :-: | 
|Screen Shot 2020-02-26 at 11 10 06|Screen Shot 2020-02-26 at 11 10 37|Screen Shot 2020-02-26 at 11 11 10


About [#618](https://github.com/owncloud/ios-app/issues/618)

If after signing/adding strokes/adding shapes to the PDF file, i see correctly the changes saved in iOS device. But, if you go to web UI and download the file stored in backend, the signature is not there. 

As this is a workaround of an iOS bug, maybe this is the best approach we can reach. Is there anything else that can be done? @hosy

hosy:
@jesmrec 
1. Misplaced shapes looks like an iOS bug. I will report it to Apple. Images are processed by the Apple framework.
2. Which PDF-Viewer did you use to show the PDF file with signature? Some clients cannot show the additional shapes and this will also processed by the Apple framework.

michaelstingl:
Some PDF viewers have an option to show/hide annotations. Please check if a shape gets baked in to the PDF itself, or if it's on the separate annotations layer that's only visible in some PDF viewers.

hosy:
@jesmrec regarding misplaced shapes in HEIC markup: It looks like this is the same behaviour as in Files.app, which means this is an iOS bug. I will file a bug and this should be added as new issue.

ownCloud/Client/Actions/EditDocumentViewController.swift
hosy: your suggested name does not reflect for what the variable is, because it is important, that the view was dismissed. What is with `dismissedViewWithUnsavedContent`?

mneuwert: @hosy then, I am also fine if you leave it as is...

jesmrec:
>  I will file a bug and this should be added as new issue

Thanks!!

HEIC files correctly prevented from Markup feature

> Some PDF viewers have an option to show/hide annotations. Please check if a shape gets baked in to the PDF itself, or if it's on the separate annotations layer that's only visible in some PDF viewers.

i used Preview, Adobe reader, and the preview of Books. But everything is OK. The point is sometimes Acrobat takes a little time to load the layers when the file is large.

Thanks for your feedback!

Approved

michaelstingl:
> HEIC files correctly prevented from Markup feature

In the future, we could convert HEIC to JPEG before users can annotate them. (Show quick warning to tell users what is going on)

hosy:
@michaelstingl yes, good point!

michaelstingl:
I think there are more unsupported image formats, like the *.NEF RAW file from my Nikon camera. Maybe useful for people in photography to annotate to communicate their ideas. In this case, it doesn't hurt to work on a copy.

jesmrec:
Good ideas!! maybe we can improve iterating in new issues.

Issue #623, opened 2020-02-20
[BUG] Uploading in a folder with no permissions causes a loop


jesmrec:
I reproduce the problem by using the regex control from server version 10.4

To set up:

open file `config/config.php`
add the rule:

`'blacklisted_files_regex' => [
  '.*\.jpg'
  ]`

so, all the files matching the rule will be blacklisted and not allowed to be synced.

### Steps to reproduce
1.  Set up (as above)
2. Upload content matching the rule (a jpg image or whatever)

### Current behaviour
UI gets in loop with the error (probably related with https://github.com/owncloud/ios-app/pull/616)

![Feb-20-2020 14-44-13](https://user-images.githubusercontent.com/14894746/74939389-1226a380-53f0-11ea-9722-9885ecde3028.gif)

even killing the app and reopening, the loop goes on

### Expected behaviour
Error message to notify and no more warnings.

### Client
**iOS version:** 13.3

**ownCloud app version:** 1.3.0

**Device model:** iPhoneX

**App logs**: 
[ownCloud_Feb_21__2020_at_8_34_13_AM.log.txt](https://github.com/owncloud/ios-app/files/4234814/ownCloud_Feb_21__2020_at_8_34_13_AM.log.txt)

### Server configuration

**ownCloud version:** 10.4RC3

Note: also reproducible 

- By sharing a folder with no permissions with another user, and the recipient trying to upload new content to the shared folder. 
- By uploading content to a external storage in "only view" mode (server >= 10.4)

michaelstingl:
@jesmrec could you post the full server response? (debug log or `mitmproxy`)

jesmrec:
> @jesmrec could you post the full server response? (debug log or mitmproxy)

sure. App logs added to the first message

michaelstingl:
This is what we get from the server:

```
# RESPONSE --------------------------------------------------------
Method:     PUT
URL:        http://10.40.40.198:28000/remote.php/dav/files/user1/Test/IMG_0025.jpg
Request-ID: 1208A71F-E843-4898-91AF-8D76B8755BDF
Error:      -
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
403 FORBIDDEN
Content-Type: application/xml; charset=utf-8
Keep-Alive: timeout=5, max=100
Pragma: no-cache
Content-Security-Policy: default-src 'none';
[...]

PR #632, opened 2020-02-26
[fix/permissions] Limit action availability based on permissions

felix-schwarz:
## Description
- limits action availability based on permissions
- provides a better error message for uploads returning with a 403 error

## Related Issue
#623 

## Types of changes

## QA

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%201.3.1/Forbidden%20error%20handling.md

BUGS & IMPROVEMENTS

- [x] (1) Duplicate operation with no creation permission https://github.com/owncloud/ios-app/pull/632#issuecomment-591868061 [FIXED]
- [x] (2) Instant uploads in banned folder https://github.com/owncloud/ios-app/pull/632#issuecomment-591883685 [FIXED]
- [x] (3) Regex ban https://github.com/owncloud/ios-app/pull/632#issuecomment-591888776 [FIXED]
- [X] (4) Create public link not posible https://github.com/owncloud/ios-app/pull/632#issuecomment-591932398 [FIXED]
- [x] (5) `Move` operation with `Change` permission https://github.com/owncloud/ios-app/pull/632#issuecomment-591935770 [FIXED]
- [x] (6) Markup not allowed with only `Create` permission https://github.com/owncloud/ios-app/pull/632#issuecomment-591939142 [FIXED]


mmattel:
Q: how do you determine a particular permission?
Is this based on core feedback ?
Because if so, what can be done if core replies no restriction but the backend as implemented one and you get post action a deny?

jesmrec:
Let's QA this. Test plan linked in the fist message.

To force `403'  errors:

- I will do private shares handling permissions
- External storage with "Read only"
- Regex expressions in `config.php`

jesmrec:
### (1)

1. Open a folder with no `Create` permissions
2. Open the card for any item

Current: `Duplicate` option is available. Duplication operation leads to correct error message.
Expected: `Duplicate` hidden (if feasible). 

iPhoneX 
iOS 13.3
commit `6259d60`

mmattel:
> External storage with "Read only"

If possible, differenciate the followingfor testing:
- since 10.4 you can set a mount point read ony (independent what the backend has)
- have the backend set to RO but keeping it RW (standard ) on core

jesmrec:
> If possible, differenciate the followingfor testing:

>since 10.4 you can set a mount point read ony (independent what the backend has)
>have the backend set to RO but keeping it RW (standard ) on core

yes... in fact, the issue that causes the current PR comes from tests performed over new features in 10.4 version like the new RO option in external storage. Thanks for the tip!

mmattel:
> But, in both cases, the client should receive the correct permissions in PROPFIND response.

The wish is correct... But if you compare the two scenarious, it might not be the case.

While on (1), the permission is controled by core, you get an immediate and correct response.
But on (2), you can get the real state only on a triggered action changing any content but not when just asking about the status what core thinks the mount has.

jesmrec:
### (2)

1. Enable instant uploads
2. Set as target folder, a folder with no permissions.
3. Take pics

Current: Error message is correct, but repeats in an infinite loop, no matter if you retry o cancel.

Expected: Several options:
- Prevent the loop in the error, as a bug fix.
- Prevent the user to select as target folder, a folder with no `Create` permissions.

First one sounds like a correct fix. Second one like a new extra feature. Any additional ideas? @mneuwert @michaelstingl @hosy 

iPhoneX
iOS 13.3
commit `6259d60`

jesmrec:
### (3)

will we finally keep the `Retry` option? 

For the same steps that i define in [#623](https://github.com/owncloud/ios-app/issues/623), the error message is improved but the loop is still there when you ban using a regex.

mneuwert:
### Re:(2)
@jesmrec second option is actually easy to implement, Just need to add filter block when instantiating directory picker view controller.

hosy:
@jesmrec (1) is fixed

jesmrec:
### (3)

1. Browse inside a folder with no `Create` permission
2. In any item inside, select `Links`
3. `Create Public Link`

Current: Nothing happens
Expected: Public link creation view is up. Only delete/edit permissions are allowed in case of a folder.  No changes in case of file.

iPhoneX
iOS 13.3
commit `548e84d6`

jesmrec:
### (5)

1.  Share a folder with other user with no `Change` permission
2. Recipient user open the card

Current: `Copy` option is available. Move is not.
Expected:  `Move` operation is also allowed. It is not in the card, but by dragging&dropping you can move files to other locations.

iPhoneX
iOS 13.3
commit `548e84d6`

jesmrec:
### (6)

We have a feature (not a bug) regarding the Markup.

Markup allows users to create new files (by saving a copy of the marked file), or change existing files (overwriting the original)

But, the option it is only displayed if the permission is `Change` (as well as Files App behaves). Users are not allowed to open a file, mark it up, and then, save in a new file (overwriting'd be forbidden in this case).

So, we have two options:

- Let it be as `548e84d` , so the users lose a feature (marking a file to save it as new copy) in a pretty specific case.

- Disable `Markup` only in case both `Create` and `Change` are not allowed. In case of only one  is allowed (f. ex. Create) and the user try to do the other one (f. ex Change),displaying only the available saving option in the `Done`, or directly show a friendly error message. 

iPhoneX
iOS 13.3
commit `548e84d6`

mneuwert:
@jesmrec Pushed a fix for (2) but still there is a room for improvement in case auto-upload directory permissions are changed after the upload folder was configured.

jesmrec:
> @jesmrec Pushed a fix for (2) but still there is a room for improvement in case auto-upload directory permissions are changed after the upload folder was configured.

should be the behaviour the same, as the directory is deleted? the change of permissions is not usual, and in this case will not lead to data loss (pics will remain in the camera roll, and can be uploaded manually). So, the error message can be enough as solution for a corner case. 

The point is, the error message is looping (before your commit). I will recheck and let you know.

hosy:
@jesmrec regarding (6)
I would choose the second option:
- Disable `Markup` only in case both `Create` and `Change` are not allowed
- If `Create ` is not allowed, disable option `Save as Copy`
- If `Change` is not allowed, disable option `Overwrite original`

hosy:
@jesmrec fixed (6) by the second option (see https://github.com/owncloud/ios-app/pull/632#issuecomment-591948404)

hosy:
@jesmrec regarding (5) I think is correct, but before the move or copy action is performed, it needs a check, if the destination directory have the correct permissions for this operation!?

jesmrec:
@mmattel 

I have checked the two differentiated cases you mention, and added to the test plan:

1) The case in which the new 10.4 option in mount point is enabled:

`Create` and `Change` operations can be prevented from client side. Such operations are handled in the same case as lacking permissions in sharing actions

2) The case in which the new 10.4 option in mount point is disabled (RW), but the content  in backend does not have permissions (RO):

Operation returns `403` and error is shown in an alert, since it can not be prevented from client side. 

Thanks for your tips! 

i will open a couple of related issues, that i realised during the 1st QA round.

felix-schwarz:
> Q: how do you determine a particular permission?

Permissions are determined based on what the server returned in response to the respective `PROPFIND`. The mapping from the server's sequence of permission letters to the permission bits is described here: https://github.com/owncloud/ios-sdk/blob/bcf820994140fbb8e89c2e921fd408e8351617bd/ownCloudSDK/Item/OCItem.h#L48 

> Because if so, what can be done if core replies no restriction but the backend has implemented one and you get post action a deny?

In that case, the app should handle the error correctly and present an error message.

Even if permissions were always correct, there'd still always be the scenario that permissions are changed on the server between the receipt of the `PROPFIND` response and the user starting an interaction.

In the end, presenting or not presenting actions based on the permissions returned by the server should really only be seen as a convenience to make the life of users easier, but not as a mechanism to prevent the possibility of errors happening/being shown.

felix-schwarz:
> ### (3)
> will we finally keep the `Retry` option?
> 
> For the same steps that i define in [#623](https://github.com/owncloud/ios-app/issues/623), the error message is improved but the loop is still there when you ban using a regex.
> 
> iPhoneX
> iOS 13.3
> commit `6259d60`

I'd keep it in case users have control over the server and decide to change the permission. The thing is that for uploads, this could be the only copy of a file that the user has, so users should at least have an option to fix the permissions and `Retry` rather than just `Cancel` and loose the file.

felix-schwarz:
> ### Re:(2)
> @jesmrec second option is actually easy to implement, Just need to add filter block when instantiating directory picker view controller.

To avoid errors when the user changes the permission lateron, https://github.com/owncloud/ios-app/blob/50f989e4c927b7d63019cbd223247e5bcdc4242f/ownCloud/PhotoKit%20Extensions/MediaUploadQueue.swift#L206 could also be a good place to check permissions.

felix-schwarz:
> @jesmrec (1) is fixed

Sorry to disagree, but that code just checks if a file could be created inside the folder that should be duplicated.

IMO that would block the duplication of a shared read-only folder (the entire folder) - which is actually possible and should be available.

The check that would need to be performed here would be to fetch the parent item of the item to duplicate and check if it has the required permissions. That's an async database operation, however, whereas the method that'd call it is run synchronously on the main thread.

The risk of a deadlock when waiting for the database seemed too high to me, though. Which is why I didn't add any permission check to that action.

However, thinking about it anew, I realized that typically, we also have access to the `OCQuery` via the `ActionContext` and can retrieve its `rootItem`, which should typically be the parent folder of the item to duplicate.

So with the latest commit, I changed the code to utilize that info where available.

felix-schwarz:
> ### (5)
> 1. Share a folder with other user with no `Change` permission
> 2. Recipient user open the card
> 
> Current: `Copy` option is available. Move is not.
> Expected: `Move` operation is also allowed. It is not in the card, but by dragging&dropping you can move files to other locations.
> 
> iPhoneX
> iOS 13.3
> commit `548e84d6`

I'm not sure I understand this one correctly.

If I share f.ex. `Photos` from `test` to `admin` without `Change` permission, `admin` will see a `Photos (2)` share:
- as `admin`, I can move the entire folder (-> action available in the app) 
- but I'm not able to move `Paris.jpg` from the `Photos (2)` to another folder, as I'm not allowed to change it (-> action not available in the app; also can't seem to do this in the web UI)

What doesn't work as it should and needs to be changed?

felix-schwarz:
> ### (4)
> 1. Browse inside a folder with no `Create` permission
> 2. In any item inside, select `Links`
> 3. `Create Public Link`
> 
> Current: Nothing happens
> Expected: Public link creation view is up. Only delete/edit permissions are allowed in case of a folder. No changes in case of file.
> 
> iPhoneX
> iOS 13.3
> commit `548e84d6`

How can I reproduce this? If I share a folder without `Create` permission, the only option available to the recipient for files in the shared folder currently is `Copy private link` - and in the web UI, sharing isn't possible, either.

jesmrec:
> > ### (4)
> > 
> > 1. Browse inside a folder with no `Create` permission
> > 2. In any item inside, select `Links`
> > 3. `Create Public Link`
> > 
> > Current: Nothing happens
> > Expected: Public link creation view is up. Only delete/edit permissions are allowed in case of a folder. No changes in case of file.
> > iPhoneX
> > iOS 13.3
> > commit `548e84d6`
> 
> How can I reproduce this? If I share a folder without `Create` permission, the only option available to the recipient for files in the shared folder currently is `Copy private link` - and in the web UI, sharing isn't possible, either.

1. user1 shares a folder with user2, with no `Create` permission

Screenshot 2020-02-28 at 11 56 39

2. user2 sees the folder in the list, then browses into it.
3. In any of the items inside the shared folder, user2 select `Links` 
4. Click on `Create Public Link`-> nothing happens

These are the steps. A GIF from `user2` 's  list of files. `1.SharedFolder` is the folder that `user1` shared with no `Create`

Feb-28-2020 12-00-52

sorry if the steps were no clear enough

felix-schwarz:
@jesmrec Thanks, (4) should be fixed now as of 8dd3355. It seems like it's not a new bug, though: previously sharing (via a public link) an item inside a folder that's shared with the user, was broken in general.

mmattel:
@felix-schwarz just a minor „issue“,
mind to sign the license/cla 😄

jesmrec:
(1) -> in UI, the `Duplicate` option is not displayed any more, but this comment is pending to solve: https://github.com/owncloud/ios-app/pull/632#issuecomment-592453732

(2) -> Actually, user is warned that the folder is not selectable because no permissions granted. But, as discussed in the call, a new mechanism is incoming. Waiting for in inside 1.3.1, or deploy the current solution?

(3) ->

> I'd keep it in case users have control over the server and decide to change the permission. The thing is that for uploads, this could be the only copy of a file that the user has, so users should at least have an option to fix the permissions and Retry rather than just Cancel and loose the file.

For me it's not a problem, even when it's not an usual case. Now it is working... only that the warning pops several times up before desisting (need more info here? logs, gifs...)

(4) -> fixed.

(5) -> 

> @jesmrec regarding (5) I think is correct, but before the move or copy action is performed, it needs a check, if the destination directory have the correct permissions for this operation!?

yes, when the target folder is selected, you check if the permissions are correct and then, perform the corresponding action. But, the point is,  moving **from** a protected folder with no `Change` permission. In this case, you are performing a delete operation (`Move` deletes the original). `Change` permission means changing the content of a file not the file itself, afaik.

> but I'm not able to move Paris.jpg from the Photos (2) to another folder, as I'm not allowed to change it (-> action not available in the app; also can't seem to do this in the web UI)

UI allows to move items placed inside a protected folder (no `Change` permission)  to another place. 

Let's illustrate:

With the following permissions:

Screenshot 2020-03-02 at 12 37 09

the following is the folder in the recipient, in which i'm able to move `ownCloud Manual.pdf` from the folder to the root folder:

Mar-02-2020 12-40-09

But in the app, the `Move` is not available in the card. But, by dragging&dropping i can move it to root:

Mar-02-2020 12-44-07

i hope it helps. Maybe confusing is the fact the web UI does not show explicitly the `Move` option, only drag&drop enabled. 

Let me know anything you need to move this forward.

felix-schwarz:
> (1) -> in app UI, the `Duplicate` option is not displayed any more, but this comment is pending to solve: [#632 (comment)](https://github.com/owncloud/ios-app/pull/632#issuecomment-592453732)

In my testing, `Duplicate` now works as it should. Is there any scenario left where it doesn't?

> (2) -> Actually, user is warned that the folder is not selectable because no permissions granted. But, as discussed in the call, a new mechanism is incoming. Waiting for in inside 1.3.1, or deploy the current solution?

I'd deploy the current solution (where the user is warned) and then adapt & improve as the new notification mechanism becomes available.

> (3) ->
> 
> > I'd keep it in case users have control over the server and decide to change the permission. The thing is that for uploads, this could be the only copy of a file that the user has, so users should at least have an option to fix the permissions and Retry rather than just Cancel and loose the file.
> 
> For me it's not a problem, even when it's not an usual case. Now it is working... only that the warning pops several times up before desisting (need more info here? logs, gifs...)

I can't reproduce this. What steps do I need to take to have the alert reappear even after pressing cancel?

> (5) ->

Should be fixed with the latest commit (https://github.com/owncloud/ios-app/pull/632/commits/7bd687cad59c2f7bf9dab0640f1394f3a99df22a).

jesmrec:
(1) -> 
>In my testing, Duplicate now works as it should. Is there any scenario left where it doesn't?

from my side as well. The point was that you disagreed with the initial fix. If it is OK now, it is fixed on my side.

(2) -> 

> I'd deploy the current solution (where the user is warned) and then adapt & improve as the new notification mechanism becomes available.

let us know when it is ready. If new stuff incoming will be appreciate to know with testing purposes. thanks!

(3) -> Steps to reproduce (reported in #623): 

1. Add `'blacklisted_files_regex' => [ '.*\.jpg' ]` to `config/config.php`
2. Upload content matching the rule (a jpg image in that case)

This regex expression support is included in oC 10.4, so it should be tested in such version. Here it is what i see using the regex above:

Mar-03-2020 10-13-57

(4) -> fixed

(5) -> fixed

(6) -> fixed

felix-schwarz:
@jesmrec Thanks for the detailed description on how to reproduce it. Found and fixed the issue as of ca802e339b19d927e76f3c51e6ff85e4cab6d806.

jesmrec:
(3) -> fixed

jesmrec:
After performing a quick 2nd round, everything is OK from my side. Approved

mmattel:
@felix-schwarz just for pleasing the license/cla bot. Mind to click on it...?

PR #640, opened 2020-03-05
Add extra check in release checklist, to avoid missing translations

jesmrec:
…in some languages


.github/ISSUE_TEMPLATE/release_template.md
```diff
@@ -18,6 +15,7 @@ Release a new version
 - [ ] [GIT] Create branch `release/[major].[minor].[patch]` (freeze the code)
 - [ ] [DEV] Update `APP_SHORT_VERSION` `[major].[minor].[patch]` in [ownCloud.xcodeproj/project.pbxproj](https://github.com/owncloud/ios-app/blob/master/ownCloud.xcodeproj/project.pbxproj)
 - [ ] [TRFX] Update translations from transifex branch.
+- [ ] [TRFX] Check for missing translations. Ping @cscherm
```
hosy: In my opinion this step is too late in the release template. As you can see, we have above the check to update the translations from Transifex.
In my opinion this task should be added to the PR template for every PR. The developer have to check, if every string was added to the `localizable.strings` and after this step the developer should ping all possible translators to check Transifex for missing strings -> DE: @hosy, @ChrisEdS RU: @mneuwert ES: @jesmrec

jesmrec: @hosy If we do that for every PR, translators in the team will have to translate X times during the same version (a bit bothering). I think it 'd be better to make all the translations together just before the release (you can decide in which moment). Otherwise, every PR with new strings will be blocked till everyone translates to his language. This is my opinion.

hosy: @jesmrec ok, you are right. What do you think about to add a checkbox to the PR template to ping translators, that new strings for translations are available (optional)? 
Maybe it would be cool, if Transifex could automatically notify a Rocket Chat channel that new translations are available.

jesmrec: works for me 👍

Issue #585, opened 2020-01-07
Ugly error caused after losing OAuth2 credentials


jesmrec:
This comes from 1.2 version: https://github.com/owncloud/ios-app/issues/550#issuecomment-565382903

Token removal/dismisal during uploads causes an ugly error:

1. Login with OAuth2
2. Start to upload a bunch of files (f.ex, an album with 50 videos)
3. While it is being uploaded, remove token (webUI -> Settings -> Security - Oauth2 -> remove iOS app as authorized)
4. Get a new token by signing in

Current: Ugly error when retries the pending uploads

![IMG_1086](https://user-images.githubusercontent.com/14894746/70792274-1f446800-1d99-11ea-9da1-b2fe74c563dd.PNG)

Dialog is not dismissed neither cancelling nor retrying. It raises again.

iPhoneX v13.2
iPAdPro v13.2

felix-schwarz:
I feel this edge case can't currently be adequately addressed:
- if an upload fails due to a non-recoverable error, the user needs to know
- retrying will retry the upload and run into the same error, letting the user know again
- cancelling will cancel and remove the upload

### Workaround:
The only way to escape this infinite loop (without cancelling the upload) is to 
- put the device in airplane mode
- log out of the account
- edit the bookmark
- turn off airplane mode and log back into the account

### Proposed Solution:
This should be addressed by the already planned move of errors to a list section in the `Status` tab, where users can interact with them on their own schedule - and without modally blocking the entire UI.

jesmrec:
Fixed in #662. To be released in 11.4

Issue #513, opened 2019-09-28
Better Alert / Notification Handling


hosy:
For connection issues, server issues, app issues the App will show the standard iOS alert view.
The problem is, if there are more issues at the same time, the user will see to much alerts and sometimes it is no longer possible to use the app.
e.g.
- oC Account has a size limit
- Photo Upload is enabled
- Size limit was reached
- There are more items in the upload queue
- Result: to much alerts

We should check, if there is a better notification system, which will not block the UI and can be easily dismissed, grouped or a detection, that the user already has seen the main issue.

Maybe we could show an overview of the notifications in the `Status` tab.

### Expected behaviour
App will not blocked by alert views, user can use the app.

### Actual behaviour
To much alert views will block the UI

### Client
**ownCloud app version:**
1.1.0

hosy:
## What the new notification system should handle:

- Show the same error or notice only one time to the user
- After the notification was shown, add it to the status view
- Give the user a possibility to interact with the message (Action, Remove Message)
- Add a button to clear all messages from the status view
- If the app is in foreground use a own notification class to show messages
- If the app is in background show important messages with local notifications

hosy:
This is a very flexible notification class for iOS. Check if we should use this:
https://github.com/Daltron/NotificationBanner

michaelstingl:
already released

Issue #505, opened 2019-09-24
Issue handling through notifications


felix-schwarz:
When issues come up during the use of the app, the app can present it to the user immediately.

If the app is working in the background or the File Provider is used, opportunities for presenting issues and asking the user to make a decision are limited or non-existent, often requiring compromises that have side effects like #362.

To improve this and provide a means to inform the user about issues and request the user's feedback at any time, a new mechanism is needed - and we have one available through iOS notifications.

Here's how it could work:
- if an issue comes up, it's first checked if it can currently be presented via a standard alert
- if it can't (f.ex. the app is in the background, issue comes from File Provider), it should be saved (f.ex. in the KVS) and then the user be notified via iOS notifications and prompted for action.
- in case the user disallowed notifications or missed them, the stored notifications should also always be available inside the iOS app - f.ex. at the top of the "Status" tab - or presented as soon as the user returns to the app.

felix-schwarz:
Implemented as of #662.

Issue #362, opened 2019-04-16
Offline folder creation in Files App


jesmrec:
it comes from release 1.0.0

In Files App:

1. Remove connection (plane mode)
2. Create a new folder

Current: After clicking on `Done`, view is not dismissed (folder is not created because the lack of connection)

Expected: View dismissed and file list displayed with the folder pending to be created (feasible?)

@felix-schwarz describes the problem in the following way:

https://github.com/owncloud/ios-app/issues/237#issuecomment-481896627

michaelstingl:
@felix-schwarz what needs to be done here?

jesmrec:
Fixed in https://github.com/owncloud/ios-app/pull/662.

PR #662, opened 2020-03-27
[feature/message-queue] Message Queue to allow flexible handling of issues and messages

felix-schwarz:
## Description
Introduces a new message queue mechanism that allows fine-grained routing, decoupled and tailored presentation, handling and automatic resolution of issues and messages. Works across the app and extensions like the File Provider.

## Related Issue
#362 #505 #513 #585 https://github.com/owncloud/ios-sdk/pull/63

### Dismissable card in-app
Issues from actions started from within the app pop up as cards. They can be dismissed by tapping in the area above the card to deal with them later.
Bildschirmfoto 2020-03-27 um 16 28 41

### Issue Handling via Status tab
Issues and messages the user didn't take action on yet are listed in the Status tab. Identical issues will eventually be grouped and offers the option to apply a choice (like "Retry" or "Cancel") to all of them at once. 
Bildschirmfoto 2020-03-27 um 16 27 09

### Notifications from File Provider
Errors and issues originating from actions started through the Files Provider / Files app are shown as notifications.
Bildschirmfoto 2020-03-27 um 16 26 33

### Responding to notifications
Issues can be dealt with directly in the notifications:
Bildschirmfoto 2020-03-31 um 00 49 22

Tapping notifications shows the message in the context of the originating account - or, if not possible - in the context of another account with an added header (pictured below):
Bildschirmfoto 2020-04-27 um 23 35 21

### Grouping of similar messages
If there are two or more similar messages, only the oldest one is presented directly, while options to apply the choice to all - or to show them all - are added below the issue:
Bildschirmfoto 2020-05-26 um 22 24 09

Turning on `Apply to all` adds an additional badge to clarify the choice will apply to more than one message:
Bildschirmfoto 2020-05-26 um 22 24 38

Tapping `Show all` shows all of the grouped messages and allows to respond to them individually:
Bildschirmfoto 2020-05-26 um 22 24 25

### Inline information
If there's a message related to an item or sync action, a ⚠️ warning sign appears next to it. Tapping it then brings up the message for the user to act on:
| In the file list | In the status tab |
| --- | --- |
| Bildschirmfoto 2020-06-17 um 22 57 28 | Bildschirmfoto 2020-06-17 um 22 57 37 |
| After tapping: | After tapping: |
| Bildschirmfoto 2020-06-17 um 22 57 34 | Bildschirmfoto 2020-06-17 um 22 57 40 |

### "All done" 
Displays an "All done" message when there are no messages or activities:
Bildschirmfoto 2020-06-25 um 15 53 48

### Per-account message count badge
In the bookmark list:
Bildschirmfoto 2020-03-31 um 00 49 47

### App Icon
Total message count badge on the app icon:
Bildschirmfoto 2020-03-31 um 00 51 25

## Types of changes

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.4.0/Message%20Queue.md

BUGS & IMPROVEMENTS:

- [X] (1) Target folder remove while uploading https://github.com/owncloud/ios-app/pull/662#issuecomment-652482467 [FIXED]
- [x] (2) Badge in app icon remains after reinstalling https://github.com/owncloud/ios-app/pull/662#issuecomment-652489908 [FIXED]
- [X] (3) Folder created in Files app is not removed even when was cancelled in the app https://github.com/owncloud/ios-app/pull/662#issuecomment-652534684 [MOVED TO https://github.com/owncloud/ios-app/issues/746]
- [x] (4) Notification points to other account https://github.com/owncloud/ios-app/pull/662#issuecomment-652976836
- [x] (5) https://github.com/owncloud/ios-app/issues/585 [FIXED]
- [X] (6) Keep both in quota oversize https://github.com/owncloud/ios-app/pull/662#issuecomment-654221845 [FIXED]
- [X] (7) Grouped errors https://github.com/owncloud/ios-app/pull/662#issuecomment-655989569 [DONE]


hosy:
@felix-schwarz thank you for your good ideas! I think this will solve a lot of problems.

What do you think to make the Dismissable card in-app a little bit more inconspicuous? More like a notification, which will automatically dismiss, without an user interaction? Because the present implementation will block the UI flow too.

It is really cool that same issues will be grouped!

felix-schwarz:
@hosy Thanks for the feedback.

I think that typically, you'd get an error as a direct consequence of an action - and you'd then likely want to get a chance to act on it immediately. From that angle, the card approach has a number of benefits over notifications:
- it's positioned at the bottom of the screen, which users can reach easily
- it can be dismissed (to be dealt with later in the Status tab) by tapping in the area above it
- it can't be missed when the user isn't constantly looking at the screen (thinking of up-/downloads that fail at some point)
- no extra taps (taps on "Status", find message, etc…) needed to act on the message

However, we have full flexibility there - and I can see how - for cases with many errors in a short time frame - having card after card appear is not a good UX.

So, what we could do is to have the first message show up as card - and then, for as long as the card is visible, show all other messages only as normal iOS notifications. That'd ensure users can act on a single message immediately, but they don't get overwhelmed if a lot of issues occur in short order. What do you think?

ownCloud/Messages/AlertView.swift
```diff
@@ -0,0 +1,217 @@
...
+			optionStackView.bottomAnchor.constraint(equalTo: enclosure.bottomAnchor, constant: -20),
+
+			titleLabel.leftAnchor.constraint(equalTo: enclosure.leftAnchor, constant: 20),
+			titleLabel.rightAnchor.constraint(equalTo: enclosure.rightAnchor, constant: -20),
+
+			descriptionLabel.leftAnchor.constraint(equalTo: enclosure.leftAnchor, constant: 20),
+			descriptionLabel.rightAnchor.constraint(equalTo: enclosure.rightAnchor, constant: -20),
+
+			optionStackView.leftAnchor.constraint(equalTo: enclosure.leftAnchor, constant: 20),
+			optionStackView.rightAnchor.constraint(equalTo: enclosure.rightAnchor, constant: -20)
```
hosy, line 188: Would be better to use constants for all these values and font sizes!?

ownCloud/Messages/MessageGroupCell.swift
```diff
@@ -0,0 +1,248 @@
...
+					badgeLabel = RoundedLabel(text: "", style: .token)
+					badgeLabel?.translatesAutoresizingMaskIntoConstraints = false
+				}
+
+				badgeLabel?.isHidden = true
+
+				if showAllLabel == nil {
+					showAllLabel = UIButton(type: .system)
+					showAllLabel?.translatesAutoresizingMaskIntoConstraints = false
+					showAllLabel?.setTitle("Show all", for: .normal)
```
hosy, line 155: `localize` is missing, be sure this string is available in localization file

ownCloud/Messages/MessageGroupCell.swift
```diff
@@ -0,0 +1,248 @@
...
+				if multiMessage, let applyAllContainer = applyAllContainer, let badgeLabel = badgeLabel {
+					containerView.addSubview(applyAllContainer)
+					containerView.addSubview(badgeLabel)
+				}
+
+				NSLayoutConstraint.activate([
+					containerView.leftAnchor.constraint(equalTo: self.leftAnchor, constant: 20),
+					containerView.rightAnchor.constraint(equalTo: self.rightAnchor, constant: -20),
+					containerView.topAnchor.constraint(equalTo: self.topAnchor, constant: 20),
+					containerView.bottomAnchor.constraint(equalTo: self.bottomAnchor, constant: noBottomSpacing ? 0 : -20),
```
hosy, line 193: Please use constants instead of the values in this view

ownCloud/Messages/MessageGroupCell.swift
```diff
@@ -0,0 +1,248 @@
...
+					])
+				}
+			}
+		}
+
+		self.accessoryType = .none
+	}
+
+	@objc func applyAllSwitchChanged() {
+		badgeLabel?.labelText = "+ \((messageGroup?.messages.count ?? 0) - 1) more"
```
hosy, line 222: `localize` is missing (please check strings for localization)

ownCloud/Messages/MessageQueueExample.swift
```diff
@@ -0,0 +1,87 @@
...
+ * For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
+ * You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
+ *
+ */
+
+import UIKit
+import ownCloudSDK
+
+/*
+		// Example of using this example:
```
hosy, line 23: Maybe better: `// Example of using the MessageQueueExample `

ownCloud/Resources/en.lproj/Localizable.strings
```diff
@@ -264,6 +264,8 @@
 
 "Forbidden Characters" = "Forbidden Characters";
 "File name cannot contain / or \\" = "File name cannot contain / or \\";
+"Item with same name already exists" = "Item with same name already exists";
+"An item with the same name already exists in this location." = "An item with the same name already exists in this location.";
```
hosy, line 268: This items are missing:
Apply choice to all similar issues
Show all

Please check if all strings was added

ownCloud/Server List/ServerListTableViewController.swift
```diff
@@ -102,6 +102,56 @@ class ServerListTableViewController: UITableViewController, Themeable {
...
+
+			OnMainThread {
+				self?.messageCountByBookmarkUUID = countByBookmarkUUID
+			}
+		})
+	}
+
+	var messageCountSelector : MessageSelector?
+
+	static let BookmarkMessageCountChanged = NSNotification.Name("boomark.message-count.changed")
```
hosy, line 129: Should you add a constant for this value: `"boomark.message-count.changed"`
Please check all other files, where this value was used.

ownCloud/Messages/AlertView.swift
felix-schwarz: Good point! Changed that.

ownCloud/Messages/MessageGroupCell.swift
felix-schwarz: Good catch! Fixed.

ownCloud/Messages/MessageGroupCell.swift
felix-schwarz: Changed the code accordingly. Thanks!

ownCloud/Messages/MessageGroupCell.swift
felix-schwarz: Thanks! Fixed.

ownCloud/Messages/MessageQueueExample.swift
felix-schwarz: Good idea. Changed that.

ownCloud/Resources/en.lproj/Localizable.strings
felix-schwarz: Thanks. Checked and "Apply to all" was also missing. Fixed all of them.

ownCloud/Server List/ServerListTableViewController.swift
felix-schwarz: Moved this to its own `NSNotification.Name` extension now. Thanks.

felix-schwarz:
@hosy Thanks for the review. Addressed all findings as of https://github.com/owncloud/ios-app/pull/662/commits/1fae60df8eddb26bd92c22388a23e8fad03f359e.

hosy:
@felix-schwarz please can you check if all new buttons and actions are accessible with keyboard shortcuts? If commands are missing these should be added that everything could be reached with the keyboard. Thanks!

jesmrec:
### (1)

this is a corner case

1. Select some big files to upload (so, you'll have time to do the following steps)
2. Using another client, remove the target folder once the uploads start

Current: notifications enter in an endless loop. Sometimes, in full screen:

Jul-01-2020 17-15-08

iPhoneXR, iOS13.4
commit `92e0103`

jesmrec:
### (2)

After executing (1), i had to remove the app from the device. After reinstalling, the badge in the app icon with the number of fails remain (status and list of bookmarks are clean). A fresh install should remove all the stuff

ezgif com-optimize (7)

If logs or whatever is needed here, let me know

iPhoneXR, iOS13.4
commit `92e0103`

jesmrec:
### (3)

(not totally sure this is related the current feature)

1. Open Files app
2. Create a folder with the same name as an existing file -> notification pops up
3. Click notification -> redirected to the app
4. Solve problem in the app (folder is cancelled)
5. Switch again to Files app

Current: Folder created en 2. and cancelled in 5. is not removed from Files App:

ezgif com-optimize (8)

Expected: Removed from Files App as well as in the app

iPhoneXR, iOS13.4
commit `92e0103`

felix-schwarz:
> @felix-schwarz please can you check if all new buttons and actions are accessible with keyboard shortcuts? If commands are missing these should be added that everything could be reached with the keyboard. Thanks!

For the "modal" alerts, I've added support for keyboard shortcuts with https://github.com/owncloud/ios-app/pull/662/commits/94233de1f4041acc086c4bd5b09551ab0890130e:

Bildschirmfoto 2020-07-02 um 12 04 55

jesmrec:
### (4) [IMPROVEMENT]


1. Attach two accounts to the app, and open one of them
2. Switch to Files App, and open the other account
3. Perform an action that causes an error (move a file to a folder with an existing name inside, create a folder with a name of a file...)
4. Notification pops up -> click on it

Current: Redirected to the app, but not to the account in which the error happened, but to the other one.
Expected: It'd be better to open the account in which the problem happened. 

iPhoneXR, iOS13.4
commit `92e0103`

jesmrec:
### (5)

about the issue attached in the first message: https://github.com/owncloud/ios-app/issues/585

Following the detailed steps: 

1. Login with OAuth2
2. Start to upload a bunch of files (f.ex, an album with 50 videos)
3. While it is being uploaded, remove token (webUI -> Settings -> Security - OAuth2 -> remove iOS app as authorized)
4. Get a new token by signing in

The error is still "ugly" for users, but presented in notification style:

Screen Shot 2020-07-06 at 14 47 55

On the other side, the only way to get the "Sing in" dialog is reopening the app. 

iPhoneXR, iOS13.4
commit `92e0103`

jesmrec:
### (6)

1. Upload a file in order to overpass the account quota

Current: This is the notification:



`Keep Both` option: not sure if this make sense...

Current: If you don't have enough space in your account, the only thing you can do is `Cancel`

iPhoneXR, iOS13.4
commit `92e0103`

felix-schwarz:
@jesmrec Going through the issues:

**(1)** I added an option to `Cancel` the upload with the latest commits.

**(2)** I think this is a bug in iOS. In the latest commit, I made a change so the badge count is set to 0 on start and then to the actual value - if that value is different. That should address the issue.

**(3)** This isn't related to the message queue, but an interesting find regardless.

**(4)** I had different thoughts when implementing this:
- if an error occurs in account A, but account B is the only one that's currently on screen (in split view for example), the tapped notification will bring up the message in account B, but with a title indicating the message belongs to account A.
- if an error occurs in account A, with account A and B open in the app, it'll try to show it in account A, but fall back to account B if it's the only one that's visible (again, with an additional title). Background here is that AFAIK iOS is allowed to close backgrounded "windows" at any time.
- leaving whatever other account a user is currently in and logging into the account the issue occurred is something I had implemented intermittently, but decided against, because it was a really confusing user experience and easy to loose a feeling of place for me as a user

**(5)** Thanks. I'll check this one tomorrow.

**(6)** `Keep both` here would add a number to the file to upload(`Photo.jpg`, `Photo (2).jpg`, …) so the existing and uploaded file can exist alongside. What should work different here?

felix-schwarz:
@jesmrec The latest commit addresses **(5)**. The error message now reads: `Couldn't upload … (error 3, oauth2: invalid_grant)`.

jesmrec:
(1) -> Makes much more sense to have the `Cancel` option. Then, other question comes to my mind: the number of problems to fix is limited to the number of maximum concurrent actions (uploads in that case). So, in case there are a large number of problems to solve (failed uploads), user will not be able to fix at the time. The option `Apply to all` means really, `Apply to all that are being processed/not queued`.  It might lead to confusion, and in case the number is really huge, it will really annoy. Is there something we can do to improve this? probably this is not in the scope of this PR, we can address in separate feature if there are other different stuff involved.

(2) -> fixed.

(3) -> i will open a separate issue

(4) ->

> if an error occurs in account A, but account B is the only one that's currently on screen (in split view for example), the tapped notification will bring up the message in account B, but with a title indicating the message belongs to account A.

i do not get this. Notification pops up in Files app, but the message does not indicate which account the problem happens:

Screen Shot 2020-07-07 at 10 47 17

do i miss something?

Other two points are OK for me.

(6) -> 

> (6) Keep both here would add a number to the file to upload(Photo.jpg, Photo (2).jpg, …) so the existing and uploaded file can exist alongside. What should work different here?

There is no other file with the same name in the account that collides with the uploading file. That error is displayed when the account quota is exhausted. In such case, you are not allowed to upload and the only posible action is `Cancel`.

jesmrec:
Regadring this: https://github.com/owncloud/ios-app/pull/662#issuecomment-652916362

you mean, the notifications that pops from the bottom of the screen. For the actions in the `Status` view, there are no keyboard shortcuts available, aren't there?

felix-schwarz:
> (1) -> Makes much more sense to have the `Cancel` option. Then, other question comes to my mind: the number of problems to fix is limited to the number of maximum concurrent actions (uploads in that case). So, in case there are a large number of problems to solve (failed uploads), user will not be able to fix at the time. The option `Apply to all` means really, `Apply to all that are being processed/not queued`. It might lead to confusion, and in case the number is really huge, it will really annoy. Is there something we can do to improve this? probably this is not in the scope of this PR, we can address in separate feature if there are other different stuff involved.

That's a good point. The Sync Engine places actions into "budgets", to make sure it doesn't take on too much work at once (this is separate from the placement of actions into "lanes", which makes sure actions are performed in a non-conflicting order).

Until the latest commit, an action waiting for user interaction reduced the available budget, at one point lowering it to zero, so other actions had to wait. With the latest commit, actions waiting for user interaction no longer use any budget, so that f.ex. queued uploads will all be tried and - if they fail - their error messages can be dismissed together.

> (3) -> i will open a separate issue

Thanks. I've spent quite some time trying to find out what's happening here, but couldn't find anything we could or should do different.

FWIW the File Provider and app correctly notify the Files app of changes, the Files app requests a new file list - and the file list is returned without the item - but the Files app still displays it. At least for the moment, I therefore believe this to be a bug in the Files app. But it's still good to track this with an issue.

> (4) ->
> 
> > if an error occurs in account A, but account B is the only one that's currently on screen (in split view for example), the tapped notification will bring up the message in account B, but with a title indicating the message belongs to account A.
> 
> i do not get this. Notification pops up in Files app, but the message does not indicate which account the problem happens:

Thanks & good catch! I was thinking only about the message coming up in the app itself.

With the latest commit, notifications now include a reference to the account, if the user has more than one account

> (6) ->
> 
> > (6) Keep both here would add a number to the file to upload(Photo.jpg, Photo (2).jpg, …) so the existing and uploaded file can exist alongside. What should work different here?
> 
> There is no other file with the same name in the account that collides with the uploading file. That error is displayed when the account quota is exhausted. In such case, you are not allowed to upload and the only posible action is `Cancel`.
> 
> keep QAing this.

Thanks for pointing this out. Internally, a wrong error code had been returned, it got the same treatment as "item already exists". Fixed that and now you should get "Cancel" and "Retry".

> Regadring this: [#662 (comment)](https://github.com/owncloud/ios-app/pull/662#issuecomment-652916362)
> 
> you mean, the notifications that pops from the bottom of the screen. For the actions in the `Status` view, there are no keyboard shortcuts available, aren't there?

Correct. For these, no shortcuts are available because there can be any number of issues in the status table, so it'd not be clear which one a shortcut is targeting.

jesmrec:
(1) -> fixed.

(3) -> https://github.com/owncloud/ios-app/issues/746.

(4) -> fixed.

(6) -> fixed. I saw a small glitch when operation is retried. Warning and progress bar overlaps:

Jul-08-2020 18-09-57

Starting the 2nd round.

jesmrec:
### (7)

The warning icon  Screenshot 2020-07-09 at 10 28 52  is not working on iPad.

in Phone: by clicking the warning icon, a card is displayed on the bottom to solve the problem individually
in iPad: nothing happens

iPhoneXR v13.4
iPadAir v13.4

felix-schwarz:
> (3) -> #746.

Thanks!

> (6) -> fixed. I saw a small glitch when operation is retried. Warning and progress bar overlap:
> 
> Jul-08-2020 18-09-57

I'm not sure what's happening here because the code that controls visibility of the progress ring and the alert button should mutually exclude that both are shown at once:

```swift
statusCircle.isHidden = hasMessage
messageButton.isHidden = !hasMessage
```

Is there any way I can reproduce this?

felix-schwarz:
Re **(7)**: it works almost like this:

iOS notifications don't support providing choices like `Retry` or `Cancel` directly with the notification. Instead, iOS wants apps to register "categories" and provide the choices with these. And then set the category's identifier for the iOS notification that should have these choices.

To support this, the SDK allows the registration of `OCMessageTemplate`s, from which this info can be derived, but which are also used to simplify the creation of error messages and avoid duplicate code.

Each of those `OCMessageTemplate`s has an identifier of its own, which is currently what is being used to group messages together:

Bildschirmfoto 2020-07-10 um 22 41 44

In the cases you described, the upload action uses the same template for uploads that failed due to a lack of space as well as those that failed because the upload targeted a non-existent directory … so these messages are grouped together.

If this should be a problem going forward, I can adapt the code to group each of these related but - at the root - different issues separately.

@jesmrec

jesmrec:
(6) is fixed
(7) clarified, and open to iterate. But the current solution is OK for me.

This is approved

Issue #43, opened 2018-05-11
Login: Switch Auth method in existing bookmark


jesmrec:
This is the scenario: 

1. User adds a basic auth bookmark
2. Server admin enables OAuth2
3. User go to "Edit bookmark"

At this point, the user (whose credentials have not changed) does not know anything about the the availability of a new and more secure authentication method. 

User have edited anything so, the `Save` and `Cancel` options keep the same status as before and the bookmark keeps the basic authentication instead of OAuth2. 

Discussion started here: 

https://github.com/owncloud/ios-app/pull/38#issuecomment-388342531

So, this is an open forum to discuss

jesmrec:
About @felix-schwarz thoughts:

> as a user I feel I should be able to make a change to f.ex. the name without being forced to upgrade the authentication mechanism.

absolutely agreed. One step over (out of the topic)... changing the Server Name and changing the credentials/auth method could be performed over different actions. So, the user would be able to change the server name itself, forgetting credentials or auth.

> I can see the clear security win for users to upgrade to a better mechanism if one is available, so the check could be made when hitting "Save", but only lead to a suggestion - and not be forced upon the user.

> Users who don't edit their bookmarks wouldn't benefit from this, however, so… maybe we should periodically check available authentication methods in general and tell the user if something better is available.

Using the `Save` button to notify that other options are available is a good option.  i think in another one, that is checking the URL at the time the `Edit` view is opened and notify in case OAuth2 is available (is it feasible?). Also, taking advantage of the  `Delete Authentication Data` option to point the user to input credentials again. 

In case we want to notify everywhere (not only in `Edit` mode), this would be an extra ball (option not available in current iOS app either). Periodically checks sounds to branding option (how often do you want i notify you that something new happened on the other side?)

This specific case will depend on how the core admin deals the change of auth method. In the worst case (core admin, who is a very lazy person, changes to OAuth2  and neither notify nor explain users how to proceed) 

IMHO, the usual scenario would be

- Server with basic auth
- Admin enables OAuth2 
- Admin notifies the change to all users by using internals (mail, phone, personally...), and points the users to Edit credentials, giving them the specific instructions to perform the change.

If we would assume this, the current issue could be closed. The point is... would we?

felix-schwarz:
IIRC this was eventually implemented.

PR #682, opened 2020-04-24
[feature/auth-migration] Authentication Method Migration

felix-schwarz:
## Description
When hitting a `401 Unauthorized` response, detects if an authentication method is no longer available and allows switching to another authentication method.

In addition, adds the ability to authenticate "inline" - that is, without leaving the account and bringing up the regular Edit Bookmark UI.

Also, authentication updates are now handled better (previously, users might have to log out and back into an account before it left `Offline` state).

## Related issue
https://github.com/owncloud/enterprise/issues/3864

## Scenarios for QA
Among other scenarios, these should be included in testing:
- [ ] create a basic auth bookmark, then, on the server turn on oauth2 and disable basic auth
- [ ] create a basic auth bookmark, then, on the server turn on oauth2 and disable basic auth, *but also change the (self-signed) server certificate (=> is the certificate change reported?)*
- [ ] create a basic auth bookmark, then, on the server turn on oauth2 and disable basic auth *while logged in*
- [ ] create an OAuth2 bookmark, then, on the server disable OAuth2
- [ ] create an OAuth2 bookmark, then, on the server disable OAuth2, *but also change the (self-signed) server certificate (=> is the certificate change reported?)*
- [ ] create an OAuth2 bookmark, then, on the server disable OAuth2 *while logged in*

### Invalid password / Password changed
Error message | Auth UI
-- | --
Bildschirmfoto 2020-04-26 um 15 07 08 | Bildschirmfoto 2020-04-26 um 15 07 32

### Invalid token
Error message | iOS prompt | Auth UI
-- | -- | --
Bildschirmfoto 2020-04-26 um 15 11 33 | Bildschirmfoto 2020-04-26 um 15 11 37 | Bildschirmfoto 2020-04-26 um 15 11 43

### OAuth2 -> Password downgrade
Error message | Auth UI
-- | --
Bildschirmfoto 2020-04-26 um 15 13 51 | Bildschirmfoto 2020-04-26 um 15 13 56

## Types of changes

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.4.0/Auth%20migration.md

- [x] (1) https://github.com/owncloud/ios-app/pull/682#issuecomment-641276873  
- [x] (2) https://github.com/owncloud/ios-app/pull/682#issuecomment-641350133
- [x] (3) https://github.com/owncloud/ios-app/pull/682#issuecomment-641842171


jesmrec:
As premises:

- Lacking `basic auth` will be a matter of oCIS, not oC10. At least for the moment

- `basic auth` can be excluded by using the env variable `oc:connection.connection-allowed-authentication-methods`. By default, such variable is not included, i will add it manually via XCode

jesmrec:
### (1)

Check the following scenario:

1. Login in a server with OAuth2
2. When the session active, remove OAuth2 from server
3. Refresh and `Sign In`

Current: Browser is opened, asking for credentials and authentication in web
Expected: OAuth2 is not supported anymore, so, it must ask for basic auth credentials. 

This is what i see after disablin OAuth2 in server:

ezgif com-optimize (6)

iPhoneX v13.4

jesmrec:
### (2)

About certifcates mentioned in the first message, i performed these steps:

1. Login in a OAuth2 (https with self signed)
2. Close the app
3. In server, disable OAuth2 and switch certificate
4. Reopen the app

At this point, new issue's certificated pops up, but after approving, it is asked again in an infinite loop. 

Same effect can be reproduced only by revoking certificate in `Settings` -> `Certificate` in a more straightforward way. In the `master` branch, this is not reproducible.

Jun-09-2020 15-15-19

iPhoneX v13.4

jesmrec:
### (3)

1. Login in a server with basic auth
2. In server, enable OAuth2
3. Restart the app

Nothing happens. Even when OAuth2 is enabled, basic auth is still available. Also, using the env variable `oc:connection.connection-allowed-authentication-methods` without `com.owncloud.basicauth`,  OAuth2 is not prompted. 

Is this the expected behaviour?

Not sure if this is a feature instead of a bug. The current PR comes from https://github.com/owncloud/enterprise/issues/3864, where it is intended to prevent the fact no `Basic` header is returned in a `404` response. In a oC10 server, `Basic` is always returned, either alone or together with `Bearer`. So, the current behaviour makes sense: app can not know whether OAuth2 is available till `404` happens... and with correct basic will never happen, regardless the local env variables.

felix-schwarz:
Re (1): I couldn't reproduce this. I did notice, though, that - testing with `demo.owncloud.com`, I needed to click `Uninstall` in the Market section a second time before the OAuth2 plugin would actually be uninstalled.

If you can reproduce it again, please provide a log so I can have a look at what the server indicates with regards to available auth methods.

jesmrec:
About (1), i keep reproducing the problem with commit `6ba16a14` and sdk `8bb142c`

These are the logs: 

[ownCloud_15_Jun_2020_at_12_57_45.log.txt](https://github.com/owncloud/ios-app/files/4779791/ownCloud_15_Jun_2020_at_12_57_45.log.txt)

(2) -> fixed

(3) -> Current implementation is OK for me. An interesting improvement (if feasible) is this one: https://github.com/owncloud/ios-app/issues/43.  If OAuth2 is enabled, relogin should point to OAuth2 instead of basic auth, because is a more prioritized auth method.

felix-schwarz:
@jesmrec Thanks for the log!

**(1)** should now be fixed as of https://github.com/owncloud/ios-app/pull/682/commits/b05419aa17a2b2c1b3c5efbbc47877839f768b4d. Previously, the use of a plain HTTP connection raised an issue and prevented the scan for available auth methods.

Now, if a permission to use a plain HTTP connection was given previously, it is respected when scanning for available auth methods.

jesmrec:
(1) is fixed

(3) -> migrating from basic to OAuth2 is only posible by re-attaching the bookmark, till https://github.com/owncloud/ios-app/issues/43 is implemented.

Issue #683, opened 2020-04-25
Swiping to next audio file doesn't stop playing previous file (2 files playing simultaneously)


yivo:
### Steps to reproduce
1. Open audio file (file A).
2. Wait until it starts playing.
3. Swipe to next audio file (file B).
4. Wait until it starts playing.

### Expected behaviour
File A should stop playing. File B should start playing.

### Actual behaviour
File A playing, file B playing (2 files playing simultaneously).

### Video

https://youtu.be/YyyVSmwpMQc

### Client
**iOS version:** 13.3

**ownCloud app version:** 1.3.2 build 158

**Device model:** iPhone 8 Plus

hosy:
Hi @yivo thank you for reporting this bug and for uploading a demo video. We will fix this soon! Thank you for using ownCloud!

jesmrec:
Fixed! will be available in the next release. Thanks for reporting!

PR #686, opened 2020-04-28
[fix/audio-stop-playing] Stop playing Audio when item loosing focus

hosy:
## Description
If an audio item is in detail view and playing audio and the user is swiping to the next item in the scroll view, the audio play does not stop. This is especially a problem, when the next item is also an audio file, because the user can here two audio files at once.

## Related Issue
#683 

## Motivation and Context
Correct handling of audio files

## How Has This Been Tested?
- place at min two audio files in a folder and open this files in detail view
- you can also try to place an image item beside this items to check correct behavior
- if the next item was not released by the system and the user swipes back, audio will resume playing on the last position

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/Viewer/DisplayViewController.swift
+++ b/ownCloud/Client/Viewer/DisplayViewController.swift
@@ -39,7 +39,7 @@ protocol DisplayViewEditingDelegate: class {
 	func save(item: OCItem, fileURL newVersion: URL)
 }
 
-class DisplayViewController: UIViewController, OCQueryDelegate, DisplayViewControllerProtocol {
+class DisplayViewController: UIViewController, OCQueryDelegate {
 	private let iconImageSize: CGSize = CGSize(width: 200.0, height: 200.0)
 	private let bottomMarginToYAxis: CGFloat = -60.0
 	private let verticalSpacing: CGFloat = 10.0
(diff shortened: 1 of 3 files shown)


ownCloud/Client/Viewer/DisplayViewController.swift
```diff
@@ -543,6 +543,12 @@ extension DisplayViewController {
 		self.core = configuration.core
 		self.state = configuration.state
 	}
+
+	func willLooseFocus() {
+	}
+
+	func willBecomeFocus() {
```
mneuwert, line 550: @hosy may be `willBecomeFocused()` would be just more grammatically correct? Also... is it not something already implemented by the responder chain stuff? All `UViewControllers` are actually subclasses of `UIResponder`.

hosy: You are right and now I´m using `viewDidAppear` and `viewDidDisappear` instead. Before I thought this will not work, when app is going to background. Thank you for that hint! Please give it a new try with the latest commit.

jesmrec:
Checks performed:

Folder with music files

- [X] Swipe forward no streaming
- [X] Swipe backwards no streaming (audio in the same point as before swiping)
- [X] Swipe forward streaming
- [X] Swipe backwards streaming

iPhoneXR v13.4

From my side is fixed

Issue #676, opened 2020-04-22
[FEATURE REQUEST] Support changing the format of uploaded image file names 


hamiller:
**Is your feature request related to a problem? Please describe.**
Please allow the definition of the file name for automatically uploaded (and from HEIFF converted) image file names.
Would be best if the EXIF information could be used for that.

**Describe the solution you'd like**
I'd like to have a text field in the configuration dialog to define the naming format of uploaded images based on their EXIF data.

**Describe alternatives you've considered**
Instead of a free input dialog, a checkbox could be used to decide if the new or the old file name format should be used:
Old:  "Photo-YEAR-MM-DD-HH-MM-S.jpg" (used by the legacy app)
New: "IMG_xxxx.jpg"

**Additional context**
I'd like to have all image file names in the style of "Photo-YEAR-MM-DD-HH-MM-S.jpg" - as it was in the legacy owncloud version.
That's because it's much easier to work with the upload folder on a file-name basis. Currently the upload folder contains files which where uploaded by the old legacy software as well as uploaded with this new one - causing bad sorting of all images.
Constantly renaming all images to the "old format" manually is not an option.

jesmrec:
This will be available in the next release.

Issue #641, opened 2020-03-07
[FEATURE REQUEST] Auto convert file attributes


cemizm:
When images are automatically converted from HEIC to JPEG, the current datetime is used for the newly created files. This leads to loosing the order of the images as well as the actual date and time the images were taken. 

The file attributes of the original image should be used for the newly created files in order to fix this issue.

michaelstingl:
@mneuwert FYI

felix-schwarz:
Thanks for the feature request!

There may still be potential to improve transfering metadata embedded inside the file.

However, changing the date of a file on the server is not supported by the server as far as I can tell.

mneuwert:
I verified and we do preserve EXIF metadata when converting and uploading the photos. Unfortunately at the time of conversion we are essentially creating a new file. And then when uploaded also new file is created is created on the server. As @felix-schwarz already mentioned there is no way to influence the file creation date on oC server.

However, there is another feature request #676 which suggest to stick to the legacy photo naming scheme and derive the name from the EXIF metadata. This can be done and would also allow for some meaningful sorting.

michaelstingl:
> However, changing the date of a file on the server is not supported by the server as far as I can tell.

I thought it's supported? `X-OC-MTime` is the http header which can be used in the `PUT` request. Value is an int based unix timestamp
https://github.com/owncloud/core/issues/33502

mneuwert:
@michaelstingl @felix-schwarz Ok... seen that SDK is using `OCItem.lastModifiedDate` internally to feed `X-OC-MTime` which is set to current timestamp if lastModifiedDate is nil. However I don't see that we can influence it on `-[OCCore importFileNamed:parentItem:...]` API level which is used in the app.

mneuwert:
this is going to be solved with PR 678

jesmrec:
Ready to next release

PR #687, opened 2020-04-28
Support changing the format of uploaded image file names

mneuwert:
## Description
Legacy App used to change names for uploaded photo assets and use naming scheme as follows: 
[Photo|Video]-timestamp-<xxxxx>.<ext>

Where:
- timestamp was a creation date of photo asset
- <xxxxx> an ID derived from original asset name (like IMG_xxxx.jpg) and ext - file extension.

Some users wanted to have these to at least be able to sort photos.. 

**In fact this PR corrects many other annoyances related to media asset uploads:**

- Performance of exports is improved by avoiding loading image assets into RAM whenever possible and also by picking correct export method for correct asset.
- Correct original image name (starting with IMG_) is now derived correctly by looking at all related asset resources and identifying primary ones.
- Nevertheless legacy naming scheme as requested in related issue is now default. Original names (IMG_xxxx) can be used if corresponding option in settings is activated ("Preserve original media asset names"). So names looking like UUIDs are gone.
- When uploaded asset is seen on the server with a correct creationDate (which is now passed from the photo library to the actual webDAV request). That implements the feature requested in #641 in fact.
- Edited photos are for now preferably uploaded over originals they are based on.
- Conversion of videos to MP4 is back with this PR.

## Related Issue
#676 
#641 

## Motivation and Context
In previous versions we relied very much on Photos.framework implementation and that it would provide a correct name but it turned out iOS photo library is not very consistent in how it handles its internal data structures and often you would end up with the name which looks like some sort of hex UUID.

So I've implemented naming as requested by #676 but also was able to make derivation of original name more reliable and added settings option for users who would rather like to keep original names as assets were named by the photo library.

Also when implementing I was able to refactor asset exporting code which improved the overall readability and also I could get some performance gains by using more efficient asset export mechanisms where appropriate.

## Types of changes

## Checklist:

_____

### QA

Test Plan:  https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.4.0/Asset%20renaming%20in%20Albums.md

- [x] (1) extension m4 is duplicated https://github.com/owncloud/ios-app/pull/687#issuecomment-640534688
- [x] (2) IMG_ prefix for videos https://github.com/owncloud/ios-app/pull/687#issuecomment-640536619
- [x] (3) Uploads from external app does not match with the setting about the file name https://github.com/owncloud/ios-app/pull/687#issuecomment-641102175


hamiller:
Hi there. Can we get an update here please?
Will this change make it into 1.4? 
Can I support? Maybe @jesmrec can have a look?

jesmrec:
This is on my queue to test. It will be included in the next release (11.4).

jesmrec:
### (1)

1. Enable "convert videos to MP4"
2. Take a video and upload, no matter if manually or with auto uploads

Current: Extension is .mp4.mp4 (duplicated)
Expected: .mp4

iPhone XR v13.4

jesmrec:
### (2)

This is not a matter of the current PR. With the original media file names, both videos and images were exported with `IMG_` prefix. Was that a problem of the library? just to be sure this is something we can not fix

jesmrec:
### (3)

Check this scenario out:

1. In app, disable `Preserve original media file names`
2. Go to Photos App
3. Select any Photo
4. Select Open In -> Save to Files -> ownCloud (any location inside the account)

Current: Photo uploaded with format `IMG_xxxx`
Expected: Should it be in format `Photo_<timestamp>_<sequence>`? because, the setting is disabled. Or, this is not in the scope f the current PR

iPadAir2 v13.4

mneuwert:
### Re (3)
This is handled by entirely different code path which is supposed to import any type of file and is not performing any special handing for media files. So, is not in scope.

mneuwert:
### Re (1)
This is fixed with the latest commit

jesmrec:
(1) fixed

(2) and (3) clarified, nothing to fix there.

Approved.

Issue #654, opened 2020-03-24
[BUG] Pin Code Buttons lost round shape


hosy:
## If you found a bug:

With issue #222 we changed the squared pin code buttons to round buttons. I just noticed, that we lost this change and the buttons are square again.

### Steps to reproduce
1. Set a pin code
2. Restart the app

### Expected behaviour
Pin code buttons are round

### Actual behaviour
Pin code buttons are square

### Client
**iOS version:**
13.3

**ownCloud app version:**
1.3.1

![Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-24 at 09 03 01](https://user-images.githubusercontent.com/736109/77402055-5372f000-6dae-11ea-84e9-3f694949ffc8.png)

PR #689, opened 2020-04-29
[fix/round-shape-button] Fix round shape for pin code buttons

hosy:
## Description
Round shape for pin code buttons was lost after changes in superclass. Adopted the code to this changes.

## Related Issue
#654

## How Has This Been Tested?
- set a pin code

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Theming/UI/ThemeRoundedButton.swift
+++ b/ownCloud/Theming/UI/ThemeRoundedButton.swift
@@ -23,13 +23,13 @@ class ThemeRoundedButton: ThemeButton {
 	override init(frame: CGRect) {
 		super.init(frame: frame)
 
-		self.buttonCornerRadius = -1
+		self.buttonCornerRadius = .round
 	}
 
 	required init?(coder aDecoder: NSCoder) {
 		super.init(coder: aDecoder)
 
-		self.buttonCornerRadius = -1
+		self.buttonCornerRadius = .round
 	}
 
 }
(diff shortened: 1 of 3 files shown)


ownCloud/Theming/UI/ThemeRoundedButton.swift
```diff
@@ -22,20 +22,14 @@ class ThemeRoundedButton: ThemeButton {
 
 	override init(frame: CGRect) {
 		super.init(frame: frame)
-		styleButton()
+
+		self.buttonCornerRadius = -1
```
mneuwert, line 26: @hosy seen in `ThemeButton` class that negative buttonCornerRadius has a special meaning -> it causes `layer.cornerRadius `to be `bounds.size.height / 2`. Can we may be replace it with some static constant with a good name?

hosy: Yes, makes sense.

hosy: I added new enum values to replace the existing values with new enum values

jesmrec:
Some checks

- [X] Creation of passcode
- [X] Deletion of passcode
- [X] Enable biometrical
- [X] Disable biometrical
- [X] Unlock with passcode
- [X] Unlock with biometrical + cancel -> passcode

everything ok

approved

PR #709, opened 2020-05-20
[feature/cellular-switch] Cellular transfer options

felix-schwarz:
## Description
Adds a UI for `OCCellularManager` to Settings, giving the user more control over what cellular transfers should be used for.

| Settings | Cellular Transfer Settings |
| -- | -- |
| Bildschirmfoto 2020-05-20 um 16 59 08 | Bildschirmfoto 2020-05-20 um 16 59 11 |

## Types of changes

## Checklist:

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.4.0/Cellular%20Switch.md

- [x] (1) Message in case of no-availability https://github.com/owncloud/ios-app/pull/709#issuecomment-641964672 [FIXED]
- [x] (2) Retrigger uploads after enabling setting https://github.com/owncloud/ios-app/pull/709#issuecomment-641970708 [WONT FIX]
- [x] (3) Downloads do not stop when setting is switched off https://github.com/owncloud/ios-app/pull/709#issuecomment-641986526 [WONT FIX]
- [X] (4) Make clearer that the feature switched off does not allow any kind of operation https://github.com/owncloud/ios-app/pull/709#issuecomment-652978305 [FIXED]


felix-schwarz:
**Re (1):** I just updated the branch with the latest SDK. Please check if that fixes the issue for you.

**Re (2), (3), (4):** The current implementation of cellular switches works by determining the value of the respective switch at the time the request for the respective action is sent to `NSURLSession` - and then uses that value to set [`NSURLRequest.allowsCellularAccess`](https://developer.apple.com/documentation/foundation/nsurlrequest/1412032-allowscellularaccess?language=objc) for the request.

Therefore, changes to the cellular switches can and will only be reflected by HTTP requests that are sent *after* the change has been made. Already scheduled HTTP requests will retain their setting.

I agree, though, that this may not always give the best user experience - especially if a request (and therefore an action) remains pending until WiFi becomes available. It won't affect the majority of users, though. So I'd go with this as a first approach and make improvements going forward - the Sync Engine and HTTP Pipeline both offer opportunities for a deeper integration of this feature.

jesmrec:
(1) -> with commit `1503bff1` and sdk `8bb142c` , the issue is still out there.

(2), (3) -> Will set as NA the "Switching `Allow cellular access`" section of the test plan, since it is not implemented yet. Only cases with switching in the WiFi will be taken in account. I'll add a new issue as soon as this is approved.

felix-schwarz:
Re *(1)*: Looking into this, I found a possible way to not only fix that but also make cellular switches a lot more dynamic. Will revisit and dive deeper into that subject later - and then follow up once more.

felix-schwarz:
@jesmrec So, regarding my previous remarks regarding (1), I've started to look at a deeper integration: so, the SDK already has the ability to indicate a set of required signals (such as "online", "authentication available", etc.) for requests, so they only get sent out if all requirements are met - but then, as soon as possible.

What I'd like to do then is to add a new signal for every cellular switch. With the signals summarizing the current state of connectivity (wifi available / cellular available) and cellular switch setting (wifi-only / any connection):
- requests that shouldn't be sent because iOS will return them with an error if wifi is not available (=> which is what prompted (1) in the first place) will not be sent until connectivity is ready for them
- requests that were previously queued to be scheduled as wifi-only will also then be scheduled immediately if the cellular switch is changed
- (1) can be solved in a consistent manner

Implementation will take a bit of time, though, so please feel free to move on with QA on other PRs in the meantime. I'll ping you when the changes are ready @jesmrec.

jesmrec:
Thanks for your effort!

jesmrec:
(1) -> fixed
(2), (3) -> NA, in this PR

felix-schwarz:
Re (4): I changed the message in the Cellular Settings to `Some cellular data may still be used. To completely avoid the usage of cellular data, please turn off access to cellular for the entire app in the Settings app.`.

It's now also only shown when turning `Allow cellular access` off. Hope that makes things clearer. @jesmrec

jesmrec:
Fixed, now it is clearer that switching the feature off, no operations will be allowed.

Approved

Issue #540, opened 2019-10-22
Include new dialog to encourage users to add a new app review in App Store


michaelstingl:
See related Android issue https://github.com/owncloud/android/issues/2071

felix-schwarz:
To ensure that only users who have actually been using the app for a while - and thereby improve the quality and fairness of the reviews - some criteria should be met before prompting the user for a review:
- at least 14 days have elapsed since the first launch of the app (to ensure the app has actually been used for a while)
- at least 7 days have elapsed since the first launch of _this version_ of the app (to ensure the version of the app that the review would be about has actually been used for a while)
- at least one bookmark has been added (to not prompt users who have never set up or connection to a server - and therefore not really used the app)

michaelstingl:
> * at least 130 days have elapsed since last prompting the user for a review (to not annoy the user)

Better 230 days before asking again?

mneuwert:
@michaelstingl just for clarification, do we eventually also want to include deep link to review page in settings for instance, to let users to add reviews any time they feel happy?;-)

Here excerpt from Apple Documentation:

_...In addition, you can continue to include a persistent link in the settings or configuration screens of your app that deep-links to your App Store product page. To automatically open a page on which users can write a review in the App Store, append the query parameter action=write-review to your product URL..._

michaelstingl:
@mneuwert Sure, why not 👍

jesmrec:
Done! ready to 11.4

PR #728, opened 2020-06-10
Include new dialog to encourage users to add a new app review in App Store 

mneuwert:
## Description
See related issue

## Related Issue
#540

## Motivation and Context
Helps to improve App Store review quality and rating

## Types of changes

## Checklist:

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.4.0/Rating%20Dialog.md


ownCloud/Tasks/ScheduledTaskManager.swift
```diff
@@ -184,9 +184,12 @@ class ScheduledTaskManager : NSObject {
 
 		// Add requirements
 		var requirements = [String : Bool]()
-		if self.wifiDetected {
-			requirements[ScheduledTaskAction.FeatureKeys.runOnWifi] = true
-		}
+        var preferences = [String : Bool]()
```
felix-schwarz, line 187: Are the changes in `ScheduledTaskManager.swift` maybe a left-over from a previous/different approach/branch? These changes don't seem to be used/needed by the rest of the PR.

ownCloudAppShared/Tools/AppStatistics.swift
```diff
@@ -0,0 +1,119 @@
...
+    public var timeIntervalSinceFirstLaunch: TimeInterval {
+        return Date().timeIntervalSince(launchDate)
+    }
+
+    public var timeIntervalSinceUpdate: TimeInterval {
+        return Date().timeIntervalSince(updateDate)
+    }
+
+    private var appVersion: String? {
+        guard let version = OCAppIdentity.shared.appVersion, let build = OCAppIdentity.shared.appBuildNumber else {
```
felix-schwarz, line 60: SwiftLint is complaining here (not sure I agree with what it wants, though).

That's what it'd like to see instead:
```suggestion
        guard let version = OCAppIdentity.shared.appVersion, let build = OCAppIdentity.shared.appBuildNumber else { return nil }
```

ownCloudAppShared/Tools/AppStatistics.swift
```diff
@@ -0,0 +1,119 @@
...
+extension TimeInterval {
+    public var days: Int {
+        return Int(self / (3600.0 * 24.0))
+    }
+}
+
+public class AppStatistics {
+
+    enum UserDefaultsKeys: String {
+        case launchDateDefaultsKey, updateDateDefaultsKey, lastVersionDefaultsKey, lastPromptedAboutReviewKey
```
felix-schwarz, line 32: Since this is a `String` enum, the name of the keys in `UserDefaults` will also include the `Key` at the end, so the prefs file will look like:
```xml
<key>launchDateDefaultsKey</key><date>2020-06-12 12:34:56</date>
```
 As the enum is already called `UserDefaultsKeys`, I think the `Key` at the end can be safely removed.

ownCloudAppShared/Tools/AppStatistics.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public static let shared = AppStatistics()
+
+    public var timeIntervalSinceFirstLaunch: TimeInterval {
+        return Date().timeIntervalSince(launchDate)
+    }
+
+    public var timeIntervalSinceUpdate: TimeInterval {
+        return Date().timeIntervalSince(updateDate)
```
felix-schwarz, line 56: No need to change this, just FYI how this can be done without creating a new `Date` object: this (and other similar code in here) could also be written as `return -updateDate.timeIntervalSinceNow`

ownCloudAppShared/Tools/AppStatistics.swift
```diff
@@ -0,0 +1,119 @@
...
+        guard let version = OCAppIdentity.shared.appVersion, let build = OCAppIdentity.shared.appBuildNumber else {
+            return nil
+        }
+        return "\(version).\(build)"
+    }
+
+    init() {
+        let now = Date()
+        launchDate = now
+        updateDate = now
```
felix-schwarz, line 69: I believe `update()` should be called here, too. Otherwise `timeIntervalSinceFirstLaunch` and `timeIntervalSinceUpdate` will return wrong values.

ownCloudAppShared/Tools/AppStatistics.swift
```diff
@@ -0,0 +1,119 @@
...
+            // Persist user defaults
+            userDefaults.synchronize()
+        }
+    }
+
+    public func requestAppStoreReview(onceInDays:Int = 0) {
+        var shallRequest = true
+
+        if onceInDays > 0 {
+            if let lastPrompDate = self.lastReviewPromptDate {
```
felix-schwarz, line 105: Little typo here: `lastPromp` t `Date`.

ownCloudAppShared/Tools/AppStatistics.swift
mneuwert: @felix-schwarz It is called in AppDelegate: `AppStatistics.shared.update()` in `didFinishLaunching` method. This is an intention, otherwise it would a bit uglier construct IMHO` let _ = AppStatistics.shared()`

ownCloudAppShared/Tools/AppStatistics.swift
mneuwert: @felix-schwarz SwiftLint didn't complain for me at this one.. I changed it, but I find both variants equally readable and matching our coding style.

mneuwert: @felix-schwarz fixed

ownCloudAppShared/Tools/AppStatistics.swift
mneuwert: @felix-schwarz fixed

ownCloudAppShared/Tools/AppStatistics.swift
mneuwert: @felix-schwarz fixed

ownCloudAppShared/Tools/AppStatistics.swift
mneuwert: @felix-schwarz fixed

ownCloudAppShared/Tools/AppStatistics.swift
mneuwert: @felix-schwarz fixed

ownCloudAppShared/Tools/AppStatistics.swift
mneuwert: @felix-schwarz Debris is mainly due to use of SwiftLint autocorrect.. And, yes, for the other approach I needed to incorporate fix from other branch into task scheduler which will make it's way into 1.4.

jesmrec:
Fixed some issues regarding negative values. 

The way to test this is turning days on seconds by changing in the file `AppStatistics.swift` the `extension TimeInterval`. The regular code is:

```
extension TimeInterval {
	public var days: Int {
		 eturn Int(self / (3600.0 * 24.0))
	}
}
```

and the code should be replaced for:

```
extension TimeInterval {
	public var days: Int {
		return Int(self)
	}
}

Just two details before the QA stage:

- When the dialog is popped up, the `Submit` option is greyed out... it will be available only for the app downloaded from App Store, won't it?

- The dialog shows `Enjoying ownCloud?` I did not see anything in the code, so... i suppose this String is by default and no changes are posible (parameterized app name, f. ex.)

- It would be needed a mechanism not to show the dialog. For example, customers could not be interested in rating its own branded app, or our app in App Store. What do you think @michaelstingl @hosy ?

jesmrec:
@mneuwert after your changes here: https://github.com/owncloud/ios-app/pull/728/commits/3dcfdebc0ef44fc09d5ac0a275db25bbd49162af

I keep seeing negative values... it's pretty strange, comparing the execution with and without the "minus" :

| With minus | Without minus |
| :-: | :-: |
|Screenshot 2020-06-18 at 10 44 04|Screenshot 2020-06-18 at 10 46 34|

I didn't want to contaminate the code much more to test the real feature as much as posible, so no further changes on my side on the code.

mneuwert:
@jesmrec I will re-visit the code once again..

Regarding review dialog:
- It is fire and forget call.. The dialog appearance and used texts can't be changed. It is entirely managed by iOS
- There is a mechanism - class setting key _"enable-review-prompt"_ similar to _"show-beta-warning"_. By default it is true. But can be overridden. Not sure how it is handled in the branding tool cause I never seen it.

mneuwert:
@jesmrec Please note that after getting feedback from @hosy I also changed the way review requested in the release notes view. I tested it but may be you would like to have another look. 

Now if you tap the footer:

IMG_41B48A6F2B20-1

AppStore will open directly with the page where you can start writing review:

IMG_408A393B4BA5-1

jesmrec:
Thanks @mneuwert for your help. Now, this is approved (test plan attached in the first message).

Please, notice that the code has been changed as mentioned [here](https://github.com/owncloud/ios-app/pull/728#issuecomment-645841323) to turn days into seconds, so the final version/release (using days) will not be the same as the tested one. No problem should happen.

Issue #725, opened 2020-06-08
[FEATURE REQUEST] Adopt iOS13 multiple-selection gesture


mneuwert:
**Is your feature request related to a problem? Please describe.**
There is a new gesture introduced in iOS13: tap and pan with 2 fingers to select multiple items. This has been made an opt-in behaviour.

**Describe the solution you'd like**
For screens using `UITableView` / `UICollectionView` and allowing multiple-selection we just have to adopt 2x new delegate methods to take advantage of the new behaviour in iOS13.

**Describe alternatives you've considered**

**Additional context**
- See https://developer.apple.com/videos/play/wwdc2019/224/ video around 33m10s for reference
- Try to use new gesture in a Files or Notes app to get a feeling for it.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

jesmrec:
Done! ready to 11.4

PR #734, opened 2020-06-17
iOS13 gesture based multiple selection

mneuwert:
## Description
Added delegate methods added to `UICollectionView` and `UITableView` allowing to enter-multiple selection mode with (two) finger pan.

Gestures:
- Two finger pan to start multiple selection of items in the file list
- In photo upload selection view, you can use one finger and pan over multiple items to select them.
- Gestures are iOS13 only and will work as long as "Prevent dragging of files and folders" option in settings is inactive

## Related Issue
#725 

## Motivation and Context
Saves few taps and allows select multiple items faster

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -528,22 +528,20 @@ class ClientQueryViewController: QueryFileListTableViewController, UIDropInterac
 	// MARK: - Navigation Bar Actions
 	@objc func multipleSelectionButtonPressed() {
 
-		if !self.tableView.isEditing {
-			if #available(iOS 13, *) {
-				self.tableView.overrideUserInterfaceStyle = Theme.shared.activeCollection.interfaceStyle.userInterfaceStyle
-			}
-
-			updateMultiSelectionUI()
-			self.tableView.setEditing(true, animated: true)
-			sortBar?.showSelectButton = false
-
-			populateToolbar()
-
-			self.navigationItem.leftBarButtonItem = selectDeselectAllButtonItem!
-			self.navigationItem.rightBarButtonItems = [exitMultipleSelectionBarButtonItem!]
-
-			updateMultiSelectionUI()
+		if #available(iOS 13, *) {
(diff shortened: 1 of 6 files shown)


tdubbins:
I tested the multiple selection of items in the file list with an iPhone 6s Plus and Software Version 13.5.1 and discovered the following problem:
When selecting items using two fingers, the screen freezes for about a minute and I cannot scroll up and down. When I try to select more items, there is a delay of multiple seconds.

mneuwert:
@tdubbins I just tested it.. and checked, the performance issue you have seen is not related to gestures in iOS13. There is code which runs a tight loop over all selected items upon each single item selection / deselection and if number of already selected items is big enough, this loop takes quite some time to execute. I will file a separate issue. Thanks for pointing at this.

jesmrec:
Feature is working fine, and no objection from my side to move forward. Only, one consideration: the setting `Prevent dragging files and folders` is perhaps cryptic for the current feature. Drag&drop has different meaning than other gestures like the tap with two fingers. Is there any general word to describe such actions? `Prevent gestures` , `Prevent finger actions` or similar?

What do you think?

jesmrec:
As discussed, we could go for a better option as improvement. For the moment, we can go with the current one.

So, approved.

hosy:
@jesmrec @michaelstingl what do you think about the new option title with a short subtitle to give the user more context:

Simulator Screen Shot - iPhone 11 Pro - 2020-06-22 at 14 51 44

michaelstingl:
Yeah, let’s try this! :+1:

jesmrec:
WFM @hosy . Let me remove the "approved" label to take another look.

hosy:
@jesmrec just pushed a new commit with the option title changes

mneuwert:
@hosy I just thought may be subtitle like "Prevent doing this and that with fingers" may be sounds a little bit strange. May be change "with fingers" to "using system defined gestures" as an idea? WDYT?

jesmrec:
latest wording works for me. Let's give a try and looking forward for users' feedback

approved again.

Issue #702, opened 2020-05-13
[BUG] Check the crashes from the field in Xcode


mneuwert:
I checked Xcode organizer **Crashes** tab and found quite some number of crashes affecting ownCloud iOS app 1.3.2 (158) and earlier versions. I think we should go through them and if possible identify what potential bugs (which we are not aware of) might have caused them.

- To see the crashes Xcode must be logged in with correct developer account.
- Users must have agreed to send diagnostic information to Apple (so normally you would see only a small fraction and from my experience the real rate of crashes would be much higher).

Example for the crash which apart from File provider XPC connection stuff is most prominent:
![Screenshot 2020-05-13 at 17 37 25](https://user-images.githubusercontent.com/228537/81833892-90529c00-9540-11ea-892f-34964c60431f.png)

Here is how often crash report was uploaded:
![Screenshot 2020-05-13 at 17 38 50](https://user-images.githubusercontent.com/228537/81834154-e1fb2680-9540-11ea-8a80-d71d2bbab2a0.png)

What is also interesting is the **Energy** logs tab, where you can see call stacks which have used high CPU load.. So here we can find hot spots which might require closer attention. For example: `-[OCDatabase _retrieveCacheItemForSQLQuery:parameters:completionHandler:]`

**Metrics** tab is also quite interesting, but somehow there are no logs for ownCloud app due to insufficient usage data. Here normally you can see some statistics on launch time, memory usage, hang rate, battery utilisation and disk writes.

One more thing... I found it in the past, that integrating active crash reporting solution is more reliable than Apple's solution built into Xcode. So we might also consider adding something similar to https://github.com/microsoft/plcrashreporter which used to be a heart of HockeyApp and now MS AppCenter and may be even use it without any fancy backend (like AppCenter or Crashalytics) but just store crash reports along with logs allowing users sharing them with us...

michaelstingl:
[`PLCrashReporter`](https://github.com/microsoft/plcrashreporter) would be a nice addition for the 11.5 👍

jesmrec:
PLCrashReporter added and merged

PR #738, opened 2020-06-24
Added PLCrashReporter to log crashes

mneuwert:
## Description
- Integrated PLCrashReporter as swift package using SPM
- At startup it is checked, if there is pending crash report from previous app session. Crash report is printed as text into the current log.

## Related Issue
#702 

## Motivation and Context
We want to enable users to report bugs related to crashes by sending us logs which will include crash reports.

## How Has This Been Tested?
- Changed the app so that it crashes when camera upload action is triggered by calling `fatalError()`
- Checked the log upon next app launch if it has crash report included.

## Types of changes

## Checklist:


ownCloud/AppDelegate.swift
```diff
@@ -206,3 +208,25 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		}
 	}
 }
+
+import CrashReporter
```
hosy, line 212: Wouldn't it be cleaner to add the import statement on the top of the file?

mneuwert: @hosy moved but not sure if it really contributes to cleanliness

jesmrec:
These are the crash traces i found:

Screenshot 2020-07-08 at 09 16 18

is this correct? i guess so, just to confirm

jesmrec:
then, crash reports are there, from my side it is OK to move forward. I did not test whether the info included in the reports is correct or not, only that causing a crash makes the app write crash errors.

If i'm missing something, let me know. If not, this is approved

Issue #733, opened 2020-06-17
[FEATURE REQUEST] Some features are missing in collections accessible from Quick-Access


mneuwert:
**Is your feature request related to a problem? Please describe.**
When you select one of the collections in Quick Access tab (e.g. Favourites):
- You'll see familiar list of items BUT, following important features for instance are missing:
 - Multiple selection
 - Tap and hold to access actions menu

The reason for this problem is apparently that those features are implemented in `ClientQueryViewController` class and not in `FileListTableViewController` which `LibrarySharesTableViewController` is subclassing.

**Describe the solution you'd like**
I guess relevant functionality can be just moved up to `FileListTableViewController` in the class hierarchy.

hosy:
Should be done in release 11.4 because the features you mentioned were implemented in this release.

hosy:
@mneuwert good catch!

PR #743, opened 2020-07-08
[fix/missing-file-list-actions] Implements missing actions in Quick Access file list

hosy:
## Description
The following actions were missing in the file list, which is used in the `Quick Access` tab:
- Contextual Menu
- Multiple Selection (with toolbar)
- Cell swipe gesture (delete action)

This actions are missing, because this file list is the parent class of the general file list and the actions were defined in the child class. This PR moves the implementation into the parent class.
For me this is not a new feature, it is a bug fix. Most actions were implemented in this milestone and should also be available in all file lists.

## Related Issue
#733 

## Motivation and Context
As a user I want to have the same actions in the Quick Access file list like in the general file list.

## How Has This Been Tested?
Check if all mentioned actions are available in the general file list and in the Quick Access file list.
This actions should not be implemented in the `Pending Shares` list.

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -41,21 +41,8 @@ struct OCItemDraggingValue {
 }
 
 class ClientQueryViewController: QueryFileListTableViewController, UIDropInteractionDelegate, UIPopoverPresentationControllerDelegate {
-	var selectedItemIds = Set<OCLocalID>()
-
-	var actions : [Action]?
-
-	let flexibleSpaceBarButton = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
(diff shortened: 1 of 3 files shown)


jesmrec:
QA Checks in `Quick view` lists:

||iPhone|iPad|
|:-:|:-:|:-:|
|Multiselecion| ✅  | ✅ 
|Contextual menu| ✅  |✅ 
|Cell swipe| ✅  |✅ 

This is only applying to `Collections`, not to `Shares`. Is this the expected behaviour @hosy ?

Shares have the contextual menu available. Neither multiselection nor cell swipe do. But, if a folder is shared, you can browse into it and then, all new actions are again available. Is this expected?

hosy:
@jesmrec I added the cell swipe to the shares too. Multiselection is not available there, because this list differs from the other lists.

jesmrec:
Approved!

Issue #688, opened 2020-04-29
[FEATURE REQUEST] Clarify uploading original assets vs edited assets


mneuwert:
**Is your feature request related to a problem? Please describe.**
iOS Library supports non-destructive editing of media assets. Thereby alongside with an original asset e.g. stored in _HEIC_ format, adjustment data is stored (as _PLIST_) and cached latest output stored as _JPEG_ image (which is basically original image data with operations described in Adjustments.plist applied).

**Describe the solution you'd like**
I think we need at least provide options to the user allowing him to choose what to upload: original, edited image or both.

**Describe alternatives you've considered**
Right now we just upload an original without providing any options

**Additional context**
Here is an example of Array of `PHAssetResouce` objects returned by `PHAssetResource.assetResources(for asset:PHAsset)`:

_[
{ type: photo, uni: public.heic, filename: IMG_3537.HEIC, isCurrent: NO},
{ type: adjustment, uni: com.apple.property-list, filename: Adjustments.plist, isCurrent: YES}
{ type: photo, uni: public.jpeg, filename: FullSizeRender.jpg, isCurrent: YES},
]_

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

jesmrec:
Ready to go in 11.4.1

Issue #685, opened 2020-04-27
[FEATURE REQUEST] Allow uploading of RAW images


mneuwert:
**Is your feature request related to a problem? Please describe.**
There are apps on iOS allowing shooting RAW images (for example: _Halide_). Also, Apple provides developer documentation on how to capture and store such images in the photo library. So, when raw photo is shot typically an Asset with two resources is created: compressed image (_JPEG, HEIF_) and RAW image stored as `PHAssetResourceType.alternatePhoto`. As of now, we would only upload compressed image not giving the user an ability to upload the highest quality ones.

**Describe the solution you'd like**
We could have an option in settings, sth. like "Prefer RAW images" in the upload section.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

jesmrec:
Ready to go in 11.4.1

PR #750, opened 2020-07-16
Display image metadata

mneuwert:
## Description
Extension built allowing to view metadata (EXIF, IPTC) on image files. Licensing: IAP.

## Related Issue
#679 

## Motivation and Context
Useful feature for users who are serious about photography.

## How Has This Been Tested?
Tested on different JPEG, TIFF, PNG, RAW image files.

## Screenshots (if appropriate):
Simulator Screen Shot - iPhone Xs - 2020-07-21 at 18 44 12

Simulator Screen Shot - iPhone Xs - 2020-07-21 at 18 37 26

Simulator Screen Shot - iPhone Xs - 2020-07-21 at 18 38 01 2

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/owncloud/ios-app/blob/master/CONTRIBUTING.md) document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

## QA

- [X] (1) Card does not include feature menu https://github.com/owncloud/ios-app/pull/750#issuecomment-679051019 [FIXED]
- [X] (2) Many decimals in Altitude https://github.com/owncloud/ios-app/pull/750#issuecomment-679054253 [DONE]
- [X] (3) Exposure bias duplicated https://github.com/owncloud/ios-app/pull/750#issuecomment-679055999 [FIXED]

Issue #510, opened 2019-09-25
[UX] Photo upload selection UI improvement suggestions


mneuwert:
### Just a collection of Ideas

Right now we show list of albums and then user can open the album and upload assets from this album.

- [ ] (1) We could add search for album name.
- [ ] (2) Introduce more user-friendly album grouping, consider using grid vs. list or mixed layout.
- [ ] (3) For smart albums we could use some cool icons highlighting their contents similar to Photos app (Slo-Mo, Bursts, Live-Photos, Screenshots etc.)
- [ ] (4) We could have an alternative view showing all assets from the camera roll, may be with photos grouped by date they were taken on. 
- [ ] (5) Support moments as album type
- [ ] (6) For videos and Live Photos we now add small icons as badges. We could have additional badges for HDR, Slo-Mo, burst-photos, panorama etc.
- [ ] (7) We could add additional badge to highlight that photo is a favourite.
- [ ] (8) Add long press gesture in the photo grid view allowing to see bigger preview of the photo with eventually some additional information (map of the location, exif metadata etc.)

---

- [x] (9) Maybe a smarter way to multiselect? Swipe/drag anything? @mneuwert any ideas? [DONE] https://github.com/owncloud/ios-app/issues/725

michaelstingl:
Maybe a smarter way to multiselect? Swipe/drag anything? @mneuwert any ideas?

michaelstingl:
Power/Pro feature to select an album and a define an upload location only for this album? Multiple album ==> location pairs?

Use Case:
WhatsApp has a setting, to archive all new media in "Photos" app. (separate album there) @cdamken want to backup only the WhatsApp media, not the images he took with the Camera app.

/cc @felixboehm

mneuwert:
> Maybe a smarter way to multiselect? Swipe/drag anything? @mneuwert any ideas?

I think we could just adopt two-finger drag to select introduced in iOS13 for table and collection views. May be separate issue would make sense? Improve selection in all screens allowing selection?

michaelstingl:
> * [ ]  We could add search for album name.

does this collide with the iOS 14 image picker?

michaelstingl:
### (4)

> * (4) We could have an alternative view showing all assets from the camera roll, may be with photos grouped by date they were taken on.

This would probably collide with the iOS 14 image picker?

michaelstingl:
### (5)

> * [ ]  (5) Support moments as album type

[...]

PR #759, opened 2020-07-27
Pro Photo Upload Settings

mneuwert:
## Description
Added two photo upload settings in Settings -> Media Upload:
- Prefer original photos (original photo instead of edited version is picked when uploading)
- Prefer RAW (RAW variant is chosen instead of JPEG version, both are usually created by apps like Halide when shooting RAW)

This feature is meant to be added to the same IAP product as EXIF meta-data

## Related Issue
#685 
#688 

## Motivation and Context
For users serious about photography we would like to add more flexibility when uploading photos shot on iPhone.

#### 1. Upload original vs. edited photo 
- Open settings menu in oC App, open "Media upload" sub-menu, activate "Prefer original photos" option.
- Shoot a picture with a stock camera app
- Edit a photo in the Photos app or any other app and e.g. apply "Noir" filter to make the picture B/W
- Open a bookmark, navigate to desired sub-folder and tap + to upload from photo library
- Choose edited photo and upload
- Original photo without applied effects shall be uploaded

#### 2. Upload RAW photos
- Open settings menu in oC App, open "Media upload" sub-menu, activate "Prefer RAW photos" option.
- Shoot a picture with an app able to shoot RAW pictures (e.g. Halide, Manual etc.). Make sure RAW mode is active in the app you have chosen.
- Open a bookmark, navigate to desired sub-folder and tap + to upload from photo library
- Make sure .DNG file is uploaded containing RAW pixel information instead of JPG

IMG_C7685F93A550-1
## Types of changes

## Checklist:

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.4.1/PhotoUploadsIAP.md

Bugs & improvements:

- [X] (1) options available for non-enterprise account if an enterprise account is added https://github.com/owncloud/ios-app/pull/759#issuecomment-696594785 [WONT FIX]
- [X] (2) RAW issues https://github.com/owncloud/ios-app/pull/759#issuecomment-696597017 [CLARIFIED]
- [x] (3) Both Settings enabled https://github.com/owncloud/ios-app/pull/759#issuecomment-697187315 [FIXED]


michaelstingl:
> * Open a bookmark, navigate to desired sub-folder and tap + to upload from photo library
> * Make sure .DNG file is uploaded containing RAW pixel information instead of JPG

This works very nice!

Could we highlight RAW images already in the image picker?

IMG_5C00B2FE77E5-1

mneuwert:
@michaelstingl Yes, absolutely.. Although I didn't explicitly mentioned RAW in issue #510, this is what I had in mind.. But we can add a badge for RAW photos in scope of this PR already.

mneuwert:
@michaelstingl Added RAW photo badge in the upload picker UI

michaelstingl:
> * Open settings menu in oC App, open "Media upload" sub-menu, activate "Prefer original photos" option.

Maybe change to "Prefer unedited photos"

felix-schwarz:
@mneuwert Please fix the merge conflicts before code review.

Does `feature/pro_photo_features` include the changes from `exif_metadata`? If yes, would it make sense to first merge `exif_metadata` - and then focus on just the changes in `feature/pro_photo_features` for code review?

mneuwert:
> @mneuwert Please fix the merge conflicts before code review.
> 
> Does `feature/pro_photo_features` include the changes from `exif_metadata`? If yes, would it make sense to first merge `exif_metadata` - and then focus on just the changes in `feature/pro_photo_features` for code review?

Yes, mainly because it is same IAP product.. yes, would make sense to merge EXIF stuff first, absolutely

felix-schwarz:
@mneuwert Thanks!

@mneuwert @jesmrec If [`exif_metadata`](https://github.com/owncloud/ios-app/pull/750) has been code reviewed and passed QA, please merge it and then fix the merge conflicts of `feature/pro_photo_features`, so I can review only the parts that make up that feature.

(I already tried to achieve that by retargeting this PR from `milestone/11.4.1` towards [`exif_metadata`](https://github.com/owncloud/ios-app/pull/750), but that added even more noise - likely because the two PRs have diverged at some point)

mneuwert:
@felix-schwarz I updated EXIF branch as well as this particular branch has got all required updates so far.. Please have a look again.

ownCloud/Client/PhotoSelectionViewController.swift
```diff
@@ -241,11 +241,11 @@ class PhotoSelectionViewController: UICollectionViewController, Themeable {
...
 		if asset.mediaSubtypes.contains(.photoLive) {
 			cell.mediaBadgeImage = PHLivePhotoView.livePhotoBadgeImage(options: .overContent)
-		}
-
-		if asset.mediaType == .video {
+		} else if asset.mediaType == .video {
 			cell.videoDurationLabel.text = durationFormatter.string(from: asset.duration)
 			cell.mediaBadgeImage = UIImage(named: "camera-badge")?.withRenderingMode(.alwaysTemplate).tinted(with: UIColor.white)
+		} else if PHAssetResource.assetResources(for: asset).filter({$0.type == .alternatePhoto}).count > 0 {
+			cell.mediaBadgeImage = UIImage(named: "raw-badge")?.withRenderingMode(.alwaysTemplate).tinted(with: UIColor.white)
```
felix-schwarz, line 248: In this and other calls following the pattern
```swift
UIImage(named: imageName)?.withRenderingMode(.alwaysTemplate).tinted(with: UIColor.white)
```
the icon is re-computed anew for every cell, and every cell then holds onto its own `UIImage` instance. Computing these icons only once - and sharing them between cells would be a quick performance and memory savings win.

ownCloud/Settings/MediaUploadSettingsViewController.swift
```diff
@@ -28,11 +29,41 @@ class MediaUploadSettingsViewController: StaticTableViewController {
 
 		if let userDefaults = OCAppIdentity.shared.userDefaults {
 			self.addSection(MediaExportSettingsSection(userDefaults: userDefaults))
+
+			let enterpriseAccountAvailable = OCBookmarkManager.shared.bookmarks.filter({$0.edition == .Enterprise}).count > 0
```
felix-schwarz, line 33: Since you wrote that line, a new way to get the number of enterprise accounts has become available: via `OCLicenseEnterpriseProvider.numberOfEnterpriseAccounts`

ownCloud/Settings/MediaUploadSettingsViewController.swift
```diff
@@ -28,11 +29,41 @@ class MediaUploadSettingsViewController: StaticTableViewController {
...
+
 			if OCBookmarkManager.shared.bookmarks.count > 0 {
 				self.addSection(AutoUploadSettingsSection(userDefaults: userDefaults))
 				// TODO: Re-add this section when we re-gain an ability to run background NSURLSessions
 				//self.addSection(BackgroundUploadsSettingsSection(userDefaults: userDefaults))
 			}
 		}
 	}
+
+	private func isProPhotoPackageLicensed() -> Bool {
```
felix-schwarz, line 47: The code in this method is using a snapshot of the current authorization status for that feature - and is dealing with an edge case as well. I see a few issues here:
- the UI only reflects the authorization status at the time the UI is invoked (consider iPad multi-window, where settings can be open while accounts get added or removed)
- handling the edge case here is externalizing part of the internal implementation
- the DispatchGroup could put the app at risk of becoming unresponsive if `OCLicenseManager.shared.perform(afterCurrentlyPendingRefreshes:)` doesn't return promptly - or ever decides to dispatch the provided block to the main thread

Therefore, please use 
```swift
OCLicenseManager.shared.observeProducts(nil, features: [ .photoProFeatures ], in: OCLicenseEnvironment(), withOwner: self) { (licenseObserver, isInitial, authStatus) in
    // call code that adds and removes the section depending on authStatus and number of enterprise accounts
    reconsiderProPhotoSettingsSection()
}
```
instead, which abstracts away the edge case and won't block the main thread.

To also react to changes in the number of enterprise accounts, you could listen for `OCBookmarkManagerListChanged` notifications and call `reconsiderProPhotoSettingsSection()` (from the above example) on that occasion, too.

felix-schwarz (changes requested):
While reviewing, I noticed that https://github.com/owncloud/ios-app/commit/72c2d80aef726f9a18a97cc1ee86a7c74d3e0d54 and https://github.com/owncloud/ios-app/commit/905f48528210c2f716372eb9dd42d2aaa2e59df9 had reverted some code (including a bugfix) as part of merges. IIRC there have been similar artifacts when you resolved conflicts in another long-running PR in the past.

Please check if there's any setting in your git tool of choice that might be causing this (maybe merge strategy `ours` vs `theirs`?).

I brought back the reverted code where I saw it in these two commits.

mneuwert:
@felix-schwarz fixed review findings

ownCloud/Settings/MediaUploadSettingsViewController.swift
```diff
@@ -18,21 +18,71 @@
...
+
+		if OCLicenseEnterpriseProvider.numberOfEnterpriseAccounts > 0 {
+			addProPhotoSettings = true
+		}
+
+		let featureAuthStatus = OCLicenseManager.shared.authorizationStatus(forFeature: .photoProFeatures, in: OCLicenseEnvironment())
+
+		switch featureAuthStatus {
+		case .unknown:
+			OCLicenseManager.shared.observeProducts(nil, features: [ .photoProFeatures ], in: OCLicenseEnvironment(), withOwner: self) { (licenseObserver, isInitial, _) in
```
felix-schwarz, line 65: Ideally an observer should be used regardless of the `featureAuthStatus`, so changes are tracked and reacted to - f.ex. in `viewDidLoad`. The supplied observer block could then be just a call to `reconsiderProPhotoSettingsSection()`.

ownCloud/Settings/MediaUploadSettingsViewController.swift
```diff
@@ -18,21 +18,71 @@
...
+				// call code that adds and removes the section depending on authStatus and number of enterprise accounts
+				if isInitial {
+					OCLicenseManager.shared.stop(licenseObserver)
```
felix-schwarz, line 68: If the `MediaUploadSettingsViewController` is provided as owner of the observer, you don't need to stop the observer: once `MediaUploadSettingsViewController` is deallocated, the observer will be removed automatically.

Stopping the observer when hitting `isInitial` also means that further changes will not be observed.

ownCloud/Settings/MediaUploadSettingsViewController.swift
```diff
@@ -18,21 +18,71 @@
...
+				}
+				self.reconsiderProPhotoSettingsSection()
```
felix-schwarz, line 70: When using `self` in this block, please make it `weak` to not cause a retain cycle (`OCLicenseManager` holds onto `block` - and `block` holds onto `MediaUploadSettingsViewController`).

felix-schwarz (changes requested):

ownCloud/Settings/MediaUploadSettingsViewController.swift
mneuwert: @felix-schwarz implemented

ownCloud/Settings/MediaUploadSettingsViewController.swift
mneuwert: @felix-schwarz done

ownCloud/Settings/MediaUploadSettingsViewController.swift
mneuwert: @felix-schwarz done

felix-schwarz (changes requested):
Thanks for making changes. I think `bookmarksChanged` needs to wrap the call to `reconsiderProPhotoSettingsSection` in `OnMainThread`. However, it would still not update in a multi-window use case when accounts are added or removed, though.

Normally, I'd agree that responding dynamically is an edge case, but since license observers make this so easy to implement, I'd go for it. Here's a version that's fully dynamic (and also shows and hides `AutoUploadSettingsSection` as needed):

```swift
//
//  MediaUploadSettingsViewController.swift
//  ownCloud
//
//  Created by Michael Neuwert on 22.05.20.
//  Copyright © 2020 ownCloud GmbH. All rights reserved.
//

/*
* Copyright (C) 2020, ownCloud GmbH.
*
* This code is covered by the GNU Public License Version 3.
*
* For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
* You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
*
*/

import UIKit
import ownCloudSDK
import ownCloudApp
import ownCloudAppShared

class MediaUploadSettingsViewController: StaticTableViewController {

	private var proPhotoSettingsSection: StaticTableViewSection?
	private var autoUploadSection : AutoUploadSettingsSection?
	private var licenseObserver : OCLicenseObserver?

	deinit {
		NotificationCenter.default.removeObserver(self, name: .OCBookmarkManagerListChanged, object: nil)
	}

	override func viewDidLoad() {
		super.viewDidLoad()
		navigationItem.title = "Media Upload".localized

		if let userDefaults = OCAppIdentity.shared.userDefaults {
			proPhotoSettingsSection = ProPhotoUploadSettingsSection(userDefaults: userDefaults)
			self.addSection(MediaExportSettingsSection(userDefaults: userDefaults))
		}

		NotificationCenter.default.addObserver(self, selector: #selector(reconsiderSections), name: .OCBookmarkManagerListChanged, object: nil)

		licenseObserver = OCLicenseManager.shared.observeProducts(nil, features: [ .photoProFeatures ], in: OCLicenseEnvironment(), withOwner: self) { [weak self] (_, _, _) in
			self?.reconsiderSections()
		}
	}

	@objc private func reconsiderSections() {
		OnMainThread {
			guard let proSettingsSection = self.proPhotoSettingsSection else { return }

			if OCBookmarkManager.shared.bookmarks.count > 0 {
				if self.autoUploadSection == nil, let userDefaults = OCAppIdentity.shared.userDefaults {
					self.autoUploadSection = AutoUploadSettingsSection(userDefaults: userDefaults)
				}

				if let autoUploadSection = self.autoUploadSection, !autoUploadSection.attached {
					self.addSection(autoUploadSection)
				}
				// TODO: Re-add this section when we re-gain an ability to run background NSURLSessions
				//self.addSection(BackgroundUploadsSettingsSection(userDefaults: userDefaults))
			} else {
				if let autoUploadSection = self.autoUploadSection, autoUploadSection.attached {
					self.removeSection(autoUploadSection)
					self.autoUploadSection = nil
				}
			}

			if OCLicenseEnterpriseProvider.numberOfEnterpriseAccounts > 0 || OCLicenseManager.shared.authorizationStatus(forFeature: .photoProFeatures, in: OCLicenseEnvironment()) == .granted {
				if !proSettingsSection.attached {
					self.addSection(proSettingsSection)
				}
			} else {
				if proSettingsSection.attached {
					self.removeSection(proSettingsSection)
				}
			}
		}
	}
}
```

mneuwert:
@felix-schwarz ok, makes sense.. agreed and merged your suggestion

felix-schwarz (approved):

jesmrec:
### (1)

Probably not bug, but a feature:

Options to select are in the `Settings` view, so, they are not under the control of the card to be enabled/disabled. Check these steps up:

1. Add an enterprise account to the device 
2. Add a non-enterprise account to the device (feature not purchased)
3. Open Settings > Media upload

Current: Options are there and available for both accounts. By enabling the `Prefer unedited` , both accounts will upload the unedited version of the pictures. 

Is this behaviour the expected one? 

iPhoneXR iOS13.7
oC server 10.5

jesmrec:
### (2)

To test the RAW feature, i'm using a free app called `RAW+` that is free for a maximum number of shots. Not sure if this is a proper app, because i see different behaviours that do not match the expected ones:

- In the albums, the `RAW` badge  mentioned [here](https://github.com/owncloud/ios-app/pull/759#issuecomment-665236440) is not there.

- iPhoneXR: no matter if i enable or disable the `Prefer RAW` option, the file is always uploaded in `DNG` format. I also have an iPadAIr (older), which uploads always produce a JPG file, no matter the Setting (probably due to an older camera)

Did you test with other devices/apps? which result did you get?

mneuwert:
### Re (1)
@jesmrec That's "works as designed". The settings are not tied to the account. Also, if user buys a feature as IAP it should work with all accounts, right?

jesmrec:
> ### Re (1)
> @jesmrec That's "works as designed". The settings are not tied to the account. Also, if user buys a feature as IAP it should work with all accounts, right?

yes, you're right. You purchase a feature and it is available everywhere. But, the mentioned case is not about purchasing a feature (at least by the user itself). Not sure if this make sense... i will write an example:

I work for owncloud, so i have an enterprise account in the corporate server that allows me to use all pro-features. 
Then, i add and account in my own server, that is not enterprise. So, in my app with these two attached servers:

- Scanner, Shortcuts and Markup: i can use them in enterprise account, but not in my account because i did not purchase anything.
- Pro uploads: i can use it in both accounts, even though i have not purchased it.

is this corner case expected and accepted as feature (not bug)? 

CC @michaelstingl @hosy

michaelstingl:
> is this corner case expected and accepted as feature (not bug)?
> 
> CC @michaelstingl @hosy

Behaviour is fine. 👍  (I don't want to make implementation too difficult – will have proper licensing API in a future backend release)

jesmrec:
ok, then (1) is done.

mneuwert:
### Re (2)
@jesmrec 
- Prefer RAW means: upload RAW if it is available. Photo asset may have multiple resources attached to it. So a photo can have JPG version which is going to be used as preview and RAW image.
- Not all iOS devices have camera which is able to capture RAW.
- Not sure if the app you used is adhering to the Apple's official recommendations summarised in the article (see link below).
- I mostly tested with Halide app running on iPhone 11 Pro. Could also check iPhone X / iPad Pro.
- iPhone XR: most probably RAW+ app is storing only one resource (DNG) in the photo library. Therefore since there is no choice it is picked.
- iPad Air you used eventually doesn't support shooting RAW at all. We could add additional check and disable prefer RAW in case device's camera doesn't support RAW. WDYT?

**Reference:**
https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/capturing_still_and_live_photos/capturing_photos_in_raw_format

jesmrec:
Thanks for the explanations. 

I will try to do it with the stuff i have. If not posible, probably any member of the team can perform some tests in p2p with me. 

> iPad Air you used eventually doesn't support shooting RAW at all. We could add additional check and disable prefer RAW in case device's camera doesn't support RAW. WDYT?

Hidding options that are not really available is always a good UX practice.

jesmrec:
### (3)

1. Take a photo with `Halide` app
2. In Settings, enable both `Prefer unedited` and `Prefer RAW`
3. Upload the pic

Current: Pic uploaded in JPG
Expected: Pic uploaded in DNG

By enabling only the `Prefer RAW` it works fine

iPhoneXR v13.7
app `325a8e33b`

mneuwert:
@jesmrec Addressed (2) and (3). As for (2) made detection of RAW photos in the upload selection view more tolerant. Apple suggests storing RAW as `alternatePhoto` resource type. RAW+ app however does store just one resource and it is main `photo` resource. Also when I try to setup the app to store both JPG and RAW, it only seems to store RAW version.

mneuwert:
@jesmrec Improved UX by not displaying the "Prefer RAW" option in case there is no physical camera HW which would support shooting RAW.

jesmrec:
(2) and (3) are fixed.

Tested with debug ipa build with Xcode12.

jesmrec:
Xcode compilation problems are fixed and now branch is correctly built with Xcode11.7

Approved.

jesmrec (approved):

Issue #735, opened 2020-06-18
[BUG] Slow performance of multiple item selection


mneuwert:
## If you found a bug:
In the file list containing big number of items, the more items are selected, the slower selection / deselection of new items becomes.

### Steps to reproduce
1. Make sure you have a folder with say 200-300 items
2. Select All
3. Try to deselect / select again some items

### Expected behavior
UI should not freeze and items should be selected / deselected without noticeable lag

### Actual behavior
It takes time and UI freezes, so you can't scroll the list for a second or two depending on how capable the iOS device is.

### Client
**iOS version: 13.5.1**

**ownCloud app version: 11.4**

**Device model: iPhone 11 Pro**

### Background
The performance bottle-neck is caused by implementation of multiple-selection in `ClientQueryViewController` class:
- There is a loop running on main thread running over all selected items and then matching them against actions available for the selected collection of items. This is done to make sure that only actions which are allowed to be performed are enabled (e.g. if at least one item in the selected list doesn't have 'delete' permission or is root of the share, then delete action is disabled)
- Action matching mechanism (see `Action` / `ActionContext`) implementation is actually requiring the array of items to check if particular action is available for those items (e.g. delete, duplicate etc). If we change this, we can improve performance dramatically.
- So instead of matching actions against whole list of items on each selection / deselection, we could just keep track of reference counted shared set of few properties for the list of selected items including e.g. item permissions, information about shared items.

**Example:**
Instead of iterating the list and trying to find a single item which doesn't allow certain action, you could have a counted set which keep tracks of how many items are writeable, deletable etc. etc.

On selection, you would increment counters of permissions according to what permissions are set in this item.

On deselection, you would decrement the counters.

felix-schwarz:
Thanks for highlighting a performance bottleneck. Did you benchmark which actions takes the longest to match?

F.ex. some actions are only available for a single item and will only check on the number of items and return immediately if they're above 1.

What I'd therefore suggest is not a change of the whole mechanism, but an extension to it: if we add a dictionary to the `ActionContext` with the summary/counter information, actions that only need that information can return a matching result quickly, based on that information. But if we have an action that needs to really check each item individually, that possibility is also still there and available.

mneuwert:
[...]

PR #764, opened 2020-07-29
Fixes improving performance during multiple selection

mneuwert:
## Description
Before applying those fixes, "Select All" action was taking around 30s on iPhone 11 Pro in a folder with 508 items. Now it is almost instant. Still there is some lag when there is huge number of selected items, and you try to deselect and reselect a single item, but it feels much better. I would probably handle it as separate topic and try to overhaul extension matching mechanism altogether which is a the source of all evil here.

## Related Issue
#735 

## Motivation and Context
Performance of multiple selection was unacceptable.

## Types of changes

## Checklist:


ownCloud/Client/Actions/Actions+Extensions/DeleteAction.swift
```diff
@@ -28,19 +28,12 @@ class DeleteAction : Action {
...
-		let sharedWithUser = forContext.items.sharedWithUser
-
-		if let core = forContext.core {
-			for sharedItem in sharedWithUser {
-				if sharedItem.isShareRootItem(from: core) {
-					return .none
-				}
-			}
+		guard forContext.items.sharedWithUser.count == 0 else {
+			return .none
```
felix-schwarz, line 32: Unfortunately, this check is too simple:
- the shared item itself can only be unshared, not deleted
- any item inside a shared folder may itself be deletable

Simplying the check in this way removes the option to delete a file for all shared items, including those inside a shared folder.

mneuwert: @felix-schwarz please have a look at commit c54e9dc. Tried to bring back required functionality and at the same time to avoid calling `OCItem.isShareRoot()` which spends a lot of time retrieving parent item from the database using `OCCore.retrieveItemFromDatabase()`.

ownCloud/Client/Actions/Actions+Extensions/DeleteAction.swift
```diff
@@ -28,8 +28,22 @@ class DeleteAction : Action {
...
-			return .none
+		let sharedWithUser = forContext.items.sharedWithUser
+
+		if let core = forContext.core {
+			if sharedWithUser.count == 1 {
+				if sharedWithUser[0].isShareRootItem(from: core) {
+					return .none
+				}
+			} else {
+				let sharedFolders = sharedWithUser.filter({$0.type == .collection})
```
felix-schwarz, line 39: This would only check folder items. The shared root item could also be a file, though, in which case, `Delete` should not be available, either.

I see no potential to speed up that check by altering or simplifying the check itself. However, for items in the same folder, what should bring a big boost is caching test results. I've already implemented this in `item.isShareRootItem(from:)` for the `item` that the method is invoked on.

Where I still see good ways for optimization through caching here, however, is the check for the parent folder. In the `isShareRootItem(from:)` method, there's the line

```
if let parentPath = self.path?.parentPath, parentPath != self.path, let parentItem = self.parentItem(from: core) { …
```

with `self.parentItem(from: core)` triggering a queue-scheduled SQLite query, causing the slowness. For multiple items in the same folder, the parent item would always be the same. So, a possible optimization here could be achieved by:
- adding a `[OCLocalID : OCItem]` property `cachedParentFolders` to `ActionContext`.
- add a parameter that takes an `ActionContext` to `isShareRootItem()`, like `isShareRootItem(from: OCItem, in context: ActionContext? = nil)`
- in `isShareRootItem()`, use the `OCItem.parentLocalID` to check if `ActionContext.cachedParentFolders` already has an `OCItem` for the parent folder:
    - if it does, re-use that item as `parentItem`
    - if it does not, request it from the database as before, via `self.parentItem(from: core)`, then store it in `ActionContext.cachedParentFolders[parentItem.localID]` for later retrieval

Now that I've written the above, I realize that directly manipulating `ActionContext.cachedParentFolders` isn't the best/cleanest way to write this probably. A method for retrieving parent items for a particular item - like `ActionContext.parentItem(for: OCItem)` - that transparently handles caching in a dict in a private instance var would be a lot better approach IMO.

Since `ActionContext`s are short-lived, memory should not be a concern. In most cases, that cache would have only a single entry anyway.

And by caching this at the `ActionContext` level, the caching would even benefit checks from multiple actions.

mneuwert: @felix-schwarz thanks for ideas.. I will revisit the implementation with them in mind..

mneuwert:
> Before looking at the actual code, I did a quick test to see if the `Delete` and `Unshare` actions are still shown correctly. Here's what I found:
> 
> ### Setup
> * log in as `admin`
> * share folder `Photos` from `admin` to `demo`
> * share `ownCloud Manual.pdf` from `admin` to `demo`
> * log in as `demo`
> 
> ### Tests
> * invoke action sheet for shared `Photos(2)` folder
>   
>   * expected: action `Unshare` available, action `Delete` not available
>   * actual: none of the actions is available
> * invoke action sheet for shared `Photos(2)/Portugal.jpg` file
>   
>   * expected: action `Delete` available, action `Unshare` not available
>   * actual: none of the actions is available
> * invoke action sheet for shared `ownCloud Manual (2).pdf` file
>   
>   * expected: action `Unshare` available, action `Delete` not available
>   * actual: none of the actions is available
> 
> Please make changes so that these tests pass with expected results.

@felix-schwarz These are fixed..

ownCloud/FileLists/QueryFileListTableViewController.swift
```diff
@@ -444,46 +451,37 @@ class QueryFileListTableViewController: FileListTableViewController, SortBarDele
...
+			updateMultipleSelectButtonItem()
+			updateActions()
 		}
 	}
 
 	override func tableView(_ tableView: UITableView, didDeselectRowAt indexPath: IndexPath) {
 		if tableView.isEditing {
-			updateMultiSelectionUI()
+			if let item = itemAt(indexPath: indexPath), let itemLocalID = item.localID {
+				selectedItemIds.removeAll(where: {$0.hash == itemLocalID.hash})
```
felix-schwarz, line 466: The risk of hash collisions should be really low, yet I'm not sure comparing hashes is a shortcut that should be made.

mneuwert: @felix-schwarz changed this.. I think I took this shortcut since compiler was failing to infer the type correctly.

mneuwert:
@felix-schwarz Finally, I found that _drag-with-two-fingers-to-multiselect_ in iOS13.x was a source of all the evil. I relied on the assumption that methods such as `tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) / tableView(_ tableView: UITableView, didDeselectRowAt indexPath: IndexPath)` are going to be called once per selection / deselection. In fact I've seen them being called multiple times when `func tableView(_ tableView: UITableView, shouldBeginMultipleSelectionInteractionAt indexPath: IndexPath)` was returning `true`.

Nevertheless, I added some checks preventing multiple addition / removal of the same item to / from `ActionContext`.

And I also deactivated multi-finger drag based selection which also has degraded multiple selection performance.

jesmrec:
I have checked the multiselection with:

- [x] 10 items
- [x] 50 items
- [x] 100 items
- [x] 300 items
- [x] 500 items
- [x] 1000 items

in all cases, all items were selected at the moment.

From my side it approved

Issue #756, opened 2020-07-21
[FEATURE REQUEST] Enable crash logging in FileProvider


mneuwert:
**Is your feature request related to a problem? Please describe.**
We added PLCrashReporter which can add crash reports to the log when the main app's process has crashed. We would like to do the same in the FileProvider extension.

**Describe the solution you'd like**
- Link FileProvider extension again PLCrashReporter library and initialise it at startup.
### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

jesmrec:
Done! ready to 11.4

PR #765, opened 2020-07-30
Add PLCrashReported to FileProvider extension

mneuwert:
## Description
Intercept system signals to catch crash causing back-traces in the FileProvider extension. Write crash reports to the common log.

## Related Issue
#756 

## Motivation and Context
We have implemented crash logging in the app, but since FileProvider runs as an isolated process it needs it's own crash handling mechanism.

## How Has This Been Tested?
- Modified the code of FP to trigger NULL pointer access five seconds after it has been initialised.

## Types of changes

## Checklist:


mneuwert:
@felix-schwarz The crash will be collected and then written to log file, next time file provider is loaded (next session however you call it)

mneuwert:
@felix-schwarz 5 sec were related to my approach to crash.. I've added `dispatch_after(5s) { *NULL = ... } `in extension `init() `method

mneuwert:
@felix-schwarz Moved crash reporter initialisation to `-[FileProviderExtension init]` which seems to work more reliably. Now I used the same method to crash the extension in `-[FileProviderExtension itemForIdentifier:] `and on the next launch, log is written.

ownCloud File Provider/FileProviderExtension.m
```diff
@@ -50,6 +51,30 @@ - (instancetype)init
...
+				OCTLogError(@[@"CRASH_REPORTER"], @"%@", crashString);
+			}
+		}
+
+		// Purge the report which we just added to the log
+		[reporter purgePendingCrashReport];
+	}
+
+	// Start intercepting OS signals to catch crashes
+	[reporter enableCrashReporter];
```
felix-schwarz, line 76: I haven't been able to provoke the generation of a crash log, but maybe that's due my testing setup and/or Xcode involvement.

I think the new location in `-[FileProviderExtension init]` could run into a problem, however, since `enableCrashReporter` may only be called once, or the crash reporter will crash by raising an exception in `PLCrashReporter.m:577`.

Since there can be multiple instances of `FileProvider` (one per account AFAIK) in a single process, this would become a problem.

Placing it in `+load` or making sure, via `dispatch_once` (using a static variable as backing) or similar, that this code is only run once, would therefore be important.

Since you know how to provoke a loggable crash already, and can try, I didn't make that change. Before noticing the `enableCrashReporter` issue, I did resolve merge conflicts and added a `CRASH_REPORTER` tag to the line logging the crash report, so that it's logged just like from the app.

mneuwert:
@felix-schwarz Made sure that the code to initialise crash reported is called once per FP extension process instance. Actually I was assuming that also init method would be called once and that every FP instance would be also separate process instance. 

Somehow I was not able to get any crash reports when initialising crash reported at class loading stage in a `+load` method. May be there is some private code in FP API messing with signal mask getting in a way. 

I used the same method to provoke crashes  as you've been suggesting (NULL ptr access in `-[FileProviderExtension itemForIdentifier:]`) and got reports added to the log. So, latest commit seems fine to me.

mneuwert:
@jesmrec Please approve as agreed in the call

jesmrec:
After performing basic checks (building, installation...), everything is OK.

Issue #613, opened 2020-02-13
[FEATURE REQUEST] Number of items inside a folder


jesmrec:
It would be interesting to know how many items are stored in every folder to complement the size information that is currently displayed on the bottom of the screen.

Something like:

```
Total : 40,5MB
4 files and 6 folders
```
or

`Total : 40,5MB (4 files and 6 folders)`

or <your ideas>

michaelstingl:
Just copy iOS Files app

jesmrec:
> Just copy iOS Files app

that makes sense

michaelstingl:
oC10 web UI and Phoenix have it on the bottom of the list:

`15 items | 250 MB`

You can check:
- demo.owncloud.com
- phoenix.owncloud.com

(from: https://github.com/owncloud/ios-app/pull/771#issuecomment-669084505)

PR #771, opened 2020-08-05
[feature/selection-counter-label] File List: Selected items counter label

hosy:
## Description
New Label, which shows how many items are selected in the file list.

## Motivation and Context
Give the user feedback how many items currently selected. Could also very handy, if all items are selected, how many items are inside this folder.

## How Has This Been Tested?
- Open file list
- Tap the `Select` button inside the sort bar
- Select one and more items

| Simulator Screen Shot - iPhone 11 Pro - 2020-08-06 at 13 51 24 | Simulator Screen Shot - iPhone 11 Pro - 2020-08-06 at 13 51 27 |
|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -104,6 +104,10 @@ class ClientQueryViewController: QueryFileListTableViewController, UIDropInterac
 						} else {
 							footerText = String(format: "Total: %@".localized, quotaUsedFormatted)
 						}
+
+						if let self = self {
+							footerText = String(format: "%@ items | ", "\(self.items.count)") + (footerText ?? "")
+						}
 					}
 
 					self?.updateFooter(text: footerText)
@@ -420,7 +424,7 @@ class ClientQueryViewController: QueryFileListTableViewController, UIDropInterac
 	override func performUpdatesWithQueryChanges(query: OCQuery, changeSet: OCQueryChangeSet?) {
 		if let rootItem = self.query.rootItem {
 			if query.queryPath != "/" {
-				let totalSize = String(format: "Total: %@".localized, rootItem.sizeLocalized)
+				let totalSize = String(format: "%@ items | ", "\(self.items.count)") + String(format: "Total: %@".localized, rootItem.sizeLocalized)
 				self.updateFooter(text: totalSize)
 			}
 
--- a/ownCloud/Client/ClientQueryViewController.swift
+++ b/ownCloud/Client/ClientQueryViewController.swift
@@ -87,7 +87,11 @@ class ClientQueryViewController: QueryFileListTableViewController, UIDropInterac
 						}
 
 						if let self = self {
-							footerText = String(format: "%@ items | ", "\(self.items.count)") + (footerText ?? "")
+							if self.items.count == 1 {
+								footerText = String(format: "%@ item | ", "\(self.items.count)") + (footerText ?? "")
+							} else if self.items.count > 1 {
+								footerText = String(format: "%@ items | ", "\(self.items.count)") + (footerText ?? "")
+							}
 						}
 					}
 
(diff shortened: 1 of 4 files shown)


mneuwert:
@hosy I quite often stumbled upon the question: how many items are actually in that folder? As ideas: selected count could be changed to "x of y items are selected". Total number of items could be always shown at the bottom in the table view header for instance..

michaelstingl:
@hosy follow iOS 14 Files app approach for select count? (top center)

ownCloud/FileLists/QueryFileListTableViewController.swift
```diff
@@ -470,6 +470,13 @@ class QueryFileListTableViewController: FileListTableViewController, SortBarDele
 			selectDeselectAllButtonItem?.target = self
 			selectDeselectAllButtonItem?.action = #selector(selectAllItems)
 		}
+		if selectedCount == 1 {
+			sortBar?.updateSelectedLabel(title: String(format: "%d Item selected".localized, selectedCount))
```
mneuwert, line 474: @hosy Not really a complain, but a short side-note: we could consider using .stringdict files for plural localizations

hosy:
@mneuwert good idea. 
@michaelstingl Current solution is using wasted space and the user is not using the information in the title.

mneuwert:
@hosy @michaelstingl Actually I quite like Files app approach: 1) you don't really need to know what folder you are in while you are in selection mode, 2) navigation bar is always visible, whereas, table view header becomes invisible as users scrolls the file list

hosy:
Scrolling is a good point. I agree and will change it.

michaelstingl:
> @hosy I quite often stumbled upon the question: how many items are actually in that folder?

This we should add to the bottom.

`15 items | 250 MB`

Also see oC10 web UI and Phoenix:
- demo.owncloud.com
- phoenix.owncloud.com

jesmrec:
Some checks were performed in iPhone and iPad with different amount of files. I stumbled upon some issues mentioned [here](https://github.com/owncloud/ios-app/pull/764), but the counter is working fine. Approved

hosy:
I also included #613, which shows the number of items of the folder in the footer label.
Furthermore I moved the number of selected items to the navigation bar title.

ownCloud/Client/ClientQueryViewController.swift
```diff
@@ -158,6 +139,29 @@ class ClientQueryViewController: QueryFileListTableViewController, UIDropInterac
 		leaveMultipleSelection()
 	}
 
+	private func updateTitle() {
+		let lastPathComponent = (query.queryPath as NSString?)!.lastPathComponent
+
+		if lastPathComponent.isRootPath, let shortName = core?.bookmark.shortName {
+			self.navigationItem.title = shortName
+		} else {
+			let titleButton = UIButton()
```
mneuwert, line 148: @hosy do we need this during selection (button to present breadcrumb UI)? Otherwise setting title textAttributes would be probably just fine

hosy: No, this was set after the selection did end. Otherwise the title bar gets not updated.

mneuwert: @hosy ok

hosy:
@jesmrec please check again after the changes I made

jesmrec:
### (1)

After attaching an account for the first time, it shows `0 items` . By browsing back and reentering, it shows the correct information:

ezgif com-optimize (6)

iPhoneXR 13.6

jesmrec:
### (2)

Do you think is it posible to separate items into files and folders?

jesmrec:
### (3)

a minor detail. If a folder contains only 1 item, it shows `1 items` instead of `1 item`

hosy:
@jesmrec (1), (3) is fixed. If the number of items is not available, I counter label

jesmrec:
### (4)

About moving the multiselection counter to the top bar:

Screenshot 2020-08-05 at 17 49 08

Depending on the language, the words `Selected` and `Deselect` could be even longer, so the top bar looks a bit overloaded.

F.ex, in spanish:

`Selected` = `Seleccionados`
`Deselect` = `Deseleccionar`

Screenshot 2020-08-05 at 17 55 11

is this standard of iOS? as personal insight i liked more the place before for the counter, in the same place as the multiselection option. What do you think? @mneuwert @michaelstingl @felix-schwarz

mneuwert:
### Re (4)
IMHO one of the solutions could be use icons instead of words.. or may be indeed just show number of selected items like Files app does in the navigation bar which contradicts of course what I asked for before.. sorry about that.

michaelstingl:
Please try:
- selected item on the top
- total items on the bottom of the list

hosy:
@jesmrec give it a new try, I switched back to short strings

jesmrec:
Everything OK @hosy 

Approved

Issue #428, opened 2019-06-30
[UI / PDF] Full screen view for PDFs


michaelstingl:
In the PDF view, the toolbar is always shown and the miniatures are always shown. Many PDF viewers can switch to a fullscreen view…

mneuwert:
@michaelstingl Played around with adding tap gesture recognisers and came to the conclusion:
- We can add tap to full-screen
- But we will loose double tap zoom... And that's not easy to fix. Tried some hacky approaches but failed so far..

michaelstingl:
> * But we will loose double tap zoom... And that's not easy to fix. Tried some hacky approaches but failed so far..

Try that with the 11.5 TestFlights, maybe with the 11.5 stable as well, then wait for user feedback? We can always revert…

mneuwert:
@michaelstingl @hosy Changed couple of things due to issues with thumbnail view which sort of started behaving correctly on iOS13, before it has been behaving pretty badly causing all sorts of crashes when it's properties (layout, thumbnail size) have been changed dynamically. So, I decided to provide cool stuff on iOS 13 or higher:

- Full screen when PDF view is tapped
- Dynamically adapting thumbnail sizes
- Vertical thumbnail strip in landscape on iPhone (compact vertical size class).
- Background colour of thumbnail strip matching system default PDF view background color which is nicer than the one used prior to iOS 13

jesmrec:
Done

PR #813, opened 2020-10-09
[UI / PDF] Full screen view for PDFs (iOS >= 13)

mneuwert:
## Description
- Tap to trigger full screen view
- Thumbnails positioned based on vertical size class after rotating the device to give the displayed document more screen real estate.
- Some other small UI tweaks, but all the goodness only available for iOS 13.x and higher though

## Related Issue
#428 

## Motivation and Context
Better UX when viewing PDFs

## Types of changes

## Checklist:

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.5/PDF%20View%20Improvements.md

- [x] (1) Some glitches in thumbnails https://github.com/owncloud/ios-app/pull/813#issuecomment-735696396 [FIXED]
- [x] (2) Thumbnails are off in iPadOS13 https://github.com/owncloud/ios-app/pull/813#issuecomment-737131706  [FIXED]
- [x] (3) Progress bar does not run smoothly https://github.com/owncloud/ios-app/pull/813#issuecomment-737135559 [WONT FIX]


hosy:
@mneuwert thank you for this nice feature. Before I will start with code review I have some findings, which may be fixed before:

- Status bar color should be changed to `.darkContent` in fullscreen view
- Thumbnail icons on iPad should always shown on the right side not on bottom (in portrait and landscape mode)
- on iOS 12 fullscreen is not working

mneuwert:
> @mneuwert thank you for this nice feature. Before I will start with code review I have some findings, which may be fixed before:
> 
> * Status bar color should be changed to `.darkContent` in fullscreen view
> * Thumbnail icons on iPad should always shown on the right side not on bottom (in portrait and landscape mode)
> * on iOS 12 fullscreen is not working

- Valid point about status bar I guess..
- Thumbnail icons positioning is rather bound to size class rather than to particular HW. So, please be aware that in multitasking mode, the app could just look like an iPhone app.
- If I enable this stuff on iOS, it is badly crashing all the time.. unfortunately. Thumbnails view seems to be very sensitive to layout changes there. It ends up in some "unrecognised selector sent to blah" exception quite often.

hosy:
> > @mneuwert thank you for this nice feature. Before I will start with code review I have some findings, which may be fixed before:
> > 
> > * Status bar color should be changed to `.darkContent` in fullscreen view
> > * Thumbnail icons on iPad should always shown on the right side not on bottom (in portrait and landscape mode)
> > * on iOS 12 fullscreen is not working
> 
> * Valid point about status bar I guess..
Cool!

> * Thumbnail icons positioning is rather bound to size class rather than to particular HW. So, please be aware that in multitasking mode, the app could just look like an iPhone app.
Yes, but currently on the iPad the thumbnail view is always on bottom. But for me it makes more sense to put in on the right side, when a larger size class is active.

> * If I enable this stuff on iOS 12, it is badly crashing all the time.. unfortunately. Thumbnails view seems to be very sensitive to layout changes there. It ends up in some "unrecognised selector sent to blah" exception quite often.

At the moment we shipping the app with iOS 12 support, so we need a running workaround for this version.

hosy:
I had a short call with @mneuwert about the following findings:
- Fullscreen support only for iOS >= 13 
- show thumbnail view (bottom or right) in dependency on current size class

mneuwert:
@hosy I tried pretty much everything.. And I didn't manage to get these status bar appearance callbacks being called... `childForStatusBarStyle`, `preferredBarStatusStyle`.. Not sure what's wrong or may be something is broken. Tried to build up the correct chain between parent and child (`DisplayHostViewController` -> `PDFViewController`) but somehow it is not working at all. Also explicitly set a property in Info.plist to allow view controller based status bar management... Any ideas?

hosy:
@mneuwert sometimes the PDF view does not render the PDF file because of this crash:

```
 [Unknown process name] Error: this application, or a library it uses, has passed an invalid numeric value (NaN, or not-a-number) to CoreGraphics API and this value is being ignored. Please fix this problem.
2020-10-21 20:46:08.275657+0200 ownCloud[67923:14333676] [Unknown process name] Backtrace:
  <_ZN12_GLOBAL__N_118transform_is_validEPK17CGAffineTransform+23>
   <CGPathCreateWithRect+23>
    <CGContextAddRect+218>
     <op_re+201>
      <pdf_scanner_handle_xname+130>
       <CGPDFScannerScan+486>
        <CGContextDrawPDFPageWithDrawingCallbacks+3654>
         <CGContextDrawPDFPageWithProgressCallback+47>
          <-[PDFPage _drawWithBox:inContext:withRotation:isThumbnail:withAnnotations:withBookmark:withDelegate:]+669>
           <-[PDFPage drawWithBox:inContext:isThumbnail:]+88>
            <-[PDFView drawPage:toContext:]+74>
             <-[PDFTilePool _renderTileForRequest:]+732>
              <_dispatch_call_block_and_release+12>
               <_dispatch_client_callout+8>
                <_dispatch_lane_serial_drain+796>
                 <_dispatch_lane_invoke+439>
                  <_dispatch_workloop_worker_thread+882>
                   <_pthread_wqthread+290>                    <start_wqthread+15>
```

hosy:
@mneuwert maybe one reason, why the statusbar color cannot be change, is the overlaying `ThemeNavigationController` which handles the global `preferredStatusBarStyle`.

Another note: Currently on iOS 12 no thumbnail view is visible.

mneuwert:
@hosy Good point about the status bar style.. Do you have more details on the crash eventually? Remember you said it happens with the user manual for OC. May be reproduction steps?

mneuwert:
@hosy can you please check if the crash is reproducible with latest changes? As for status bar, I am not sure.. probably we need to think a bit more how it is handled globally considering also current UI mode: dark / light

hosy:
@jesmrec probably the PDF crash depends on the an Apple bug in PDFKit. If you run into this problem, please check if you can reproduce it.

hosy (approved):

hosy:
@mneuwert the thumbnail view does not appear immediately on all devices. After a rotation it appears. 
Furthermore in landscape view on the iPad it should be shown on the right, not on the bottom.

jesmrec:
### (2)

Probably, related to (1)

Reproducible with an iOS13 iPadAir

1. Download a PDF file, not opened before

Current: After downloading, file is displayed and thumbnails do not appear. Changing orientation, they do not appear either. Closing the file and reopening, thumbnails appear.

Expected: After first downloading, thumbnails appear.

iPadOS13, iPadAir.

jesmrec:
### (3)

This one is a regression, not a matter of this PR. Just for getting some information.

The progress bar just above the page indicator:

Screenshot 2020-12-02 at 11 18 16

behaves in a strange way. It seems to have only three positions (on the left, on the right and in center), instead of moving smoothly through the screen width. Is there something to do? 

As i clarified above, this is not a matter of this PR. In case it could be fixed, it's not a problem to do it in a separate ticket.

mneuwert:
### Re (3)
- The scrollview which displays the scroll bar is entirely managed by `PDFView` which comes from Apple's `PDFKit`. So we can't really fix it.
- I found another issue..  The white horizontal stripe below the scroll bar. It shouldn't have been there. Now removed.

mneuwert:
### Re (1) & (2)
Should be fixed with latest commit

jesmrec:
(1) and (2) fixed

jesmrec:
Approved on my side. There are some conflicts before merging, could you take a look?

jesmrec (approved):

mneuwert:
@jesmrec Fixed merge conflicts

Issue #418, opened 2019-06-26
[UX / PDF] Not possible to navigate between search results


mneuwert:
Right now if user searches in PDF, he will be presented list of search matches and as soon as he taps on one of the hits, PDF viewer navigates to the page where it is found and highlights the search hit. However all other matches are lost.

So, it would be nice to have some temporary overlay (as an idea) showing number of search matches as well as which match is currently being presented (e.g. 2 of 10) and may be to have "back" and "forth" buttons allowing to navigate between different matches.

See an example from PDF Pro 3 app:
![IMG_FABA22791795-1](https://user-images.githubusercontent.com/228537/60198904-fbb94f80-9842-11e9-942c-630c446217b8.jpeg)

jesmrec:
Done!

PR #836, opened 2020-11-15
[PDF] Search Results Navigation

mneuwert:
## Description
Additional UI is displayed on top of PDF document after the search has been performed. It allows to jump back and forth between found search hits.

## Related Issue
#418 

## Motivation and Context
As a user I would like to be able to jump back and forth between search results to easier find the one I was looking for.

Screenshot 2020-11-15 at 21 39 31

## Types of changes

## Checklist:

## QA

QA checks: https://github.com/owncloud/ios-app/pull/836#issuecomment-729611334

Bugs & improvements:

- [X] (1)Items selected besides of highlighted https://github.com/owncloud/ios-app/pull/836#issuecomment-729623867   [FIXED]
- [X] (2) Clicking area seems to be small https://github.com/owncloud/ios-app/pull/836#issuecomment-729625931 [FIXED]


hosy (approved):

hosy:
@mneuwert Great feature and implementation!

jesmrec:
QA checks:

- [X] 0 results -> no feature
- [X] 1 result -> no navigation
- [ ] +1 result -> navigation displayed portrait
- [ ] +1 result -> navigation displayed landscape
- [ ] Navigation: go forth
- [ ] Navigation: go back
- [ ] Navigation: close

jesmrec:
### (1)

Not sure if this is an issue or a feature:

1. Open PDF
2. Click lens icon to search
3. Enter some string in order to get many results in different pages.
4. Select the first result of the list

Current: 

First result is correctly highlighted:

Screenshot 2020-11-18 at 12 21 47


Browse to the next result with the navigation bar that is in another page. This result is not only highlighted, is also selected:

Screenshot 2020-11-18 at 12 23 12


From now on, while browsing every result will be selected, displaying also the contextual menu. This is not bad, but i stumbled upon glitches like:

Screenshot 2020-11-18 at 12 25 28

in which the contextual menu overlaps the navigation bar

- Is correct that the results are selected instead of highlighted? 
- In case selection is the correct way to do, the first result should be also selected (now, only highlighted)

iPhoneXR iOS14.1
iPadAir2 iPadOS 13.7
`f262bcd`

jesmrec:
### (2)

Another thought that came to my mind while i was checking the feature with the iPhone: i realised that the clicking area around the navigation arrows seems to be kind of small, so the tap must be pretty accurate to navigate. Someone with big fingers could have some problems to use the menu. What do you think?

michaelstingl:
Auto-selection the 2nd result isn’t necessary. Highlighting it would be sufficient.

mneuwert:
@jesmrec 

### Re (1)
Seen this when compiled with Xcode 11.7. With Xcode 12 selection menu doesn't appear when result is highlighted.. And I also can't built with Xcode 11 anymore since @hosy has updated the branch. Tested on iPhone X running iOS 13.6 and iPhone 11 Pro running iOS 14.2.

### Re (2)
Made touch targets slightly bigger

jesmrec:
(1)  Actually, i use Xcode 12.1 (12A7403) to compile. This is not a very bad behaviour or a blocker, but i still reproduce it with iPhone 14.1 and iPad 13.7. Could it be a matter of the iOS version?

(2) fixed

mneuwert:
> (1) Actually, i use Xcode 12.1 (12A7403) to compile. This is not a very bad behaviour or a blocker, but i still reproduce it with iPhone 14.1 and iPad 13.7. Could it be a matter of the iOS version?
> 
> (2) fixed

Regarding (1).. what annoys me the most is that behaviour is inconsistent. I changed the way we jump to a search hit slightly. Lets see if it brings some improvement on your side.. If it doesn't I could just jump to the page where the hit was found, but don't highlight the actual word, then I guess the selection menu won't appear. However it would be harder for the user to follow where on the page the search hit is actually located.

jesmrec:
(1)

checking the last commit `3af51af`, behaviour in both iPad and iPhone has been improved. The selection menu is not there anymore at the time the results are highlighted. 

As you've commented, different devices/version could show tiny differences, but probably, this is something we can improve by iterations in case new feedback arrives to us. 

From my side we can move this forward.

mneuwert:
@jesmrec agree! Thanks!

PR #850, opened 2020-12-03
[feature/enhanced-drag-drop] Enhanced drag & drop support

felix-schwarz:
## Description
Fix drag and drop and improve support to run the iOS app on M1 Macs:
- add drag-out support for files that are not locally available yet
- improve drag-in support for files, picking the best available representation that can be retrieved as data
- support for drag & drop in the log file browser

## Types of changes

### DIFF
--- a/ownCloud/Settings/LogFilesViewController.swift
+++ b/ownCloud/Settings/LogFilesViewController.swift
@@ -242,13 +242,11 @@ class LogFilesViewController : UITableViewController, UITableViewDragDelegate, T
 
 				try FileManager.default.copyItem(atPath: logRecord.url.path, toPath: shareableLogURL.path)
 			} catch {
+				Log.error("Error providing copy of log \(logRecord.url.path) at \(shareableLogURL.path): \(error)")
 			}
 		
 			completionHandler(shareableLogURL, shareableFileName, {
-				do {
-					try FileManager.default.removeItem(at: shareableLogURL)
-				} catch {
-				}
+				try? FileManager.default.removeItem(at: shareableLogURL)
 			})
 		} else {
 			completionHandler(logRecord.url, shareableFileName, nil)
--- a/ownCloud/Settings/LogFilesViewController.swift
+++ b/ownCloud/Settings/LogFilesViewController.swift
@@ -17,7 +17,10 @@
 */
 
 import UIKit
+import CoreServices
+
 import ownCloudSDK
+import ownCloudApp
 import ownCloudAppShared
 
 class LogFileTableViewCell : ThemeTableViewCell {
@@ -57,7 +60,7 @@ extension OCLogFileRecord {
 	}
 }
 
-class LogFilesViewController : UITableViewController, Themeable {
+class LogFilesViewController : UITableViewController, UITableViewDragDelegate, Themeable {
 
 	var logRecords = [OCLogFileRecord]()
 
@@ -87,6 +90,8 @@ class LogFilesViewController : UITableViewController, Themeable {
 		super.viewDidLoad()
 		Theme.shared.register(client: self, applyImmediately: true)
 		self.tableView.register(LogFileTableViewCell.self, forCellReuseIdentifier: LogFileTableViewCell.identifier)
+		self.tableView.dragDelegate = self
+		self.tableView.dragInteractionEnabled = true
 		self.title = "Log Files".localized
 	}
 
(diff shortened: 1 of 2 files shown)


ownCloudAppShared/Client/File Lists/ClientQueryViewController.swift
```diff
@@ -476,10 +478,38 @@ extension ClientQueryViewController: UITableViewDropDelegate {
...
+					kUTTypeMovie,
+					kUTTypePDF,
+					kUTTypeData
+				]
+				var useUTI : String?
+				var useIndex : Int = Int.max
+				
+				for typeIdentifier in typeIdentifiers {
+					if typeIdentifier != ItemDataUTI, !typeIdentifier.hasPrefix("dyn.") {
+						for preferredUTI in preferredUTIs {
```
mneuwert, line 495: @felix-schwarz does it make sense to iterate over all preferredUTIs? Could also just `filter()` conforming one? Anyway first conforming UTI would win here.

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -192,31 +231,50 @@ class LogFilesViewController : UITableViewController, Themeable {
...
+				try FileManager.default.copyItem(atPath: logRecord.url.path, toPath: shareableLogURL.path)
+			} catch {
 			}
-
-			try FileManager.default.copyItem(atPath: logRecord.url.path, toPath: shareableLogURL.path)
-		} catch {
+		
+			completionHandler(shareableLogURL, shareableFileName, {
+				do {
+					try FileManager.default.removeItem(at: shareableLogURL)
```
mneuwert, line 249: @felix-schwarz just use `try?` or do something in `catch` block: add log entry for instance?

ownCloud/Settings/LogFilesViewController.swift
```diff
@@ -192,31 +231,50 @@ class LogFilesViewController : UITableViewController, Themeable {
...
-		let shareableLogURL = FileManager.default.temporaryDirectory.appendingPathComponent(shareableFileName)
+			do {
+				if FileManager.default.fileExists(atPath: shareableLogURL.path) {
+					try FileManager.default.removeItem(at: shareableLogURL)
+				}
 
-		do {
-			if FileManager.default.fileExists(atPath: shareableLogURL.path) {
-				try FileManager.default.removeItem(at: shareableLogURL)
+				try FileManager.default.copyItem(atPath: logRecord.url.path, toPath: shareableLogURL.path)
```
mneuwert, line 243: @felix-schwarz just use try? or do something in catch block: add log entry for instance?

mneuwert (changes requested):

ownCloudAppShared/Client/File Lists/ClientQueryViewController.swift
```diff
@@ -476,10 +478,38 @@ extension ClientQueryViewController: UITableViewDropDelegate {
...
+					kUTTypeImage,
+					kUTTypeMovie,
+					kUTTypePDF,
+					kUTTypeData
+				]
+				var useUTI : String?
+				var useIndex : Int = Int.max
+				
+				for typeIdentifier in typeIdentifiers {
+					if typeIdentifier != ItemDataUTI, !typeIdentifier.hasPrefix("dyn.") {
```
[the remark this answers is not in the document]
felix-schwarz: `filter({ $0 != ItemDataUTI && !$0.hasPrefix("dyn.")})` would add an extra iteration over the array and create a new array from the filter results, over which the code would then need to iterate again.

The way it is now, the array is iterated only once and no additional array needs to be created.

ownCloudAppShared/Client/File Lists/ClientQueryViewController.swift
felix-schwarz: As in the comment above, it'd add an extra iteration and creating additional objects. An overhead that is avoided this way.

ownCloud/Settings/LogFilesViewController.swift
felix-schwarz: In this particular instance, I use it so that, if `removeItem` fails, `copyItem` is not attempted. I'll add logging, though.

ownCloud/Settings/LogFilesViewController.swift
felix-schwarz: Indeed, `try?` is a better choice here. Changing that.

mneuwert (approved):

jesmrec:
Some questions regarding this PR:

> add drag-out support for files that are not locally available yet

that means, a file that is not locally available (not downloaded) is able to be dragged?. That happened also before AFAIK. I assume that this is in the same account, not posible to drag&drop between accounts without downloading. I'm missing something here

> improve drag-in support for files, picking the best available representation that can be retrieved as data

could you clarify? what does "best available representation" mean? (thanks in advance)

> support for drag & drop in the log file browser

this one works fine 

OTOH, should i expect copying text and dragging into an account to create a new txt file? (i think we commented this somewhen). In case it is, is this dependant from the version? the iPad i'm using in iOS13

hosy:
> Some questions regarding this PR:
> 
> > add drag-out support for files that are not locally available yet
> 
> that means, a file that is not locally available (not downloaded) is able to be dragged?. That happened also before AFAIK. I assume that this is in the same account, not posible to drag&drop between accounts without downloading. I'm missing something here
Yes, we had this behavior already, because it is not supported to download an item while dragging. Only locally available files can be dragged out. 
Dragging inside the app is possible, because we only need the file id for copying between accounts.

> 
> > improve drag-in support for files, picking the best available representation that can be retrieved as data
> 
> could you clarify? what does "best available representation" mean? (thanks in advance)

A file could have multiple representation and I the best format for saving will be chosen.

> 
> > support for drag & drop in the log file browser
> 
> this one works fine
> 
> OTOH, should i expect copying text and dragging into an account to create a new txt file? (i think we commented this somewhen). In case it is, is this dependant from the version? the iPad i'm using in iOS13

I think it should be possible to drag text inside, which should be saved as a new txt file.

felix-schwarz:
> > add drag-out support for files that are not locally available yet
> 
> that means, a file that is not locally available (not downloaded) is able to be dragged?. That happened also before AFAIK. I assume that this is in the same account, not posible to drag&drop between accounts without downloading. I'm missing something here

It means you can drag a file that is not yet downloaded - and when you drop it, the OC app will download the file and deliver it to the target you dropped it on.

> > improve drag-in support for files, picking the best available representation that can be retrieved as data
> 
> could you clarify? what does "best available representation" mean? (thanks in advance)

When an item is dragged in, it can offer more than one representation. Previously, the last in the list was picked, whatever it was. Now the implementation prefers images, then movies, then PDFs, then raw data. IIRC previously we had the issue that on an M1 Mac mini, a file was dragged in - and instead of the file, a placeholder (stored as last type) was imported.

> > support for drag & drop in the log file browser
> 
> this one works fine
> 
> OTOH, should i expect copying text and dragging into an account to create a new txt file? (i think we commented this somewhen). In case it is, is this dependant from the version? the iPad i'm using in iOS13

I'd have to look into this to see what data types we are offered for this.

Issue #847, opened 2020-11-29
Check: Upload video files without re-conversion


mneuwert:
Related:
https://github.com/owncloud/ios-legacy/issues/1040

==> @mneuwert

mneuwert:
@michaelstingl Shall we eventually add "Prefer original videos" as new upload option? Technically, slo-mo for instance is edited once: 120fps shot video + transformation applied. So if you upload original it is going to be a video shot at higher FPS but without any slo-mo effect.

michaelstingl:
Yes, “prefer original video” would be perfect 👍

michaelstingl:
> @michaelstingl Shall we eventually add "Prefer original videos" as new upload option?

Yes, sounds cool 👍

jesmrec:
Done

PR #852, opened 2020-12-19
Video upload improvements

mneuwert:
## Description
- Add ability to upload slo-mo videos etc
- Add option to allow uploading original videos

## Related Issue
#847 

## Motivation and Context
More flexibility for photo uploads

## Types of changes

## Checklist:


michaelstingl:
> * Add ability to upload slo-mo videos etc

Looks very good 👍 

CleanShot 2020-12-23 at 01 38 59

michaelstingl:
Hm, interesting. Apple's new 10bit Dolby Vision format doesn't get re-encoded at all:

CleanShot 2020-12-23 at 01 52 11

michaelstingl:
I found, "Prefer original videos" is only respected when upload is initiated from inside the app. Video uploads initiated from Photos app via Share Sheet or via FileProvider always have "Preparing" phase before upload, so they get re-encoded. Add a note under the switch?

mneuwert:
@michaelstingl hm.. actually these shall be uploaded as is without any processing. It can be any binary file not just video, right?

michaelstingl:
Check Settings > Media upload:

“Prefer original videos” and “Prefer RAW photos” seems connected. Enabling the one, also enabled the other. Disabling results in same behavior:


https://user-images.githubusercontent.com/214010/103462625-d0ac9780-4d26-11eb-86b1-41ff39c67d72.MOV

mneuwert:
> Check Settings > Media upload:
> 
> “Prefer original videos” and “Prefer RAW photos” seems connected. Enabling the one, also enabled the other. Disabling results in same behavior:
> 
>  B6FD2EC3-ECAC-49D2-85FF-D78277CC2082.MOV

This is fixed with latest commit

fastlane/Fastfile
```diff
@@ -613,14 +613,14 @@ end
...
-      add_badge(custom: "fastlane/badge.png")
-    end
+    #if values[:BETA_APP_ICON]
+    #  commit = last_git_commit
+    #  short_hash = commit[:abbreviated_commit_hash] # short sha of commit
+    #  sh "brew install librsvg"
+    #  sh "sed -e \"s/\#version#/" + version + "/\" -e \"s/\#githash#/" + short_hash + "/\" badge.svg > badge_tmp.svg"
+    #  sh "rsvg-convert badge_tmp.svg > badge.png"
+    #  add_badge(custom: "fastlane/badge.png")
+    #end
```
hosy, line 623: This needs to be uncommented

ownCloud/Settings/ProPhotoUploadSettingsSection.swift
```diff
@@ -98,14 +109,20 @@ class ProPhotoUploadSettingsSection: SettingsSection {
 
 		self.add(row: preferOriginalsRow)
 
-		if AVCaptureDevice.rawCameraDeviceAvailable() {
```
hosy, line 101: Why did you remove the `if` clause? Raw camera support is not available on every device.

hosy (changes requested):

ownCloud/Settings/ProPhotoUploadSettingsSection.swift
michaelstingl: > Why did you remove the `if` clause? Raw camera support is not available on every device.

What about raw files stored on devices without the latest camera? I created the file in iPhone 12 Pro Max, but did my upload testing with the iPhone 8 test device.

hosy: Ok, good point. I missed that option!

fastlane/Fastfile
felix-schwarz: I commented that out because it failed and prevented Bitrise builds a few weeks ago (IIRC). When putting it back in, please check whether IPAs are still being built.

hosy: Should be fixed with the latest `master` merged inside

felix-schwarz (approved):

ownCloud/Settings/ProPhotoUploadSettingsSection.swift
mneuwert: @hosy that was exactly the intension to allow users upload RAW photos from their iCloud synced photo library shot on more capable device.

jesmrec:
Checked the slo-mo videos, they are correctly uploaded and played. @michaelstingl collected all relevant encoding info above. 

About uploading original videos:

- [x] Settings option works fine. It is only visible in enterprise version servers, and under `Prefer original videos` IAP feature
- [x] Option enabled + video edited = upload video without the edit. 
- [x] Option disabled + video edited = upload video with the edit.
- [x] An original video is always correctly uploaded, no matter the Setting value

About encoding, i will add a report about the edited video with and without the Setting enabled. For me it's ok, but you can take a look if any parameter is more relevant

| Original | Edited |
| :-: | :-: |
|Screenshot 2021-01-15 at 11 13 59|Screenshot 2021-01-15 at 11 14 21|

jesmrec:
Approved

jesmrec (approved):

Issue #851, opened 2020-12-06
Deal with iOS 14 Photo permissions


mneuwert:
See e.g.
https://www.andyibanez.com/posts/understanding-limited-photo-library-ios-14/

fschade:
steps to reproduce

1.) Log into your app and connect it to a server
2.) hit the plus icon on the right top side and select „upload from your photo library“
3.) iOS asks you which collections or photos you wanna make available, select a few
4.) close overlay
5.) hit the plus icon on the right top side and select „upload from your photo library“ again
6.) only previously white listed photos or collections available, now try whitelist more photos or collections

expectations: a button or setting where I am able to select what parts are available for the OC iOS app

michaelstingl:
Needs research if apps can ask for more permissions. 

Workaround:
1. Open iOS Settings
2. Search for ownCloud iOS app
3. Fix Photos permissions 

![DC68C790-A2EB-447A-BE1B-3262CE92F1B4](https://user-images.githubusercontent.com/214010/102484714-f1a69480-4066-11eb-8e70-41b922d2cc04.jpeg)

jesmrec:
Done! Ready to 11.5

PR #853, opened 2020-12-20
[iOS 14] New photo picker / permissions model

mneuwert:
## Description
- Using new PHPhotoPicker introduced in iOS14 instead of our custom picker.
- Dealing with the photo permission model introduced in iOS14 where user can grant access just to specific photo assets or albums.

## Related Issue
#851 

## Motivation and Context
- New photo picker in iOS14 has got an album tab and allows searching based on people faces, objects, locations etc.
- Since the user can limit the access, we could provide a convenient UI option allowing to raise access level without the need opening the Settings app.

## Types of changes

## Checklist:

## QA

Test plan:

Bugs & improvements:

- [X] (1) Bad UX in `Selected` option https://github.com/owncloud/ios-app/pull/853#issuecomment-769832478 [FIXED]


mneuwert:
@michaelstingl Improved photo selection process on iOS14 using new picker:
- If limited access feature allowing to disable access to certain photos is not active, everything just works.
- If user limits access to certain photos / videos, they can be selected.. And then similar picker used by the system to change permissions is presented allowing the user to revise his choice regarding limited access and to grant access.. (however it is a pity you can't actually highlight stuff which is selected for upload but is missing access permission).
- When user confirms his new limited access selection, then assets are re-fetched and if they are available, then they are uploaded. If none of the assets are accessible, warning alert is shown.

hosy (approved):
Great work!

mneuwert:
@hosy Thanks! Appreciate!

jesmrec:
I have to finish QA here it. First rounds seems pretty good.

Just one comment. Check these steps out:

1. In Device > Settings > ownCloud > Photos, select `Selected Photos`
2. Select some photos there
3. in ownCloud app, select `Upload from Photo Library`
4. Select photos that were not selected in step 2.

at this point, i expected an alert error, letting user know that it is not allowed (lack of permissions). Something similar to the error that happens when `None` is selected in step 1. 

Instead, the selecting view (the same view that appears in step 1) is displayed to allow users to select new items to allow them for the app. This could be a little confusing, but my question is: is that way to do iOS-native? or is something implemented for the feature?

mneuwert:
> I have to finish QA here it. First rounds seems pretty good.
> 
> Just one comment. Check these steps out:
> 
> 1. In Device > Settings > ownCloud > Photos, select `Selected Photos`
> 2. Select some photos there
> 3. in ownCloud app, select `Upload from Photo Library`
> 4. Select photos that were not selected in step 2.
> 
> at this point, i expected an alert error, letting user know that it is not allowed (lack of permissions). Something similar to the error that happens when `None` is selected in step 1.
> 
> Instead, the selecting view (the same view that appears in step 1) is displayed to allow users to select new items to allow them for the app. This could be a little confusing, but my question is: is that way to do iOS-native? or is something implemented for the feature?

- That's in fact confusing since new photo picker shows all pictures irrespective of photo library permissions. 
- So, if you try to fetch photos which are not available (but you have selected them), second almost identical picker is shown allowing to add more permissions. It comes from the system and there is no way to customise it.
- So, if you don't implement special handling, this second picker is appearing only once and if user doesn't change permissions, he is never asked again by the system. I did try to improve this though.. There is a way to suppress it via setting special key in Info.plist and to trigger it manually. But again, you can just tell the system to show it and can't pass any parameters or so... The only improvement I managed to achieve is: it is always consistently shown whenever current permissions don't match the selection of assets to upload.

So there are some tradeoffs when using new photo picker...

michaelstingl:
Before Apple’s dialogue pops up, could we show an own thing, that tells users what to do?

mneuwert:
> Before Apple’s dialogue pops up, could we show an own thing, that tells users what to do?

yes, we could

mneuwert:
> Before Apple’s dialogue pops up, could we show an own thing, that tells users what to do?

Added an alert conveying that access to media is limited, then when user taps " Change" Apple's dialogue is shown, where photos to which access is granted are displayed with a checkmark.

jesmrec:
> Added an alert conveying that access to media is limited, then when user taps " Change" Apple's dialogue is shown, where photos to which access is granted are displayed with a checkmark.

that's much better

jesmrec:
Approved on my side

PR #869, opened 2021-01-12
[feature/tokenauth-uidenforce] Enforce User ID when updating token-based bookmarks

felix-schwarz:
## Description
This PR requires the user ID to remain the same when updating token-based bookmarks. If the user logs in as a user other than the one with which the bookmark was originally created, an error will be presented.

Error trying to log in as different user (edit bookmark) | Regular login error | Error trying to log in as different user (inline editing in "opened" account)
---------------------------|----------------------|--------------------
Bildschirmfoto 2021-01-12 um 23 52 44 | Bildschirmfoto 2021-01-12 um 23 53 10 |  Bildschirmfoto 2021-01-12 um 23 53 23

## Types of changes


hosy:
@felix-schwarz For a while we talked about adding the username as parameter￼ for the authentication web view, so that the username value is automatically set. 
Does it make sense to implement it inside this PR?￼

felix-schwarz:
@hosy Yeah, that could make sense as it is closely related. Last time we tried adding it for OC OAuth2, however, it crashed the web view (and curiously also Safari, when opening the URL there), hinting at a WebKit bug. So I'd skip that for OC OAuth2.

For OIDC, however, it's certainly worth looking into. I'll check the RFC if there is a parameter we can send along.

felix-schwarz:
@hosy Ok, checked and for OIDC the SDK is already passing along the username as `login_hint` ([spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)). However, it is not used by the IDP of the `ocis.owncloud.works` test server.

felix-schwarz:
Re OAuth2 I just re-checked if Apple fixed the WebKit crash in iOS 14.3, but unfortunately, they didn't:
Bildschirmfoto 2021-01-13 um 10 49 45

ownCloud/Client/ClientRootViewController.swift
```diff
@@ -666,6 +629,56 @@ extension ClientRootViewController : OCCoreDelegate {
...
+
+		alertController.addAction(UIAlertAction(title: authFailureIgnoreLabel, style: authFailureIgnoreStyle, handler: { (_) in
+			completionHandler()
+		}))
+
+		if authFailureHasEditOption {
+			alertController.addAction(UIAlertAction(title: "Sign in".localized, style: .default, handler: { [weak self] (_) in
+				completionHandler()
+
+				var notifyAuthDelegate = true
```
mneuwert, line 645: @felix-schwarz reduce nesting with `guard let `statements?

mneuwert (changes requested):

ownCloud/Client/ClientRootViewController.swift
felix-schwarz: Sorry, but I didn't find any opportunity to simplify the logic there with `guard let` statements. Still could reduce nesting by unifying `if`s that were nested unnecessarily.

mneuwert (approved):

jesmrec:
QA checks

- [X] Edit OAuth2 bookmark, logging in with different user -> correct error
- [X] Revoke OAuth2 token, logging in with different user -> correct error
- [X] Edit OIDC bookmark, logging in with different user -> correct error

- [X] Edit OAuth2 bookmark, logging in with same user -> correct login
- [X] Revoke OAuth2 token, logging in with same user -> correct login
- [X] Edit OIDC bookmark, logging in with same user -> correct login

jesmrec:
Approved on my side

jesmrec (approved):

PR #872, opened 2021-01-19
[feature/tls-diff] TLS certificate comparison

felix-schwarz:
## Description
When logging into an account and experiencing a different certificate that does not fulfill the rules for automatic acceptance as replacement, the issue it brings up now shows the differences between the two certificates to allow an informed decision by the user.

Certificate diff | Certificate diff (chain segment) | Diff of next certificate in the chain
--------------|--------------|--------------
Bildschirmfoto 2021-01-19 um 15 25 56 | Bildschirmfoto 2021-01-19 um 15 26 08 | Bildschirmfoto 2021-01-19 um 15 30 38

## Types of changes


mneuwert (approved):

jesmrec:
any way to build a server in which this feature can be tested? 

CC @michaelstingl

michaelstingl:
> any way to build a server in which this feature can be tested?

### Valid cert ==> invalid cert

first connect to real server with SSL (demo.owncloud.org), then connect through `mitmproxy` with hacky-hacky cert. cert diff should show you plenty of differences

### Valid cert ==> other valid cert

Setup server with Traefik and let's encrypt (Cloudflare DNS only), then enable Cloudflare SSL proxy in front of it…
(maybe too mich effort for this tiny feature)

jesmrec:
i will try that. I was trying to avoid much extra effort setting up environments, but there will be room for learning.

felix-schwarz:
@jesmrec If you haven't tested this in an other way, yet: there is a quick guide on how to test this here in the SDK repo: https://github.com/owncloud/ios-sdk/blob/master/doc/TESTING.md#simulating-certificate-changes

The files used in that example are next to it and can be found [here](https://github.com/owncloud/ios-sdk/tree/master/doc/testing-resources/mitmproxy).

jesmrec:
> @jesmrec If you haven't tested this in an other way, yet: there is a quick guide on how to test this here in the SDK repo: https://github.com/owncloud/ios-sdk/blob/master/doc/TESTING.md#simulating-certificate-changes
> 
> The files used in that example are next to it and can be found [here](https://github.com/owncloud/ios-sdk/tree/master/doc/testing-resources/mitmproxy).

Thanks a lot!! that was totally helpful!!

It works fine from my side. Information correctly shown in different themes, orientations, devices and fields.

|||||
|:-:|:-:|:-:|:-:|
|Photo-2021-01-28-13-20-05_1404|Photo-2021-01-28-13-21-09_1405|Photo-2021-01-28-13-21-15_1406|Photo-2021-01-28-13-21-21_1407

just two considerations from my side:

- The option `Show/Hide` works only in case the difference exists. If certificate is shown for the first time or revoked, option `Show/Hide` does nothing. Nothing bad nor blocker.

- Current branch points to an sdk commit that does not build. I've tested agains latest one `3a73b20`. Just for you to take in account when merging.

hosy:
@jesmrec I commit the latest SDK commit into this branch. Not sure if this solves the problem, but you can have a look again. If this issue still exists, we can move forward after adding a new issue for that problem and we can solve this in 11.5.1. But you are right, UI elements without function should be hidden.

jesmrec:
> @jesmrec I commit the latest SDK commit into this branch. Not sure if this solves the problem, but you can have a look again. If this issue still exists, we can move forward after adding a new issue for that problem and we can solve this in 11.5.1. But you are right, UI elements without function should be hidden.

SDK correctly updated. The UX issue is not a blocker, just a minor detail.

approved

jesmrec (approved):

felix-schwarz:
@jesmrec Good catch. I made a change so that the `Show ±` button only appears if there are any differences to be shown.

jesmrec:
> @jesmrec Good catch. I made a change so that the `Show ±` button only appears if there are any differences to be shown.

works perfect! now it's more than approved.

PR #884, opened 2021-01-31
Fixing Media Player Behaviour

mneuwert:
## Description
Fix for an issue when playing multiple items in the same directory. If e.g. image item is the next one, multi media playback would stop.

## Related Issue
n/a

## Motivation and Context
User would like to play all audio-visual files in a row without interruption.

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/Viewer/DisplayHostViewController.swift
+++ b/ownCloud/Client/Viewer/DisplayHostViewController.swift
@@ -19,6 +19,7 @@
 import UIKit
 import ownCloudSDK
 import ownCloudAppShared
+import CoreServices
 
 class DisplayHostViewController: UIPageViewController {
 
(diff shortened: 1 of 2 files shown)


mneuwert:
@michaelstingl I ensured that all media files are played continuously.. Everything else in current directory (pictures, documents) is filtered out. Don't think we should try to do special handling for pictures in scope of this change especially considering that playing media also does work when the app is in background (I think we discussed that may be we should sort of have a gallery mode in which we skip to the next item when picture was displayed for couple of seconds.. guess this would need significantly more work given current preview architecture).

michaelstingl:
@mneuwert okay, perfect 👍

Could be future option: Include images is auto-play.

mneuwert:
> @mneuwert okay, perfect 👍
> 
> Could be future option: Include images is auto-play.

May be also separate it? Like if you open picture, then you just have pictures changed one after another? Not mixing pictures and audio / video files?

ownCloud/Client/Viewer/DisplayHostViewController.swift
```diff
@@ -19,6 +19,19 @@
...
+	var isPlayable: Bool {
+		guard let mime = self.mimeType else { return false }
+
+		guard let uti = UTTypeCreatePreferredIdentifierForTag(kUTTagClassMIMEType, mime as CFString, nil) else {
+			return false
+		}
+
+		return UTTypeConformsTo(uti.takeUnretainedValue(), kUTTypeAudiovisualContent)
+	}
+}
```
hosy, line 34: Nice new addition to the `OCItem` class.
We already have a `OCItem+Extension`. I think it makes sense to move this code into this class, because it is more discoverable there.

hosy (changes requested):

ownCloud/Client/Viewer/DisplayHostViewController.swift
mneuwert: Not sure, this is context specific extension.. so for me it was easier to have it closer to the context in which it is used

hosy: I think this extension is also interesting for an upcoming shortcut extension. For that I the `OCItem+Extension` is a better place in my opinion.

mneuwert: Moved to the extension you mentioned.

hosy (approved):

jesmrec:
Now, media files played in a row (audio & video). Other files, filtered.

Approved

Issue #704, opened 2020-05-17
[FEATURE REQUEST] Prevent sleeping screen, in detail view


hosy:
**Is your feature request related to a problem? Please describe.**
When a file is open in detail view it often makes sense to prevent the screen to go in sleep mode.
Example:
I have a PDF file open, and I want to show a customer the file but the screen should not go in sleep mode, because I cannot touch the screen.

**Describe the solution you'd like**
Option in settings or always on, when file is in detail view
### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

michaelstingl:
> Option in settings

Yes, preferred 👍

jesmrec:
Done.

PR #893, opened 2021-02-13
[feature/prevent-sleep-detail-view] Presentation Mode

hosy:
## Description
Added an action in detail view menu which enables presentation mode.
Presentation mode prevents the display from sleep mode as long as the detail view is dismissed. Furthermore the navigation bar will be hidden and the file opens in full screen.

## Related Issue
#704 

## Motivation and Context
This could be very useful when a user wants to present a file or on a talk.

## How Has This Been Tested?
- open item in detail view
- open the more menu in the navigation bar
- select `Presentation Mode` and `Enable`
- be sure the device will not go into sleep mode (needs to be tested on a real device)

| Simulator Screen Shot - iPhone 11 Pro Max - 2021-02-12 at 17 43 36 | Simulator Screen Shot - iPhone 11 Pro Max - 2021-02-12 at 17 43 47 |
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|

## Types of changes

## Checklist:

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.6/Presentation%20mode.md

Bugs & improvements:

- [x] (1) Presentation mode in non-supported files https://github.com/owncloud/ios-app/pull/893#issuecomment-791405724
- [x] (2) txt & MSO can't escape from fill screen mode https://github.com/owncloud/ios-app/pull/893#issuecomment-791413512


changelog/unreleased/704
```diff
@@ -0,0 +1,6 @@
+Change: Presentation Mode
+
+Added an action in detail view menu which enables presentation mode.
+Presentation mode prevents the display from sleep mode as long as the detail view is dismissed. Furthermore the navigation bar will be hidden.
```
michaelstingl, line 4:
```suggestion
Presentation mode prevents the display from sleep mode as long as the detail view is closed. Furthermore the navigation bar will be hidden.
```

Not sure if "dismissed" is the common term?

ownCloud/Client/Actions/Actions+Extensions/PresentationModeAction.swift
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	// MARK: - Action implementation
+	override func run() {
+		guard context.items.count > 0, let hostViewController = context.viewController else {
+			completed(with: NSError(ocError: .insufficientParameters))
+			return
+		}
+
+		let alertController = UIAlertController(title: "Presentation Mode".localized, message: "Enabling presentation mode will prevent the display from sleep mode until the view is dismissed.".localized, preferredStyle: .alert)
```
michaelstingl, line 47:
```suggestion
		let alertController = UIAlertController(title: "Presentation Mode".localized, message: "Enabling presentation mode will prevent the display from sleep mode until the view is closed.".localized, preferredStyle: .alert)
```

See comment in `changelog/unreleased/704`

michaelstingl (commented):
unsure about the usage of "dismissed" vs. "closed"

hosy:
`Close` also works for me!

ownCloud/Client/Actions/Actions+Extensions/PresentationModeAction.swift
hosy: Fixed wording!

felix-schwarz (changes requested):
Code-wise I noticed that `DisplaySleepPreventer`'s existing method set was not covering a use case like Presentation Mode too well. F.ex. if the OC app is opened in two windows on an iPad, side-by-side, and one half shows the Status tab (which also prevents sleep), then the Presentation Mode action would not prompt the user - but leaving the viewer would end sleep prevention; even if Presentation Mode wasn't used.

Since I wrote `DisplaySleepPreventer` I took the liberty to extend it a bit, so that it supports "reasons" for preventing sleep, allowing Presentation Mode to track its status independently.

During testing, I found another issue: if I present `Example.odt` (or any other file for which there is no viewer support) and then select Presentation Mode, I end up with no way to return:
Simulator Screen Shot - iPhone 12 mini - 2021-02-24 at 00 33 15

hosy:
@felix-schwarz just pushed a new commit to prevent fullscreen support on unsupported subclasses. Please give it a new try. Thanks!

felix-schwarz:
@hosy Thanks. That solves the issue nicely.

felix-schwarz (approved):

jesmrec:
### (1)

About your last finding (open Presentation mode for non-supported files)

Would it make sense to remove the option from there? in case there is no file to display, presentation mode is not needed

jesmrec:
### (2)

1. Open a txt file or a MS Office file (doc, xls, ppt...)
2. Open Menu and select `Presentation Mode`

Current: Preview turns to full screen and there is no way to dismiss it. Other kind of files like PDF or pictures work fine (full screen is dismissed with a tap or a double tap).

Expected: txt and MSOffice files can prevent the full screen mode

iPhoneXR v14.4
`e1b6fb9`

hosy:
@jesmrec regarding (1)
You mean to do only show the menu item, if the file is locally available? 
Could make sense!

hosy:
@jesmrec regarding (2)
It would make sense to show this file types in fullscreen too. I will check if I can add a tap handler to exit the fullscreen mode.

jesmrec:
> @jesmrec regarding (1)
> You mean to do only show the menu item, if the file is locally available?
> Could make sense!

yes, and also, if the file is not previewable (f. ex., a zip file), it will be never presented so it is not needed to be shown.

jesmrec:
(1) and (2) fixed

2nd round finished.

Approved on my side

jesmrec (approved):

Issue #916, opened 2021-03-04
[FEATURE REQUEST] Japanese(2-Byte character) input support


Bables55:
**Is your feature request related to a problem? Please describe.**
When users take a picture with the "Scan Document" function and enter the file name, there is a problem with entering Japanese (2-byte characters).
I've written about it in my forum, including the current issues, so please take a look there as well.
https://central.owncloud.org/t/support-for-2-byte-characters-in-scan-document/29856

**Describe the solution you'd like**
I want to use Japanese (2-byte characters) in file names without any problems.

**Describe alternatives you've considered**
The only way to solve this problem is to use 1-byte characters.

**Additional context**
1: https://od.lk/s/M18yNDA3ODk3NDdf/oc-flickjapanese.MP4 2
2: https://od.lk/s/M18yNDA3ODg4ODNf/oc-qwertyjapanese.mp4

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

hosy:
@Bables55 thank you for reporting this bug. 
The problem is fixed now and will be available with next version 11.6.

jesmrec:
Done

PR #929, opened 2021-03-12
[fix/japanese-input-support] Japanese Input Support

hosy:
## Description
Fixed a problem in scan view when renaming the file name and using a Japanese keyboard layout (2-Byte character).
After entering a character inside the file name the text cursor jumped to the end.

## Related Issue
#916 

## Motivation and Context
Japanese keyboard support for all text input

## How Has This Been Tested?
- only testable on a real device
- add Japanese keyboard layout to the iOS keyboards
- Scan a document 
- Rename the file name inside the scan view
- please check if the following text fields working as expected: Public Link > Edit Link Name, Bookmark > Edit Bookmark Name

## Types of changes

## Checklist:

### DIFF
--- a/ownCloudAppShared/User Interface/StaticTableView/StaticTableViewRow.swift
+++ b/ownCloudAppShared/User Interface/StaticTableView/StaticTableViewRow.swift
@@ -444,10 +444,6 @@ open class StaticTableViewRow : NSObject, UITextFieldDelegate {
 			cellTextField.bottomAnchor.constraint(equalTo: (cell?.contentView.bottomAnchor)!, constant:-14).isActive = true
 		}
 
-		self.updateViewFromValue = { [weak cellTextField] (row) in
-			cellTextField?.text = row.value as? String
-		}
-
 		self.updateViewAppearance = { [weak cellTextField] (row) in
 			cellTextField?.isEnabled = row.enabled
 			cellTextField?.textColor = row.enabled ? Theme.shared.activeCollection.tableRowColors.labelColor : Theme.shared.activeCollection.tableRowColors.secondaryLabelColor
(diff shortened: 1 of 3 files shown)


ownCloudAppShared/User Interface/StaticTableView/StaticTableViewRow.swift
```diff
@@ -444,10 +444,6 @@ open class StaticTableViewRow : NSObject, UITextFieldDelegate {
 			cellTextField.bottomAnchor.constraint(equalTo: (cell?.contentView.bottomAnchor)!, constant:-14).isActive = true
 		}
 
-		self.updateViewFromValue = { [weak cellTextField] (row) in
```
felix-schwarz, line 447: This would remove the ability to update the contents of text fields by setting the `value` property of `StaticTableViewRow`s.

The problem seems to be that:
- entering text into the textfield calls `textFieldContentChanged()`
- `textFieldContentChanged()` sets `.value` to the entered text
- setting `value` invokes `.updateViewFromValue()`, which sets the textfield's string to the entered text, which seems to cause the issue

By making sure that `.updateViewFromValue()` is not called when `.value` is updated via `textFieldContentChanged()`, this should be possible to avoid. I've quickly put together a proposal on how that could be achieved in branch `fix/japanese-input-support-alternative` in commit https://github.com/owncloud/ios-app/commit/34176217c8b6426aaed96dd99b29a2443157ba9b (can be merged directly into this PR if you find it to work (I didn't test it) and want to use it).

felix-schwarz (changes requested):

ownCloudAppShared/User Interface/StaticTableView/StaticTableViewRow.swift
hosy: @felix-schwarz yes, you are right. Just used your alternative. Thanks for your proposal!

hosy:
@felix-schwarz thank you for your proposal. I merged it inside and you can add your CR again.

felix-schwarz (approved):

jesmrec:
Tested with the japanese keyboard and scan document feature... now all characters are written in a row, without jumping. 

From my side is approved, but the real approval for this should be given by the japanese user.

jesmrec (approved):

PR #545, opened 2019-10-24
[feature/globalsearch] Global search using custom queries

felix-schwarz:
## Description
- POC implementation of global search using custom queries

## Related issues
#53

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Screenshots
Folder | Global / All
-- | --
Bildschirmfoto 2019-10-24 um 22 05 59 | Bildschirmfoto 2019-10-24 um 22 06 01

Issue #53, opened 2018-05-18
Search


michaelstingl:
Related:
- https://github.com/owncloud/ios/issues/17

Needs server API:
- https://github.com/owncloud/core/issues/25373
- https://github.com/owncloud/core/issues/13915

michaelstingl:
Search API has been implemented on oC server and is planned to be included in oC 10.0.10: https://github.com/owncloud/core/issues/25373#issuecomment-400967608

felix-schwarz:
Implemented in #1401.

PR #933, opened 2021-03-17
[feature/localsearch] Local account-wide search using custom queries

felix-schwarz:
## Description
Local account-wide search using custom queries. An arrow is added to the right of the search results to jump directly to the folder containing the results, scrolls the item into view and highlights it.

Local search segments the input into terms and filter keywords. Some examples:

Term | Meaning
------|---------
`engineering` | find items with `engineering` in their name
`engineering demos` | find items with `engineering` and `demos` in their name
`"engineering demos"` | find items with `engineering demos` in their name
`"engineering "demos` | find items with `engineering` and `demos` in their name
`type:pdf` | limit search to items with a pdf suffix
`type:pdf,mov` | limit search to items with a pdf or mov suffix
`before:2021` | limit search to last modified before 2021-01-01
`before:2021-02` | limit search to last modified before 2021-02-01
`before:2021-02-03` | limit search to last modified before 2021-02-03
`after:2020` | limit search to last modified after 2020-01-01
`after:2020-02` | limit search to last modified after 2020-02-01
`after:2020-02-03` | limit search to last modified after 2020-02-03
`on:2020-02-03` | limit search to last modified on 2020-02-03 (0h-24h)
`on:2020-02-03,2020-02-05` | limit search to last modified on 2020-02-03 (0-24h) or 2020-02-05 (0-24h)
`smaller:200mb` | limit search to items smaller than 200 MB (supports tb / tib / gb / gib / mb / mib / b)
`greater:1gb` | limit search to items greater than 1 GB (supports tb / tib / gb / gib / mb / mib / b)
`owner:cscherm` | limit search to items owned by user cscherm
`:file` | limit search to files
`:folder` | limit search to folders
`:image` | limit search to images (MIME type starting with `image/`)
`:video` | limit search to videos (MIME type starting with `video/`)
`:year` | limit search to items last modified since Jan 1st of the current year
`:month` | limit search to items last modified since the current months's first day
`:week` | limit search to items last modified since beginning of the current week's Monday 
`:today` or `:d` | limit search to items last modified on the current day
`:5d` | limit search to items last modified  in the last 5 days
`:2w` | limit search to items last modified in the last 2 weeks
`:1m` | limit search to items last modified this or last month
`:m` | limit search to items last modified this month
`:1y` | limit search to items last modified this or last year
`:y` | limit search to items last modified this year
`-[term]` | negates the search term, i.e. `-:image` to exclude images, or `-wrd` to exclude all matches containing `wrd`. This works with any of the examples above.

Other, related changes in this pull request:
- database migrations with inline progress reporting
- improved scene restoration (now also covering back to server list)

Other, unrelated changes in this pull request:
- check if `displayName` and `userName` have at least one character before using them to build `OCBookmark.shortName` for displaying in the server list
- more differentiated status 503 handling as described in https://github.com/owncloud/enterprise/issues/4476
- sync status scrubbing: removes sync status from items with invalid sync status (from previous versions), avoiding items whose sync actions are fully processed from continuing to "spin" in the list

## To Do
- [x] ability to show more than 100 matches
- [x] reveal action to jump to the enclosing folder of a match
- [x] ~a database migration screen at app launch (migrating all accounts' databases one by one)~ -> #938
- [x] search remains active after choosing an item and returning from it

## Related Issues
- #53, replaces #545 
- https://github.com/owncloud/product/issues/42

Database upgrade / migration | Search with Reveal buttons
----------|-------------------
Bildschirmfoto 2021-03-22 um 22 15 31 | Bildschirmfoto 2021-03-24 um 23 53 25

Show more results (censored) | Revealed/highlighted item 
-------------|--------------
Bildschirmfoto 2021-03-25 um 10 37 36 | Bildschirmfoto 2021-03-25 um 10 34 47

## Types of changes

## QA

Test Plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.6/Local%20search.md

Reports:

- [x] (1) `type` not working https://github.com/owncloud/ios-app/pull/933#issuecomment-820159829 [FIXED]
- [x] (2) Term with two strings starting with quotation marks does not work https://github.com/owncloud/ios-app/pull/933#issuecomment-820162903 [FIXED]
- [ ] (3) Sorting bar on the left (landscape)  https://github.com/owncloud/ios-app/pull/933#issuecomment-820165553
- [x] (4) Searching by `:day` https://github.com/owncloud/ios-app/pull/933#issuecomment-820198920 [FIXED]
- [x] (5) Guide users to all the searching options https://github.com/owncloud/ios-app/pull/933#issuecomment-820209293  [WONT DO HERE]
- [x] (6) File names matching criteria names https://github.com/owncloud/ios-app/pull/933#issuecomment-820245197 [FIXED]
- [x] (7) Saving in directory picker via search https://github.com/owncloud/ios-app/pull/933#issuecomment-820988961 [FIXED]
- [x] (8) Keyboard commands for "Search" are not available in directory picker https://github.com/owncloud/ios-app/pull/933#issuecomment-820992634  [FIXED]
- [X] (9) Multiselection in folder picker https://github.com/owncloud/ios-app/pull/933#issuecomment-821093095 [FIXED]


hosy:
@felix-schwarz  at first: thank you for implementing this helpful feature!
Could it make sense, if you add a help view, which is visible the first time the new search is used and show an explanation which keywords are available and how to use it?
Maybe it makes sense to have a generic class to show help content and remember, when it was shown. 
Furthermore all available help topics should be available in the settings.

What do you think?

felix-schwarz:
@hosy Thanks for the ideas and feedback. I like them a lot! 

My plan is to implement search in two steps:
- Step 1: pretty much this PR, to have search available quickly while also allowing to evaluate which keyword "power features" might be useful and needed.
- Step 2: refactoring for simplicity, search, inline actions, suggestions and grid view:
    - introduce a new `ClientItemSource` class that provides a list of entries, with subclasses that
        - house just a list of `OCItems`
        - feed themselves from `OCQuery`
        - compose their own items from other `ClientItemSource`s - for search maybe even leveraging `OCExtension` to create an extensible infrastructure
        - return `ClientItem`s (not sure if a class or just a category), which can represent
            - classic `OCItem`s
            - (new) actions (like the "Show more" entry in this PR)
            - (new) suggestions or search history (like in your suggestions) 
    - refactor the `ClientQueryViewController` and parent classes back into a single, "simple" controller that takes its items from a new `ClientItemSource` object
    - a grid view implementation/replacement for `ClientQueryViewController` would also benefit from the simplicity of only having to handle a `ClientItemSource`

I think showing help is a good idea to make these features accessible, but I'd move that to step 2, where it's easier to determine what's still needed after f.ex. adding suggestions. One of the "suggestions" could, btw, also be a "link" that opens an "advanced search" UI, which provides a UI for the supported keywords, composing the keywords for them in the end.

ownCloudAppFramework/SDK Extensions/OCQueryCondition+SearchSegmenter.m
```diff
@@ -0,0 +1,304 @@
...
+	else if ([segmentStringLowercase isEqual:@":week"])
+	{
+		return ([OCQueryCondition where:OCItemPropertyNameLastModified isGreaterThan:[NSDate startOfRelativeWeek:0]]);
+	}
+	else if ([segmentStringLowercase isEqual:@":month"])
+	{
+		return ([OCQueryCondition where:OCItemPropertyNameLastModified isGreaterThan:[NSDate startOfRelativeMonth:0]]);
+	}
+	else if ([segmentStringLowercase isEqual:@":year"])
+	{
```
hosy, line 134: The keywords below supports localized strings. Shouldn't this available for the strings `folder`, `file`, `image`, `video`, `today`, `week`, `month`, `year` too?

ownCloudAppFramework/SDK Extensions/OCQueryCondition+SearchSegmenter.m
```diff
@@ -0,0 +1,304 @@
...
+					else if ([modifier isEqual:@"smaller"] || [modifier isEqual:OCLocalizedKeyword(@"smaller")])
+					{
+						NSNumber *byteCount = [parameter byteCountNumber];
+
+						if (byteCount != nil)
+						{
+							condition = [OCQueryCondition where:OCItemPropertyNameSize isLessThan:byteCount];
+						}
+					}
+					else if ([modifier isEqual:@"greater"] || [modifier isEqual:OCLocalizedKeyword(@"greater")])
```
hosy, line 201: String missing `Localizable.strings` file

ownCloudAppFramework/SDK Extensions/OCQueryCondition+SearchSegmenter.m
```diff
@@ -0,0 +1,304 @@
...
+							onStartDate = [onStartDate dateByAddingTimeInterval:-1];
+							onEndDate = [onStartDate dateByAddingTimeInterval:60*60*24+2];
+
+							condition = [OCQueryCondition require:@[
+								[OCQueryCondition where:OCItemPropertyNameLastModified isGreaterThan:onStartDate],
+								[OCQueryCondition where:OCItemPropertyNameLastModified isLessThan:onEndDate]
+							]];
+						}
+					}
+					else if ([modifier isEqual:@"smaller"] || [modifier isEqual:OCLocalizedKeyword(@"smaller")])
```
hosy, line 192: String missing `Localizable.strings` file

ownCloudAppShared/Client/File Lists/ClientQueryViewController.swift
```diff
@@ -136,6 +254,20 @@ open class ClientQueryViewController: QueryFileListTableViewController, UIDropIn
...
+ 	}
+//
+//	open override func startQuery(_ theQuery: OCQuery) {
+//		if theQuery == customSearchQuery {
+//			self.updateCustomSearchQuery()
+//		} else {
+//			super.startQuery(theQuery)
+//		}
+//	}
+
```
hosy, line 270: Is this comment needed or can it be removed?

ownCloudAppShared/Client/File Lists/QueryFileListTableViewController.swift
```diff
@@ -72,6 +75,7 @@ open class QueryFileListTableViewController: FileListTableViewController, SortBa
 
 	public init(core inCore: OCCore, query inQuery: OCQuery) {
 		query = inQuery
+		searchScope = .global
```
hosy, line 78: shouldn't this var set to `.local`. I found `.local` as default below as initial variable value and in an other source file as default value?

ownCloudAppShared/Client/User Interface/SortBar.swift
```diff
@@ -35,13 +35,45 @@ public class SegmentedControl: UISegmentedControl {
...
+public enum SearchScope : Int, CaseIterable {
+	case global
+	case local
+
+	var label : String {
+		var name : String!
+
+		switch self {
+			case .global: name = "all".localized
+			case .local: name = "folder".localized
```
hosy, line 47: Strings missing in `Localizable.strings`

ownCloudAppShared/Client/User Interface/SortBar.swift
```diff
@@ -56,7 +88,7 @@ public class SortBar: UIView, Themeable, UIPopoverPresentationControllerDelegate
 	}
 
 	// MARK: - Constants
-	let sideButtonsSize: CGSize = CGSize(width: 44.0, height: 44.0)
+	let sideButtonsSize: CGSize = CGSize(width: 22.0, height: 22.0)
```
hosy, line 91: I thought we set this to 44.0 before, because it was not so good to tap, because of the small size before. Please also have a look, if it is aligned with the `…` of the more button in the file list cell. Before the UI looked a little bit "jumpy".

ownCloudAppShared/Client/User Interface/SortBar.swift
```diff
@@ -93,57 +139,82 @@ public class SortBar: UIView, Themeable, UIPopoverPresentationControllerDelegate
...
 			selectButton.translatesAutoresizingMaskIntoConstraints = false
+			searchScopeSegmentedControl.translatesAutoresizingMaskIntoConstraints = false
 
 			sortButton.accessibilityIdentifier = "sort-bar.sortButton"
 			sortSegmentedControl.accessibilityIdentifier = "sort-bar.segmentedControl"
+			searchScopeSegmentedControl.accessibilityIdentifier = "sort-bar.searchScopeSegmentedControl"
+
+			for scope in SearchScope.allCases {
+				if let image = scope.image {
+					searchScopeSegmentedControl.insertSegment(with: image, at: scope.rawValue, animated: false)
```
hosy, line 189: Does it make more sense to only show the titles? Your chosen images are great, but it looks a little bit to large for the small segmented control UI element. But needs to check, if the width fits for all UI elements in this row on a small iPhone display.
Can you try a build with titles only please?

hosy (changes requested):

ownCloudAppFramework/SDK Extensions/OCQueryCondition+SearchSegmenter.m
felix-schwarz: I missed those. The latest commit allows localizing these, too, but now uses a normalization + localization step to both simplify the code and prepare the next steps where queries could always be saved in English to allow a common syntax across client and server.

ownCloudAppFramework/SDK Extensions/OCQueryCondition+SearchSegmenter.m
felix-schwarz: Fixed with the latest commit.

ownCloudAppShared/Client/File Lists/ClientQueryViewController.swift
felix-schwarz: Thanks, that's a leftover from a previous refactoring. Removed it with the latest commits.

ownCloudAppShared/Client/File Lists/QueryFileListTableViewController.swift
felix-schwarz: That line is indeed not needed and even could overwrite the last user setting. I removed it. Thanks!

ownCloudAppShared/Client/User Interface/SortBar.swift
felix-schwarz: Fixed with the latest commits.

ownCloudAppShared/Client/User Interface/SortBar.swift
felix-schwarz: I did initially try with titles, but ran into space problems and had issues finding a short word that explains the purpose as well as the icons. "all" fits, but "everywhere" or "all folders" would be better. Same with "folder" vs. "this folder". But then the UI no longer fits.

If you feel the icons are too large, we could also try slightly scaled down versions. What do you think?

ownCloudAppShared/Client/User Interface/SortBar.swift
felix-schwarz: That was also my worry initially, but it didn't seem to be an issue in tests when I tried. I chose a smaller size so that the "Reveal" arrow also fits in without excessive white space between the "more dots" and the "reveal" arrow.

In my testing, the dots still aligned well between navigation bar and table view rows:

More | More + Reveal
--|--
Bildschirmfoto 2021-04-08 um 22 59 56 | Bildschirmfoto 2021-04-08 um 23 00 26

hosy: The problem is on a real device. On simulator everything works fine, but on the iPhone it is difficult to touch.

ownCloudAppShared/Client/User Interface/SortBar.swift
hosy: I just tested the label solution on a small screen size (iPhone 6) and it fits. I also discovered, that the `Multiselect` button needs to be hidden, when search is active, because the UI elements are hidden in this mode. This also give us more space for the search scope segmented control.
Maybe it makes sense to use the label `Account` instead of `All` for the global search scope.

If you agree, I can push my changes, but I'm also interested in your thoughts.
Here you can see the result:

Simulator Screen Shot - iPhone 6s - 2021-04-09 at 11 12 55

ownCloudAppShared/Client/User Interface/ClientItemCell.swift
```diff
@@ -36,7 +37,8 @@ open class ClientItemCell: ThemeTableViewCell, ItemContainer {
 	private let smallSpacing : CGFloat = 2
 	private let iconViewWidth : CGFloat = 40
 	private let detailIconViewHeight : CGFloat = 15
-	private let moreButtonWidth : CGFloat = 60
+	private let moreButtonWidth : CGFloat = 45
```
hosy, line 40: I figured out, that changing this button size, will lead in the right aligned button will no longer aligned in a row.
I would change this back and set 
```
private let moreButtonWidth : CGFloat = 60


moreButtonWidthConstraint?.constant = showRevealButton ? revealButtonWidth : moreButtonWidth
```

I can commit this change, because I already tested it.

hosy:
@felix-schwarz I pushed my code review suggestions into feature/localsearch-cr-suggestion

ownCloudAppShared/Client/User Interface/SortBar.swift
felix-schwarz: "Account" and "Folder" are good choices for labels IMO.

My worry, however, is with iOS 12 on the smallest supported devices (iPhone 5s). 

Even in English, it's already getting really tight, but in German there are overlaps:

English | German
--------|--------
Simulator Screen Shot - iPhone 5s - 2021-04-09 at 14 58 17 | Simulator Screen Shot - iPhone 5s - 2021-04-09 at 15 04 08

felix-schwarz: I also just noticed that we don't change the placeholder text when switching between account and folder scope (always "Search this folder"). Will fix that.

ownCloudAppShared/Client/User Interface/ClientItemCell.swift
felix-schwarz: I just merged your suggestion branch. Thanks!

ownCloudAppShared/Client/User Interface/SortBar.swift
felix-schwarz: On the other hand, we can't use the SF Symbols on iOS 12, so would either need to provide our own icons - or drop back to text there. Taking into account that iOS 12 + iPhone 5s should be a really small cohort, the only issue for these is purely cosmetic in nature, and that text communicates the purpose a lot clearer, I came to the conclusion that we should indeed use text here and merged the changes from your branch.

Also fixed the placeholder text.

hosy (approved):

jesmrec:
### (1)

`type:` is not working in Folder scope

- In a folder with pdf files (or any other type of file), enter `type:pdf` using `Folder` scope -> no results ❌ 
- In a folder with pdf files (or any other type of file), enter `type:pdf` using `Account` scope -> correct results ✅ 

The corresponding negative search does not work either

iPhoneXR V14.4
`e09822ec`

jesmrec:
### (2)

The following term:

Term | Meaning
------|---------
`"engineering "demos` | find items with `engineering` and `demos` in their name

does return nothing. is there any typo? (missing quotation marks or similar).

`copy 280.txt`
Search: `"copy "280`
Result: empty

iPhoneXR V14.4
`e09822ec`

jesmrec:
### (3)

I've realised that the sorting bar has been misplaced from the center to the left side of the screen, when the device is in landscape orientation

Screen Shot 2021-04-15 at 08 49 53

is this intended?

iPhoneXR V14.4
`e09822ec`

jesmrec:
### (4)

`day:` not working

1. Modify a file (that means, set todays' date) or create a new folder in the current folder
2. Search with `:day` 
3. Search with `-:day`

Current: 

2. no results
3. the complete list

Expected:

2. Modified file and new folder as results
3. Everything but the modified file and the new folder

iPhoneXR V14.4
`e09822ec`


NOTE: i took as reference the chart in the top message. Searching the items by today's date works with `:d` , that is not specified there. So, this is probably a typo in the chart, but just in case something is bad in the code...

jesmrec:
### (5) (improvement)

The amount of searching criteria covers almost everything that a user can imagine. But, as a user, i see the top bar with the "Search account" tip and i don't imagine that i can do a `-after:2020-08-01` (f.ex.).

We should find a way to point users to any guide with all the existing options, otherwise, this will be a hidden feature.

Just a crazy idea: next to the lens icon in the search bar, set a '?' or something similar. By clicking, a new view containing all the options in the top message.

What do you think?

jesmrec:
### (6)

This is just a comment or clarification

Files with names that matches the searching criteria weren't taken in account for the testing, as corner cases which fix probably does not make sense. File names like:

`:week`
`-after:2020`
`:month`
`-`

are not matching, so, they are not displayed in case of searching even though they are valid file names. 

If you think that make sense to fix those corner cases, it is not a problem for me and will be added to the test plan. If not, they will not block the approval process.

michaelstingl:
> otherwise, this will be a hidden feature

okay for the initial release. we'll work with @tbsbdr and the web team on UX after the release

jesmrec:
> okay for the initial release. we'll work with @tbsbdr and the web team on UX after the release

ok, i will mark as future feature

michaelstingl:
@felix-schwarz what about share status, favourites, and tags (future)? BTW I like the [GH search syntax](https://docs.github.com/en/github/searching-for-information-on-github/searching-issues-and-pull-requests).

felix-schwarz:
@jesmrec 
**Re (1):** Thanks! That was a bug. I fixed it with https://github.com/owncloud/ios-app/pull/933/commits/372d798a35c0560c8e12b98caa2db5c744247056

**Re (2):** It worked here™. I noticed, however, that instead of a plain `"` I get a localized `„` in the Simulator now - and then it doesn't work. Maybe it is the same issue for you. I added support for all variants of quotation marks I could find in the macOS letter overview as of https://github.com/owncloud/ios-app/pull/933/commits/38ad27d1dddbd27ac38e9dd52f78fa8b00460902

**Re (3):** Another bug, thanks! Fixed as of https://github.com/owncloud/ios-app/pull/933/commits/6ae824201b22e9c8f9efcca95759b7005a0d14f9

**Re (4):** I think the original wording was indeed `:day` but I changed it to `:today` for clarity and forgot to update the table. Sorry!

**Re (5):** Making this more discoverable and maybe even adding a rich UI is targeted by step 2 for search, described here: https://github.com/owncloud/ios-app/pull/933#issuecomment-810938799

**Re (6):** Good point! I made changes so that quoting a term disables keywords, i.e. if you want to find a file with `:today` in it, you're now able to type `":today"` and it will be found. Likewise `-":today"` also works. Changed in https://github.com/owncloud/ios-app/pull/933/commits/2e67485ef35bf1bb4cfe299bb3a6d5076d9675bd

felix-schwarz:
> @felix-schwarz what about share status, favourites, and tags (future)? BTW I like the [GH search syntax](https://docs.github.com/en/github/searching-for-information-on-github/searching-issues-and-pull-requests).

Structurally it already matches GitHub search (except for the "super short shortcuts" with a leading `:`). Are there any keywords that you'd like to see renamed or added?

Share status should be relatively easy to add. Favorites and tags are more of a challenge to keep up-to-date as long as metadata changes don't propagate. But I'd like to eventually have that, too.

I was also thinking about searching for users (owners, …). What do you think?

michaelstingl:
> I was also thinking about searching for users (owners, …). What do you think?

Yeah, I like owners 👍

hosy:
## (7) Saving in directory picker via search

- Execute Action `Move` on a file item
- Directory Picker will be opened
- Search in directory picker for a folder
- In the search result list tap on the `Right Arrow` to directly jump into the folder
- `Save Here` button is hidden in the toolbar and folder cannot be selected as save destination

hosy:
## (8) Keyboard commands for "Search" are not available in directory picker

### On iPad (with external keyboard connected):
- Execute Action `Move` on a file item
- Directory Picker will be opened
- Long press `Command` key
- No keyboard commands available for: `Enable Search`, `Toggle Search Scope`, `Change Sort Type/Order`

hosy:
(8) is fixed with commit https://github.com/owncloud/ios-app/pull/933/commits/002ee67fd10cd4c91b4faf700dc3765e953a8167

felix-schwarz:
# (9) Multiselect possible after search in directory picker

When searching in the directory picker (f.ex. during a Move action) and exiting from search, the multi select button appears. Fixed in upcoming push.

felix-schwarz:
@hosy @jesmrec  **(7)** and **(9)** are addressed by https://github.com/owncloud/ios-app/pull/933/commits/4fddedfdebd13fe48d8b1b48d9d00682c575e512

jesmrec:
Thanks for your help with the QA ;)

i keep checking

jesmrec:
(1), (2), (4), (6), (7), (9) -> fixed

(3) -> not fixed. Sorting options are still alligned to the left. 

(5) -> won't fix here


1st QA rond complete, will do a second round with iPad (including the (8) check)

felix-schwarz:
> > I was also thinking about searching for users (owners, …). What do you think?
> 
> Yeah, I like owners 👍

As of https://github.com/owncloud/ios-app/pull/933/commits/399ed6d5ed41a882b529cb79b507755a5722e0bc, `owner:[userID]` is now supported.

felix-schwarz:
@jesmrec Re (3): please share a screenshot and mark what part exactly is misplaced. I thought it'd be about the "name | kind | size | …" part of the Sort Bar.

jesmrec:
about (3): the sorting bar `name | kind | size | date | shared` is moved to the left side of the screen in landscape mode, if we compare it with the base branch (`milestone/11.6`)

Then, not a bad behaviour or a blocker, but a difference that can be intended or not (this is my question)

milestone/11.6 | current
------|---------
Screen Shot 2021-04-19 at 08 41 41|Screen Shot 2021-04-19 at 08 39 26

`399ed6d`

jesmrec:
Checking (8). New shortcuts are correct, but i miss one: `Change Sort Type/Order` is now available in the folder picker but it's missing in the file list. @hosy, kind of regression?

jesmrec:
OTOH, the maintenance error is not prevented correctly when an external storage is not accesible (showing cached content)

hosy:
@jesmrec @felix-schwarz  regarding (3):
for me the left aligned sort bar does also work, and maybe does make more sense, because it is also left aligned on the iPhone

regarding (8):
Fixed duplicated keyboard commands:
changed keyboard command for toggle search order and share item

jesmrec:
>regarding (8):
>Fixed duplicated keyboard commands:
>changed keyboard command for toggle search order and share item

fixed!!

>@jesmrec @felix-schwarz regarding (3):
>for me the left aligned sort bar does also work, and maybe does make more sense, because it is also left aligned on the iPhone

it is not a problem for me to keep the bar in the left-side. In terms of quality, it does not mean any loss. This is more UX-issue than quality issue.

hosy:
@jesmrec I think we should try the left aligned version. For me it fits into the current UI design. Most UI elements are left aligned and right aligned.

jesmrec:
Ok, let's give a try. As i commented, i only reported this because is a non-reported difference with the base branch.

Then, QA is finished (test plan in first message) and no reports are pending

jesmrec (approved):

PR #941, opened 2021-04-01
[fix/enable-markup-editing] Improved Enabling Markup Mode

hosy:
## Description
In some cases enabling markup mode failed. Now implemented a solution, which checks periodically, if the markup button is available, before enabling. Furthermore implemented a check, if item could be previewed in markup action and shows an error, if not possible.

## Related Issue
https://github.com/owncloud/enterprise/issues/4467
https://github.com/owncloud/enterprise/issues/4468

## Motivation and Context
Enable markup mode reliable

## How Has This Been Tested?
Open a PDF or image file in markup mode.

Please test on iPad and iPhone, iOS 14 and iOS 13

## Types of changes

## Checklist:

### DIFF
--- a/ownCloud/Client/Actions/EditDocumentViewController.swift
+++ b/ownCloud/Client/Actions/EditDocumentViewController.swift
@@ -30,6 +30,7 @@ class EditDocumentViewController: QLPreviewController, Themeable {
 	var itemTracker: OCCoreItemTracking?
 	var modifiedContentsURL: URL?
 	var dismissedViewWithoutSaving: Bool = false
+	var timer: DispatchSourceTimer?
 
 	var source: URL {
 		didSet {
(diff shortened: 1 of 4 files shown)


ownCloud/Client/Actions/EditDocumentViewController.swift
```diff
@@ -95,26 +96,45 @@ class EditDocumentViewController: QLPreviewController, Themeable {
...
+
+	override func viewDidLoad() {
+		super.viewDidLoad()
+
+		let queue = DispatchQueue(label: "com.owncloud.edit-document-queue")
+		timer = DispatchSource.makeTimerSource(queue: queue)
+		timer!.schedule(deadline: .now(), repeating: .seconds(1))
+		timer!.setEventHandler { [weak self] in
+			OnMainThread {
+				if (UIDevice.current.isIpad && self?.navigationItem.rightBarButtonItems?.count ?? 0 > 1) || (!(self?.navigationController?.isToolbarHidden ?? false)) {
```
mneuwert, line 116: @hosy may be split this long if condition in few guards?

ownCloud/Client/Actions/EditDocumentViewController.swift
```diff
@@ -95,26 +96,45 @@ class EditDocumentViewController: QLPreviewController, Themeable {
...
+	deinit {
+		Theme.shared.unregister(client: self)
+		timer = nil
+	}
+
+	override func viewDidLoad() {
+		super.viewDidLoad()
+
+		let queue = DispatchQueue(label: "com.owncloud.edit-document-queue")
+		timer = DispatchSource.makeTimerSource(queue: queue)
```
mneuwert, line 112: @hosy timer always smells after race conditions.. Wouldn't it be enough to put the code within delayed block to `viewDidAppear()` for instance?

mneuwert (changes requested):

ownCloud/Client/Actions/EditDocumentViewController.swift
hosy: We had this implementation before and it does not work, because some items are available immediately and some items needs more seconds to load.

ownCloud/Client/Actions/EditDocumentViewController.swift
hosy: Could not split this in guard statements, because one statement is related for the iPad and the other one for the iPhone. If I add guard statements for each device, the second guard statement could not be reached.

If I am wrong, please append a proposal to the CR.

mneuwert (approved):

jesmrec:
I'm not able to reproduce the initial problem. Is there any way to? Otherwise, i will test the fix in regression mode.

hosy:
@jesmrec This problem is difficult to reproduce. Important is, that the edit mode will be enabled, when opening the document. You could try to check different file types, small and large document sizes.

jesmrec:
After hundreds of attempts, i was not able to reproduce the initial problem.

But i detected the following ones:


1. Turn the device in landscape
2. Select `Markup

Current: Markup mode is not enabled. `Share with` is displayed in the top right corner instead of the `Undo` and `Redo` actions

Screenshot 2021-04-07 at 14 43 50

Expected: Markup mode enabled (same as portrait orientation). `Share with` is not visible. `Undo` and `Redo`  are visible.

iPhoneXR V14.4
`a288a8dd`

hosy:
@jesmrec thank you for finding this edge case. I pushed an update and you can give it a new try now.

jesmrec:
The `Share with` icon is correct there? only appears in landscape. The undo/redo and the initial edition mode are fixed

hosy:
Yes, this is correct it will be set by the system and cannot be configured. ￼

jesmrec:
ok, then the problem of landscape mode is fixed.

The initial problem is not reproducible on my side after a many attempts with different files. Should i keep testing or let the final user give us feedback about it?

hosy:
Mark as approved and wait for negative user feedback. Thanks!

jesmrec (approved):

PR #955, opened 2021-04-20
[feature/simple-branding] Support for simplified branding / static login profile

felix-schwarz:
## Description
- create OCClassSettings bridge for StaticLoginProfile
- standardize static login profile keys, also allowing the use of OCClassSettings key notation ("bookmark-name" (new) vs "bookmarkName" (old))
- add placeholder metadata for static login profile keys

## Types of changes


ownCloudAppFramework/Branding/Branding.m
```diff
@@ -49,7 +49,7 @@ + (void)load
 
 				@"can-add-account" : @(YES),
 				@"can-edit-account" : @(YES),
-				@"enable-review-prompt" : @(NO)
+				@"enable-review-prompt" : @(YES)
```
hosy, line 52: Default value for branded clients should be `NO`

ownCloudAppShared/Branding/Branding+App.swift
```diff
@@ -45,7 +45,8 @@ extension Branding : BrandingInitialization {
 				.sendFeedbackAddress : "ios-app@owncloud.com",
 
 				.canAddAccount : true,
-				.canEditAccount : true
+				.canEditAccount : true,
+				.enableReviewPrompt : true
```
hosy, line 49: Default value for branded clients should be`false`

ownCloud/Static Login/StaticLoginProfile.swift
```diff
@@ -41,60 +42,270 @@ class StaticLoginProfile: NSObject {
...
+					.status		: OCClassSettingsKeyStatus.advanced,
+					.category	: "Branding",
+					.subCategory	: "Profile"
+				],
+
+				StaticLoginProfile.Key.allowedAuthenticationMethods.settingsKey : [
+					.type 		: OCClassSettingsMetadataType.stringArray,
+					.label		: "Allowed authentication methods",
+					.description 	: "The identifiers of the authentication methods allowed for this profile. Allows to f.ex. force OAuth2, or to use Basic Auth even if OAuth2 is available. See connection.allowed-authentication-methods for possible values.",
+					.status		: OCClassSettingsKeyStatus.advanced,
```
hosy, line 279: Does it makes sense to add the possible values? This will show inside the documentation which values are possible for that field.

```
.possibleValues : [
"Basic Auth" : "com.owncloud.basicauth", 
"OAuth2" : "com.owncloud.oauth2"
]
```

hosy (changes requested):

ownCloudAppFramework/Branding/Branding.m
felix-schwarz: I assumed that this value would be used even if the app is not branded, but now saw that this assumption was wrong. I changed it back to `NO` / `false` and added a note to the documentation for this parameter.

ownCloudAppShared/Branding/Branding+App.swift
felix-schwarz: I assumed that this value would be used even if the app is not branded, but now saw that this assumption was wrong. I changed it back to `NO` / `false` and added a note to the documentation for this parameter.

ownCloud/Static Login/StaticLoginProfile.swift
felix-schwarz: Yeah, it absolutely makes sense. I changed the SDK to make the dynamically generated values of `connection.allowed-authentication-methods` easily available for this, then returned from here, so it's now inlined as a table into the documentation.

hosy (approved):

hosy:
Decided to move QA after merge

michaelstingl:
Simplified branding makes admins life much easier! 😍 

```
  <dict>
    <key>branding.organization-name</key>
    <string>Demo-Organisation</string>
    <key>branding.profile-bookmark-name</key>
    <string>demo.owncloud.com</string>
    <key>branding.profile-url</key>
    <string>https://demo.owncloud.com</string>
    <key>branding.profile-allow-url-configuration</key>
    <false></false>
    <key>branding.can-add-account</key>
    <false></false>
    <key>branding.can-edit-account</key>
    <false></false>
  </dict>
```

IMG_0001 | IMG_0002 | IMG_0003
-- | -- | --

PR #961, opened 2021-04-29
[fix/upload-conflicts] Fix upload conflict handling bugs

felix-schwarz:
## Description
This PR fixes upload conflict handling bugs that can be reproduced as follows:
- view an image on two devices, annotate on both and save the changes only 1-2 seconds apart, provoking a collission
- the device that finished the upload later will prompt for:
a) `Cancel`: previously would keep the file around, with no way to download the server version. Now deletes the local copy.
b) `Replace`: previously could put the user in an infinite error -> response loop. Now "force replaces" the copy on the server by not sending any `If-Match` or `If-No-Match` headers along with the upload.
c) `Keep both`: previously would store the second copy with the same FileID in the database, causing issues. Now stores the changed file with a distinct FileID in the database.

QA extra: please also test if uploading a new file to the same location from different clients at the same time triggers the correct reaction and error handling.

## Types of changes

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Version%2011.6/Conflict%20Handling.md

Reports:

- [x] (1) `Cancel` behaviour https://github.com/owncloud/ios-app/pull/961#issuecomment-831220951 [NOT REPR]
- [x] (2) Glitch in dialog https://github.com/owncloud/ios-app/pull/961#issuecomment-831223610 [FIXED]
- [x] (3) Timestamp in new file after `Keep Both` https://github.com/owncloud/ios-app/pull/961#issuecomment-831225985 [FIXED]


hosy (approved):

jesmrec:
### (1)

Let me know what do you think about the following behaviour:

1. Open image with any client, and the same image with iOS client
2. Mark image up with both.
3. Save first the other client's version, and then, the iOS client 
4. Conflict dialog in iOS -> Select `Cancel` 

Current: Image modified locally stays in the screen

Expected: Local image is deleted, it should not be displayed. After clicking `Cancel` , prompted to file list (this is my suggestion, given the expected behaviour described above in top message)

iPhoneXR 14.4
iPadAir 13.4

`ad1a8105`

jesmrec:
### (2)

An small glitch in iPad's dialog when the file name is a bit long:

Screen Shot 2021-05-03 at 13 34 31

jesmrec:
### (3)

1. Conflict -> select `Keep Both` 

Current: New file contains a timestamp, that seems to be CEST -2. Created at (2021-05-03 13:55:10), timestamp in the file name  (2021-05-03 11:55:10)

Expected: correct timestamp

iPhoneXR 14.4
iPadAir 13.4

`ad1a8105`

felix-schwarz:
@jesmrec Fixed (3) in https://github.com/owncloud/ios-app/pull/961/commits/a854a0c2d28190e4e14d0b2b43409a9d780c0194 - looking into the other two now.

felix-schwarz:
@jesmrec Regarding (1):
The expected behaviour is that the "cancelled" local image should be discarded and the new version be downloaded and displayed.

I could not reproduce the issue. Was that a large image?

jesmrec:
About (1), i have rechecked and i can not reproduce anymore. I will check again after new commits or in the regression stage

felix-schwarz:
@jesmrec Issue (2) should now also be fixed as of 5ddff36.

jesmrec:
(2) is fixed.

So, feature approved. Tests performed here will be included in the [global regression test](https://github.com/owncloud/QA/blob/master/Mobile/iOS-app/Regression.md)

jesmrec (approved):

PR #982, opened 2021-06-07
[fix/media-import-blocking] Fix media import blocking

felix-schwarz:
## Description
- avoid multiple KVS instances (created by the media upload) competing for time in the same process, by using new SDK semantics for shared KVS instances

## Related Issue
https://github.com/owncloud/enterprise/issues/4547

## Types of changes


hosy (approved):

hosy:
This one is difficult to test and I think we should drop QA and test it in wild directly with the issue reporter.

PR #991, opened 2021-06-21
[fix/media-import-blocking] Fix media import blocking

felix-schwarz:
## Description
MediaUploadStorage:
- use dedicated key value store rather than that of the bookmark
- migrate existing jobs to new, dedicated KVS

## Related Issue
https://github.com/owncloud/enterprise/issues/4547

## Types of changes


hosy (approved):

hosy:
@jesmrec I think this one is hard to test. Do you agree when I move forward without QA or what is your opinion?

jesmrec:
yes, i took a look when the ticket was opened and it is not easy to reproduce... the best way to check this internal stuff is with the direct feedback from affected users.

It is OK from me moving forward with no specific QA @hosy

PR #1017, opened 2021-08-06
[fix/photouploadcrash] Fix automatic photo upload crash on iOS 15

felix-schwarz:
## Description
On iOS 15, automatic photo upload seems to consume more resources than are available, leading to a crash. This pull requests reduces the number of concurrent photo upload operations from `available cores` to `1`.

## Types of changes


hosy (approved):

jesmrec:
This one is related with performance and not easy to reproduce. I'd move it forward and wait for feedback in case the problem persist. What do you think?

hosy:
@jesmrec yes I think this is not easy to reproduce and you can finish without further testing. Thanks!

PR #1029, opened 2021-09-09
[feature/mdm-lockinterval] MDM-configurable App Lock Interval

felix-schwarz:
## Description
New MDM / class setting option `passcode.lockDelay` to enforce locking after `N` seconds. Example entry in a `Branding.plist` for 5 minutes (600 seconds):

```xml
<key>passcode.lockDelay</key>
<integer>600</integer>
```

## Related Issue
https://github.com/owncloud/enterprise/issues/4741

## Types of changes


hosy (approved):

jesmrec:
### (1)

I've detected a small glitch. Maybe it's not worthy to fix it... release your thoughts about:

1. Install app from MDM (don't enable passcode yet!)
2. Set the lockDelay via MDM
3. Push file to device via MDM
4. Go to `Settings` and enable Passcode

Current: 

`Lock Application` shows `Immediately` instead of the value pushed via MDM. By killing the app and reopening, correct value is displayed

iPhoneXR v14.7

`fe91cdc0a`

this shouldn't be a blocker, but fixes are always welcome. I'll delegate on you the decision of fixing now or addressing in separate issue. Both work for me.

jesmrec:
### (1)

I've detected a small glitch. Maybe it's not worthy to fix it... release your thoughts about:

1. Install app from MDM (don't enable passcode yet!)
2. Set the lockDelay via MDM
3. Push file to device via MDM
4. Go to `Settings` and enable Passcode

Current: 

`Lock Application` shows `Immediately` instead of the value pushed via MDM. By killing the app and reopening, correct value is displayed

iPhoneXR v14.7

`fe91cdc0a`

this shouldn't be a blocker, but fixes are always welcome. I'll delegate on you the decision of fixing now or addressing in separate issue. Both work for me.

felix-schwarz:
@jesmrec Re (1): new MDM settings are only fully applied after a restart of the app to avoid inconsistencies:

- if the app was already running while new settings were pushed via MDM, the user will get a notification to relaunch the app.
- if the app was not running when new settings were pushed via MDM, the new settings should be active as soon as the app is launched the next time.

jesmrec:
Thanks for clarification. We will survive with the glitch, for sure. 

Feature is approved.

jesmrec (approved):

PR #1049, opened 2021-10-14
[fix/single-accouont-display-name] Show Display Name in Branded Single Account View

hosy:
## Description
Username could look like a UUID, which is not user friendly. Now showing the `displayName`, if available, other show the `userName`.

## Related Issue
https://github.com/owncloud/enterprise/issues/4798

## How Has This Been Tested?
- brand app as oc.online
- configure an account
- check header for `displayName`

## Types of changes

## Checklist:


hosy:
@felix-schwarz it looks like, that the property `display-name` is only available, after a connection was initiated.
Is it possible to make this property available, without a connection?
https://github.com/owncloud/ios-app/blob/e7abd1d643f3d4c76bde48a2949ac0d12061d8ef/ownCloudAppFramework/SDK%20Extensions/OCBookmark%2BAppExtensions.m#L23

felix-schwarz:
@hosy Unfortunately that's not possible, since that info (typically the full name of a person) can only be retrieved from the server. It is set in `BookmarkViewController.swift:569` from values returned from `user.php` during account setup.

hosy:
> @hosy Unfortunately that's not possible, since that info (typically the full name of a person) can only be retrieved from the server. It is set in `BookmarkViewController.swift:569` from values returned from `user.php` during account setup.

Is it possible to add a workaround, loading user data, when this information is needed?

felix-schwarz:
@hosy In that case, you can use the bookmark to create an `OCConnection` and then call `- (NSProgress *)connectWithCompletionHandler:(void(^)(NSError *error, OCIssue *issue))completionHandler` on that instance. Then, when the `completionHandler` is run and no `error` or `issue` was returned, you can grab the info from `connection.loggedInUser.displayName`.

hosy:
@jesmrec now I'm retrieving the users display name from server and reloading the view, after it was retrieved. It cannot be loaded, if the server is not available. Please have another look. Thanks!

jesmrec:
Tested:

- [x] Static display name
- [x] Display name's changes

Everything correct on my side

Approved

jesmrec (approved):

PR #1054, opened 2021-10-16
[feature/modular-localization] Modular localization

felix-schwarz:
## Description
Switches `OCLocalized`, `OCLocalizedString` and `String.localized` from `NSLocalizedString` and `NSBundle` to `OCLocale`, allowing complex customization of localized strings with variables, value sources and complete text replacements.

That makes it possible to change the localization of specific strings via class settings. To f.ex. change `Settings` to `Preferences` (en) and `Präferenzen` (de), one would use:

```xml
<key>locale.overrides</key>
<dict>
	<key>Settings</key>
	<dict>
		<key>en</key>
		<string>Preferences</string>

		<key>de</key>
		<string>Präferenzen</string>
	</dict>
</dict>
```

Furthermore, localized strings can now contain variables that are getting replaced during localization.

### Global variables
For example:
```swift
Log.log("{{app.name}} {{app.process}} {{app.version}} {{app.build}} {{device.model}} {{device.model-id}} {{os.name}} {{os.version}}".localized)
```
prints
```
ownCloud ownCloud 11.7.1 201 iPhone x86_64 iOS 15.0
```
when run in the Simulator.

### Template variables
Template variable support can be used this way:

```swift
Log.log("{{product}} is awesome!".localized([ "product" : "ownCloud" ]))
```
prints
```
ownCloud is awesome.
```

In the `Localized.strings` file, templates appear like any other string:

```
"{{product}} is awesome!" = "{{product}} ist großartig!";
```

### Global vs. template variables
As a convention to avoid conflicts:
- global variables **must** use dot syntax (i.e. `app.name`), whereas
- template variables **must not** use dot syntax (i.e. `appName`)

## Related Issue
https://github.com/owncloud/enterprise/issues/4801

Changes in en | Changes in de
--|--
Simulator Screen Shot - iPhone 12 mini - 2021-10-16 at 22 28 40 | Simulator Screen Shot - iPhone 12 mini - 2021-10-16 at 22 28 17

## Types of changes

## Checklist:

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,7 +3,7 @@ excluded:
     - ownCloudScreenshotsTests
     - external
 
-function_body_length: 100
+#function_body_length: 100
 # warning_threshold: 5
 # line_length:
 #  warning: 300
@@ -33,8 +33,10 @@ disabled_rules:
     - function_body_length
     - implicit_getter
     - computed_accessors_order
-    - unneeded_notification_center_removal
     - function_parameter_count
+    - comment_spacing
+    - unused_closure_parameter
+#   - unneeded_notification_center_removal
 custom_rules:
   empty_line_after_guard_statement:
     included: ".*\\.swift"
--- a/ownCloudAppFramework/Branding/Branding.h
+++ b/ownCloudAppFramework/Branding/Branding.h
@@ -51,6 +51,7 @@ typedef NSString* BrandingImageName NS_TYPED_EXTENSIBLE_ENUM;
 @property(assign,nonatomic,readonly) BOOL brandingPropertiesFromLocalFile;
 
 @property(strong,nullable,nonatomic,readonly) NSString *appName; //!< Custom app name
+@property(strong,nonatomic,readonly) NSString *appDisplayName; //!< Branded app name, drawing from .appName, .organizationName and OCAppIdentity, with "ownCloud" as fallback
 @property(strong,nullable,nonatomic,readonly) NSString *organizationName; //!< Custom organization name
 @property(strong,nullable,nonatomic,readonly) NSArray<BrandingFileImportMethod> *disabledImportMethods; //!< Disabled file import methods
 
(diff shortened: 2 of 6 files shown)


jesmrec:
@fschwarz regarding https://github.com/owncloud/ios-app/pull/1054:  When i set a device in a language (f.ex. Spa) and no entries for spanish in `locale.overrides` for an specific string, it's getting the first in the list, no matter which one. Is that correct? F.ex:

```
<key>Settings</key>
			<dict>
				<key>de</key>
				<string>Präferenzen</string>
				<key>fr</key>
				<string>Paramètres</string>
			</dict>
```

the device in Spanish language will display the German translation because it is the first in the list. If i add the english, i will take english. Does that mean that english is default and "mandatory"?

felix-schwarz:
> @fschwarz When i set a device in a language (f.ex. Spa) and no entries for spanish in `locale.overrides` for an specific string, it's getting the first in the list, no matter which one. Is that correct? F.ex:

No, that's not correct. For overrides like this, the app will pass a list of available localization languages (`de`, `fr` in your example) to iOS and iOS will then return them ordered by user preference. That order can be changed by the user in iOS Settings > General > Language & Region > Preferred language order.

You can get the list from the system by pausing execution of the app, then typing

```
po NSLocale.preferredLanguages
```

in the debugger. For my iPhone, it returns:

```
<__NSArrayM 0x13bf20390>(
de-DE,
en-DE
)
```

I'm not sure what language iOS will return if none of the languages is in the list of preferred languages. The documentation hints at it finding the next best locale based on the user's region: https://developer.apple.com/documentation/foundation/nsbundle/1417249-preferredlocalizationsfromarray?language=objc

If that doesn't provide the needed results, I can add a fallback option that checks if `NSLocale.preferredLanguages` contains any of the languages that the override provides - and if it doesn't, uses a particular localization instead.

> the device in Spanish language will display the German translation because it is the first in the list. If i'd add the english, i will take english. Does that mean that english is default and "mandatory"?

Based on how locale selection appears to work, I'd assume it picks the next best understood language for the average person in the user's region. So if you have set `es` and `en` as preferred languages in iOS Settings, iOS seems to assume the average person in your region is more likely to understand `de` than `fr`, and also more likely to understand `en` than `de`.

jesmrec:
> I think we can leave it as is for now, since localization overrides would typically be provided for all languages used in an organization. If the need arises nonetheless, we can still add a fallback option locale.override-fallback-language to specify a language to fall back to if there are no matching locales for what the user has set up on the device.

totally agreed. Let's move this forward having in mind the corner case and its fix, just in case it arises.

PR #1080, opened 2021-12-20
[fix/connection-stuck] Fix eternal "Connecting…" 

felix-schwarz:
## Description
- avoids parallel use of `OCConnection` for the same bookmark
- uses the SDK's new `OCBookmark.userDisplayName` instead of retrieving the user's displayName on the fly

## Related Issue
https://github.com/owncloud/enterprise/issues/4925

## Types of changes


jesmrec:
Tested with different servers

- [X] OIDC with failed DCR (base case)
- [X] OIDC with correct DCR
- [X] OIDC without DCR
- [X] OAuth2 (regression)
- [X] Basic auth (regression)

hosy (approved):

PR #1083, opened 2021-12-29
[fix/pdf-editing] Fix PDF editing

felix-schwarz:
## Description
- fixes bug that prevents changes to PDFs being saved in place.
- after editing, the viewer no longer asks if the PDF file should be refreshed, but instead the viewer refreshes it automatically to reflect the changes.

## Related Issue
https://github.com/owncloud/enterprise/issues/4934

## Types of changes

### DIFF
--- a/ownCloud/Client/Actions/Actions+Extensions/DocumentEditingAction.swift
+++ b/ownCloud/Client/Actions/Actions+Extensions/DocumentEditingAction.swift
@@ -82,9 +82,10 @@ class DocumentEditingAction : Action {
 				if let fileURL = files.first?.url, let item = self.context.items.first {
 
 					if QLPreviewController.canPreview(fileURL as QLPreviewItem) {
-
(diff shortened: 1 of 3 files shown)


hosy (approved):

jesmrec:
About this one:

- The original problem reported in [central](https://central.owncloud.org/t/cant-save-edited-pdfs-in-owncloud-error-35/35684) is not reproducible on my side. I can edit and save overwritten files correctly with `11.8.1`. Tested with two different devices (iPad/iPhone) and different version. Also i tried different combinations of operations (overwrite + save, save + overwrite, save + overwrite + save etc...). I got never the error 35.

- The dialog `Showing latest version` only appears now if the file is updated externally, not after a local update  ✅  ✅ 

So, we can go with the approach here to get extra feedback from the reporter.

jesmrec (approved):

Issue #972, opened 2021-05-25
[FEATURE REQUEST] Rename accounts without the login flow


fvdm:
**Is your feature request related to a problem? Please describe.**
To simply rename an account you have to complete the login procedure all over again.

Steps:
1. Go to the Accounts list
2. Swipe left on one of the accounts
3. Tap the _Edit_ button
4. Alter the _Name_ field
5. Tap _Continue_ to save the new value

Instead of going back to the _Accounts_ list, you are asked to login at the server for the access permissions.

**Describe the solution you'd like**
Save the new account name without the login procedure.

**Describe alternatives you've considered**
Maybe move the account renaming to _Manage_ instead of _Edit_ if a save button is not possible there. And then relabel _Edit_ in the swipe menu to _Connection_.

**Additional context**

- ownCloud app v11.6.0
- The account is properly linked and working before the renaming.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

jesmrec:
Done and released in 11.9.1

PR #1097, opened 2022-01-31
[feature/rename-accounts] Rename Account without Authentication Flow

hosy:
## Description
Check if only the account name was changed in edit mode: save and dismiss without re-authentication

## Related Issue
#972 

## Motivation and Context
Easier account editing, without annoying the user with re-authentication

## How Has This Been Tested?
- Edit existing account
- Change name
- Tap on `Continue`
- View should be dismissed without authentication flow
- Try again with changing the URL or only changing the URL
- Authentication should now happen

## Types of changes

## Checklist:

## QA

Reports:

- [ ] (1) Bad credentials pass https://github.com/owncloud/ios-app/pull/1097#issuecomment-1028916480
- [ ] (2) Username missing https://github.com/owncloud/ios-app/pull/1097#issuecomment-1028917382
- [ ] (3) OAuth2 flow https://github.com/owncloud/ios-app/pull/1097#issuecomment-1028925436

### DIFF
--- a/ownCloud/Bookmarks/BookmarkViewController.swift
+++ b/ownCloud/Bookmarks/BookmarkViewController.swift
@@ -535,11 +535,14 @@ class BookmarkViewController: StaticTableViewController {
 	}
 
 	func completeAndDismiss(with hudCompletion: @escaping (((() -> Void)?) -> Void)) {
+		guard let userActionCompletionHandler = self.userActionCompletionHandler else { return }
+
+		self.userActionCompletionHandler = nil
+
 		OnMainThread {
 			hudCompletion({
 				OnMainThread {
-					self.userActionCompletionHandler?(self.bookmark, true)
-					self.userActionCompletionHandler = nil
+					userActionCompletionHandler(self.bookmark, true)
 				}
 				self.presentingViewController?.dismiss(animated: true, completion: nil)
 			})
--- a/ownCloud/Bookmarks/BookmarkViewController.swift
+++ b/ownCloud/Bookmarks/BookmarkViewController.swift
@@ -343,20 +343,8 @@ class BookmarkViewController: StaticTableViewController {
 
 		// Check if only account name was changed in edit mode: save and dismiss without re-authentication
 		if mode == .edit, nameChanged, !urlChanged, let bookmark = bookmark {
-			self.originalBookmark?.setValuesFrom(bookmark)
-			if let originalBookmark = self.originalBookmark, !OCBookmarkManager.shared.updateBookmark(originalBookmark) {
-				Log.error("Changes to \(originalBookmark) not saved as it's not tracked by OCBookmarkManager!")
-			}
-
-			OnMainThread {
-				hudCompletion({
-					OnMainThread {
-						self.userActionCompletionHandler?(self.bookmark, true)
-					}
-					self.presentingViewController?.dismiss(animated: true, completion: nil)
-				})
-			}
-
+			updateBookmark(bookmark: bookmark)
+			completeAndDismiss(with: hudCompletion)
 			return
 		}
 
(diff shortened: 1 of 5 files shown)


ownCloud/Bookmarks/BookmarkViewController.swift
```diff
@@ -523,6 +534,18 @@ class BookmarkViewController: StaticTableViewController {
 		}
 	}
 
+	func completeAndDismiss(with hudCompletion: @escaping (((() -> Void)?) -> Void)) {
+		OnMainThread {
+			hudCompletion({
+				OnMainThread {
+					self.userActionCompletionHandler?(self.bookmark, true)
+					self.userActionCompletionHandler = nil
```
felix-schwarz, line 542: This sending between threads and completion handlers carries the risk of `userActionCompletionHandler` being called more than once if the user if the user taps the Continue button in quick succession.

The previous version protected against this by putting `self.userActionCompletionHandler` into a variable, then setting it to `nil` - and only then sending off the block to the main thread.

Here's what I'd suggest:
```swift
	func completeAndDismiss(with hudCompletion: @escaping (((() -> Void)?) -> Void)) {
		guard let userActionCompletionHandler = self.userActionCompletionHandler else {
			return
		}

		self.userActionCompletionHandler = nil

		OnMainThread {
			hudCompletion({
				OnMainThread {
					userActionCompletionHandler(self.bookmark, true)
				}
				self.presentingViewController?.dismiss(animated: true, completion: nil)
			})
		}
	}
```

felix-schwarz (changes requested):

felix-schwarz (approved):

jesmrec:
### (1)

I got this one:

1. Edit account (basic auth)
2. In edit view, change name and then, enter incorrect credentials 
3. Click on `Continue`

Current: No error displayed. When list of files is opened, the `Authorization failed` dialog raises up

Expected:  (only an idea) when clicking on `Continue`, credentials checked in case they were filled up. If no credentials typed,  only name saved. 

iPhoneXR v15.3, iPadAir v14.8
`b2d3f3ff`

jesmrec:
### (2)

Probably related with (1)

In the `Edit account` view for basic auth:

Current: username is missing and editable.
Before: username was set and not editable.

Is this behaviour change expected?

iPhoneXR v15.3, iPadAir v14.8
`b2d3f3ff`

jesmrec:
### (3)

With browser authentication (OAuth2, OIDC) it's not posible to re-authenticate if the name changes. As defined above, URL has to change (i guess URLs does not use to change) to trigger the authentication flow. That means:

1. Credentials change in server
2. Edit Login (OAuth2) in app
3. Change name

Current: `Continue` changes the name but does not trigger the auth flow. As the credentials changed (this is the main reason to do an edit), the flow should be triggered.
Expected: Authentication flow triggered regardless name changes (am i right?)

For me it's not clear the expectation for this corner case in which both name and credentials are modified at the time. 

iPhoneXR v15.3, iPadAir v14.8
`b2d3f3ff`

hosy:
@jesmrec (1), (2), (3) are fixed:

- show UI to delete authentication data in edit mode (please test, with one or more accounts configured, context menu, swipe action, edit table view action, if delete authentication UI appears)
- only proceed without authentication, if authentication data is not nil

jesmrec:
@hosy 

(1) and (2) are fixed

(3) -> the way to re-authenticate with OAuth2 is with `Delete Authentication Data` right? otherwise, only the name changes. Please, confirm this is correct.

hosy:
@jesmrec yes, this is correct.

jesmrec:
Tested with two accounts together (Basic & OAuth2). Checks done:

- [X] Edit action in swipe - Change name
- [X] Edit action in swipe - Delete & Change authentication data - Enter correct credentials
- [X] Edit action in swipe - Delete & Change authentication data - Enter bad credentials
- [X] Edit action in swipe - Delete & Change authentication data + Change name
- [X] Edit action in swipe - Change nothing
- [X] Edit action in contextual - Change name
- [X] Edit action in contextual - Delete & Change authentication data - Enter correct credentials
- [X] Edit action in contextual - Delete & Change authentication data - Enter bad credentials
- [X] Edit action in contextual - Delete & Change authentication data + Change name
- [X] Edit action in contextual - Change nothing
- [X] Add new account with name
- [X] Add new account without name

jesmrec:
Functional tests has passed and feature works as expected. 

But, the re-authentication flow has changed and now authentication data must be cleaned. This is something new and maybe users don't expect that behaviour and try to authenticate by using the `Save` button as before. 

Should we include a tip about editing credentials just below the `Delete Authentication Data`? something like:

`Please, before typing new credentials delete old ones. "Save" button will only save the bookmark name`

jesmrec:
For me this is approved!

jesmrec (approved):

PR #1105, opened 2022-03-10
[fix/token-timetolerance] Retry requests that failed with a 401 during a token refresh

felix-schwarz:
## Description
Re-schedules/retries requests that were:
- sent **before** a token refresh AND
- received a response **after** that token refresh AND
- **then** failed with a 401 status code
using the latest token.

Different tokens are identified by comparing the `OCAuthenticationDataID` of the authentication data used when authorizing the request with the latest `OCAuthenticationDataID` of the bookmark.

This PR also adds an improved representation of the `Authorization` header in the logs. Those now come with a counter that issues a new number whenever the content of the `Authorization` header changes.

Those counters are specific to every process, so they likely will not match between File Provider and app. But it allows to see in the logs when token changes occurred in the respective processes. The original idea - logging the `OCAuthenticationDataID` - would have provided the same IDs for all processes, but was thrown out for privacy/security concerns: `OCAuthenticationDataID` is a checksum computed from `authenticationData` - and there would have been the remote possibility that this would provide hints to the actually `[redacted]` content.

Example: the `Authorization` header before a change to authentication data:
```
# REQUEST ---------------------------------------------------------
…
PROPFIND /remote.php/dav/files/demo HTTP/1.1
Authorization: Bearer [redacted:0]
```

… and after:

```
# REQUEST ---------------------------------------------------------
…
PROPFIND /remote.php/dav/files/demo HTTP/1.1
Authorization: Bearer [redacted:1]
```

## Related Issue
https://github.com/owncloud/enterprise/issues/5059

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)


ownCloud.xcodeproj/project.pbxproj
```diff
@@ -4730,8 +4730,8 @@
 				APP_BUILD_FLAGS = "$(inherited)";
 				APP_BUILD_FLAGS_SWIFT = "$(APP_BUILD_FLAGS)";
 				APP_PRODUCT_NAME = ownCloud;
-				APP_SHORT_VERSION = 11.8.2;
-				APP_VERSION = 208;
+				APP_SHORT_VERSION = 11.8.3;
+				APP_VERSION = 210;
```
hosy, line 4734: Versioning should be removed, because it will be also merged into the `milestone/11.9.0`

ownCloud.xcodeproj/project.pbxproj
```diff
@@ -4799,8 +4799,8 @@
...
```
hosy, line 4803: Versioning should be removed, because it will be also merged into the `milestone/11.9.0`

hosy (changes requested):

hosy:
@felix-schwarz please set the correct merge branch

felix-schwarz:
Merged as part of SDK changes into `11.9.0`.

PR #1169, opened 2022-10-26
Add OWNCLOUD_TRUSTED_DOMAINS to the Docker example

michaelstingl:
Needed because of this change: https://github.com/owncloud-docker/server/pull/340

## Description

Add `OWNCLOUD_TRUSTED_DOMAINS` to the Docker example

## Related Issue

https://github.com/owncloud-docker/server/pull/340

## Motivation and Context

Change in the ownCloud Docker now require to configure the trusted domains

## How Has This Been Tested?

Run:

```
docker run --rm -d \
  --name owncloud \
  -p 18080:8080 \
  -e OWNCLOUD_TRUSTED_DOMAINS=192.168.68.64 \
  -e OWNCLOUD_APPS_ENABLE=oauth2 \
  -e ADMIN_USERNAME=admin \
  -e ADMIN_PASSWORD=admin \
  owncloud/server:latest
```

Then check if it works with browser:
http://192.168.68.64:18080/login

## Types of changes

## Checklist:

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -97,7 +97,7 @@ docker run --rm -d \
  http://localhost:18080/  
  user: admin  
  password: admin
-4. To test the  app on a iOS device in your local network, you can use the IP of your machine as the Server URL
+4. To test the  app on a iOS device in your local network, you can use the IP of your machine as the Server URL. To do so, you'll need to add your IP as a trusted domain with a new environment variable `OWNCLOUD_TRUSTED_DOMAINS=192.168.68.64` to the Docker command
 5. To remove the ownCloud Docker, just run `docker kill owncloud`
 
 Here you can find more information about running ownCloud with Docker:


SETUP.md
```diff
@@ -97,7 +97,7 @@ docker run --rm -d \
  http://localhost:18080/  
  user: admin  
  password: admin
-4. To test the  app on a iOS device in your local network, you can use the IP of your machine as the Server URL
+4. To test the  app on a iOS device in your local network, you can use the IP of your machine as the Server URL. To do so, you'll need to add your IP as a trusted domain with a new environment variable `OWNCLOUD_TRUSTED_DOMAINS=192.168.68.64` to the Docker command
```
hosy, line 100: Does it make sense to add that 192.168.68.64 is a sample IP address and needs to be adopted to the own local network?

hosy (changes requested):

SETUP.md
michaelstingl: @hosy propose a change? Add "(example)" to make it more clear?

DeepDiver1975: When accessing the docker container via localhost:8080 there is no need in setting up the trusted domain env var.

Why is the docker container started via 18080?

jesmrec: > When accessing the docker container via localhost:8080 there is no need in setting up the trusted domain env var.

for mobile devices it's required. I guess 18080 is just an example as any other port.

DeepDiver1975: > for mobile devices it's required.

why?

jesmrec: it's not required if you are in localhost as you stated, so, useless verbosity on my side (sorry!). In case of accessing from mobile, that env variable will add the public ip to the whitelist to access from outside of `localhost` (f. ex, from an external device).

PR #1281, opened 2023-09-22
generate.sh for easy theme generation

michaelstingl:
@steelcuts open for improvements…

### DIFF
--- a/ownCloud/Resources/Theming/generate.sh
+++ b/ownCloud/Resources/Theming/generate.sh
@@ -8,17 +8,21 @@ current_dir=$(pwd)
 theming_dir="/ownCloud/Resources/Theming"
 theme=$1
 
-git submodule init
-git submodule update
+if [ -d "$theming_dir" ]; then
+	git submodule init
+	git submodule update
 
-rename 's/current./theme./' $current_dir$theming_dir/*.*
-mv $current_dir$theming_dir/theme.$theme $current_dir$theming_dir/current.$theme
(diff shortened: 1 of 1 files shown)


ownCloud/Resources/Theming/generate.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+# PREREQUISITE
+# Create symlink
+# ln -s SYNCDIR\ Config\ Data/5500Z000003poZfQAI/ios/current ~/Developer/github.com/owncloud/ios-app/ownCloud/Resources/Theming/theme.damken
+
+current_dir=$(pwd)
+theming_dir="/ownCloud/Resources/Theming"
+theme=$1
+
```
hosy, line 10: I would add a check, if the script is executed in the app repository path. Otherwise script could fail, when the `current_dir` is not the root of the repository path:

`if [ -d "$theming_dir" ]; then`
`# add all lines below`
`fi`

hosy (changes requested):

ownCloud/Resources/Theming/generate.sh
michaelstingl: Sounds perfect 👍 I also thought about this. Could you commit in the branch?

hosy: yes, I will do!

hosy (approved):

PR #1343, opened 2024-04-08
[fix/available-offline-policy-cleanup] Fix cleanup of Available Offline policies targeting unavailable spaces

felix-schwarz:
## Description
Fix an issue arising from Available Offline policies targeting unavailable/detached drives and removes the respective policies.

## Types of changes

### DIFF
--- a/ownCloud.xcodeproj/xcshareddata/xcschemes/ownCloud.xcscheme
+++ b/ownCloud.xcodeproj/xcshareddata/xcschemes/ownCloud.xcscheme
@@ -482,6 +482,11 @@
             value = "false"
             isEnabled = "NO">
          </EnvironmentVariable>
+         <EnvironmentVariable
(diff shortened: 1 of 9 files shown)


jesmrec:
Any way to reproduce the initial problem?

felix-schwarz:
@jesmrec What should work:
1) Build and run a prior version (`milestone/12.2` should do)
2) Add an ocis account with a space
3) Cut the internet connection (or stop the server) so the app no longer has a connection to the server
4) In the app mark a file in the space as "available offline"
5) On the server, delete the space
6) Bring the app or server back online

The app will then try to download the file, but keep re-scheduling the download even if the download failed or you cancelled it.

Switching to `fix/available-offline-policy-cleanup` then will fix that issue and the download will not be re-scheduled after it failed or has been cancelled.

PR #1344, opened 2024-04-11
[feature/url-shortcuts] Open and create .url shortcut files

felix-schwarz:
## Description
Adds the ability to open shortcuts stored in `.url` files as well as to create them.

The MDM option `action.open-shortcut-mode` provides control over what kind of shortcuts (URLs, items) should be allowed to be opened. Defaults to `all`.

Action in empty folder and on `+` button | Creating a shortcut |  Shortcut in file list
--|--|--
Simulator Screenshot - iPad Air (5th generation) - 2024-04-17 at 15 34 12 | Simulator Screenshot - iPad Air (5th generation) - 2024-04-17 at 15 30 20 | Simulator Screenshot - iPad Air (5th generation) - 2024-04-17 at 15 33 20

Create file shortcut | Create URL shortcut | Opening a URL shortcut
--|--|--
Simulator Screenshot - iPad Air (5th generation) - 2024-04-17 at 15 31 13<br>Simulator Screenshot - iPad Air (5th generation) - 2024-04-17 at 15 31 17 | Simulator Screenshot - iPad Air (5th generation) - 2024-04-17 at 15 33 14 | Simulator Screenshot - iPad Air (5th generation) - 2024-04-17 at 15 33 25

## Types of changes


hosy (changes requested):
I would suggest to remove the URLDisplayViewController. Instead I would choose one of the following approaches: 
- show an `UIAlertViewController` with question "Open Link?", on "Yes" it opens the link in the system browser
- open the link directly In-App in a `SafariViewController`

The detail view does not give any further content to the user and can be removed in my opinion.

jesmrec:
I did a first look into the feature (before QA - CR), with a couple of comments on my side

- Shortcuts to url show the following question: `Should it be opened in the default browser?` Well, this question is not really a question since there is no other choice than opening in the default browser (cancel is not a real option to open). So, i'd change that for an statement like  `It will be opened in the default browser` to clarify what's going to happen. More complex options like letting user select another browser to open are not worthy IMO (future improvement¿?). But, questions to users should have at least two real choices. Also, removing the question is not a bad option... the easier the better. Just as a suggestion.

- As improvement: if the target URL does not exist (for example, a link to a file or folder that has been removed from account), the message is `Operation couldn't be completed`.  It would be cool a more accurate message like `file does not exist`. Private links show: `item is not known to the server` as another example

Will do deeper testing after QA but it really looks good at this point.

felix-schwarz:
@jesmrec Thanks for the feedback. I've trimmed down the alert asking for opening the link and removed the browser reference entirely. Also, the error message when the target of a shortcut can't be found is now on point and informative (or at least I hope so 😉).

I also added a `action.open-shortcut-mode` MDM option that allows admins to control which type of shortcut content they want to allow (all, URLs only, items only, none).

hosy (approved):
Awesome feature and implementation!

jesmrec:
QA checks:

- [ ] Option in `+` 
- [ ] URL or name empty
- [ ] URL of web 
- [ ] URL of file in oC (same or different account)
- [ ] URL of file out of oC
- [ ] Incorrect URL (points nowhere)

jesmrec:
### (1)

1. `Create shortcut` 
2. Add as URL a IP. I used the ip of a oC server like: `https://192.168.1.20:9200`
3. Create shortcut
4. Click on the shortcut on list of file

Current: not opened (does the shortcut need a real URL and not a IP?? )

iPhoneXR iOS17
`f1b6e0e6`

jesmrec:
### (2)

As a question: shortcuts over files are only allowed over items in the same account, right? not posible to create a shortcut that points to a file in other account attached to the device (moving to there).

felix-schwarz:
@jesmrec Regarding (1): 

The current implementation supporting `.url` files depends on the MIME type `text/uri-list`, which OC10 appears not to use for `.url`-files:

```xml
<?xml version="1.0"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:oc="http://owncloud.org/ns">
  <d:response>
    <d:href>/remote.php/dav/files/admin/Files.url</d:href>
    <d:propstat>
      <d:prop>
        <d:resourcetype/>
        <d:getlastmodified>Fri, 03 May 2024 15:48:47 GMT</d:getlastmodified>
        <d:getcontentlength>51</d:getcontentlength>
        <d:getcontenttype>application/octet-stream</d:getcontenttype>
        <d:getetag>"ba2042d735ee8e0bd77c95655aedb8e6"</d:getetag>
        <oc:id>00009374oc60lgadtioc</oc:id>
        <oc:size>51</oc:size>
        <oc:permissions>RDNVW</oc:permissions>
        <oc:favorite>0</oc:favorite>
        <oc:share-types/>
        <oc:owner-id>admin</oc:owner-id>
        <oc:owner-display-name>admin</oc:owner-display-name>
        <oc:checksums>
          <oc:checksum>SHA1:9405c1b721fdd574d50d639dc14983f8906f45fd MD5:cdc846c7f7b2980ca79b452fb3015f9b ADLER32:f32e1229</oc:checksum>
        </oc:checksums>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
</d:multistatus>
```

Once OC10 returns `text/uri-list` for `.url`-files, handling in the iOS app should work as expected.

Regarding (2):
There's currently no support for cross-server shortcuts.

jesmrec:
That's approved on my side.

jesmrec (approved):

jesmrec:
Before merging, the branch should point to a different base branch like `milestone/12.3`

PR #1347, opened 2024-04-18
[fix/lock-bypass-prevention] Detect attempts to bypass a lock timeout by changing the clock

felix-schwarz:
## Description
- detect attempts to bypass a lock timeout by changing the clock
- code cleanups/modernizations of `AppLockManager`
- supersedes #1324

## Related Issue
https://github.com/owncloud/security-tracker/issues/413

## Types of changes

### DIFF
--- a/changelog/12.2.1_2024-04-30/1347
+++ b/changelog/12.2.1_2024-04-30/1347
@@ -1,0 +1,7 @@
+Fix: Detect attempts to bypass a lock timeout by changing the clock
+
+- detect attempts to bypass a lock timeout by changing the clock
+- code cleanups/modernizations of `AppLockManager`
+- supersedes #1324
+
+https://github.com/owncloud/ios-app/pull/1347


ownCloudAppShared/AppLock/AppLockManager.swift
```diff
@@ -417,54 +435,122 @@ public class AppLockManager: NSObject {
...
+	private func showOrHideLockCountdownAsNeeded() {
+		if shouldDisplayCountdown {
 			performPasscodeViewControllerUpdates { (passcodeViewController) in
 				passcodeViewController.keypadButtonsHidden = true
 				passcodeViewController.view.setNeedsLayout()
 			}
+
 			updateLockCountdown()
 
 			lockTimer = Timer.scheduledTimer(timeInterval: 1, target: self, selector: #selector(self.updateLockCountdown), userInfo: nil, repeats: true)
```
hosy, line 518: `lockTimer` needs to be checked if `nil` to prevent duplicate init of the timer. This was also a bug in the previous implementation. could lead in multiple timers.

```suggestion
if lockTimer == nil {
				lockTimer = Timer.scheduledTimer(timeInterval: 1, target: self, selector: #selector(self.updateLockCountdown), userInfo: nil, repeats: true)
			}
```

hosy (changes requested):

ownCloudAppShared/AppLock/AppLockManager.swift
felix-schwarz: Good catch! I recall timers stopping to fire under certain outside circumstances, so I think re-creating the timer here can't hurt. But the previous timer should definitely be stopped and dropped from memory. To achieve this, I added a `lockTimer?.invalidate()` before `lockTimer = Timer.scheduledTimer(…)`.

hosy (approved):

jesmrec:
Checks over the fix:

After three failed attempts and locking condition:

- [X] Time forward -> countdown restarted 
- [X] Time back -> countdown restarted 
- [X] Date forward -> countdown restarted 
- [X] Date back -> countdown restarted 

With open keyboard 

- [X] Time forward -> keyboard not locked 
- [X] Time back -> keyboard not locked 
- [X] Date forward -> keyboard not locked 
- [X] Date back -> keyboard not locked 

Checked other arbitrary cases, just to know if the brute force protection can be broken. Result is OK

From my side this is approved @hosy @felix-schwarz

jesmrec (approved):

PR #1404, opened 2024-10-17
[fix/xcstring-formatting] add step to normalize/beautify xcstrings files

felix-schwarz:
## Description
Adds a step to normalize/beautify xcstrings files after pulling from Transifex.


DeepDiver1975:
nice - we need this on ios-sdk as well .... but I assume this is know. thx

hosy (approved):
Awesome trick!

PR #1419, opened 2024-11-02
[feature/server-side-search] Add server-side search support

felix-schwarz:
## Description
This PR adds server-side search support to the app.

### TODO
- [x] add search scope backed by `OCSearchResult`
- [x] improve error and result handling in `OCSearchResult`
- [x] add KQL syntax generator to `OCQueryCondition` (similar to the SQL syntax generator)
- [x] handle offline state
- [ ] ~~possibly integrate/merge with account-wide search~~

## Related Issue

SDK part: https://github.com/owncloud/ios-sdk/pull/129

Name only | Contents only
----|----
Simulator Screenshot - iPhone 16 Pro Max - 2024-11-23 at 23 14 51 | Simulator Screenshot - iPhone 16 Pro Max - 2024-11-23 at 23 14 56

Name/Contents choice popup | Image files not containing `ownCloud`
----|----
Simulator Screenshot - iPhone 16 Pro Max - 2024-11-23 at 23 15 04 | Simulator Screenshot - iPhone 16 Pro Max - 2024-11-23 at 23 16 05

Search Settings | Without connection
----|----
Simulator Screenshot - iPhone 16 Pro Max - 2024-11-29 at 17 10 13 | Bildschirmfoto 2024-11-30 um 16 23 45

## Types of changes

## QA

Test plan:

Reports:

- [X] (1) Glitch in the `name+contents` https://github.com/owncloud/ios-app/pull/1419#issuecomment-2682145672


hosy (approved):

jesmrec:
Server issue open, since the result is not always correct: 

https://github.com/owncloud/ocis/issues/11060

jesmrec:
### (1)

There is an small glitch when selecting the option `name + contents`, misplacing the `Type` option in this way:

https://github.com/user-attachments/assets/bcb50a3c-bc8c-4f60-aea0-60bd2cc5d5c4

iPhone XR iOS18
`ced1e81901`

felix-schwarz:
@jesmrec (1) should be addressed as of 6181305.

jesmrec:
(1) fixed

jesmrec:
After QA checks in test plan, the feature is OK from iOS client perspective. But...

- Server does not retrieve the capability correctly https://github.com/owncloud/ocis/issues/11059
- Results retrieved are sometimes random and does not fit with special characters https://github.com/owncloud/ocis/issues/11060
- Searching by date does not work, server retrieves void responses https://github.com/owncloud/ocis/issues/11060#issuecomment-2684917665

Is it really worthy to move forward a client feature that is not working as expected in server side? client is the front-end and final users will see it as a problem in iOS side.

Should we wait for fixes in the server before releasing this one?

felix-schwarz:
@jesmrec While https://github.com/owncloud/ocis/issues/11059 should be addressed, the issues in https://github.com/owncloud/ocis/issues/11060 may be caused by the implementation or configuration details of the tokenizer. F.ex. searching by date worked on the instance I tested against at the time once I had figured out the server wanted a [golang-specific encoding of dates](https://github.com/owncloud/ios-sdk/blob/5d34fd82d56af3709b92a62caf5a76e1cfd67ab3/ownCloudSDK/Query/Condition/OCQueryCondition%2BKQLBuilder.m#L54).

Since the feature is ready from the iOS app side, I'd like to merge this PR to avoid the code to diverge too far.

If needed, we could temporarily disable server-side search until the issue(s) is/are fixed in an ocis release.

jesmrec:
> Since the feature is ready from the iOS app side, I'd like to merge this PR to avoid the code to diverge too far.

> If needed, we could temporarily disable server-side search in the app until the issue(s) is/are fixed in an ocis release.

Both work for me. 

CC @DeepDiver1975

jesmrec (approved):

PR #1420, opened 2024-11-06
watermark feature for iOS App

blicknix:
## Description
We implemented a watermark feature which can be enabled with MDM Tools.
Also we implemented a feature which allows to add a notification if a screenshot was made in the app.

## Motivation and Context
For audit reasons we want to add watermarks to each screen which shows personal data.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


DeepDiver1975:
@blicknix THX for your contribution - please sign the CLA as pointed out by the bot above. THX

felix-schwarz (changes requested):
Thanks for the contribution.

Please see my comments regarding specific code snippets.

On a more general note, I'd like to add that modularity is an important aspect in this project and additions like yours should be as modular as possible. Right now, your changes are woven in directly in many places that don't have any direct contextual relationship to the added features.

To fit in the project, I'd therefore like to ask you to change that structure:
- please create a new folder `ownCloudAppShared/User Interface/Watermark` and put your code additions in that folder as far as possible
- please put your code in files of their own rather than appending it to existing source code files
- the approach to add watermark views in every `UIViewController` subclass that might show content has a number of drawbacks and weaknesses:
   - new views controllers can emerge where this is not thought of or _can't_ be accounted for
   - new subclasses of already modified classes may/will add views on top
   - the approach generally fails if a modified view controller dynamically adds and removes views (as the Watermark will then no longer be the front most view and be covered by the views that were added after it)
   - the code changes spread to unrelated sourcecode, weakening modularity
- that said, `ThemeWindow` (containing all view controllers) or _especially_ `BrowserNavigationViewController` (containing all (content) view controllers) strike me as good, central places to apply a watermark without having to modify so many different view controllers and micromanage their views 
- _ideally_ the watermark feature should be removable from the app by deleting a folder and otherwise removing 1-3 lines of codes elsewhere

Other than that:
- please run SwiftLint and correct the warnings in your code (there are many, mostly on empty spaces)
- please read and accept the CLA

hosy (changes requested):
Some general thoughts about this feature:

This feature isn’t secure if it’s possible to open files from third-party apps. For instance, if you can open the same file via File Provider with another app or Files.app and take a screenshot, this feature isn’t secure.

If this feature is enabled, I suggest not delivering files via File Provider or showing an error message if the user wants to open or copy the file with a hint. Instead, open the ownCloud app to view the file. Additionally, it shouldn’t be possible to share files from the app.

Another idea to consider is an option instead of watermarking, where you can completely hide the content or disallow taking screenshots (perhaps with an overlay view or DRM content view).

hosy:
@blicknix please add a detailed description about the MDM parameters to the PR and add meta data to the class settings.

PR #1425, opened 2024-11-24
[feature/xcstrings-build-normalization] Normalize xcstrings on every build

felix-schwarz:
## Description
Adds a xcstrings normalization build step that normalizes the `xcstrings` files on every build. Since Xcode's format slightly diverges from the normalized version created by `jq`, this normalization step seeks to avoid large diffs for what is effectively few or no changes.

### DIFF
--- a/.github/workflows/pull-transifex.yml
+++ b/.github/workflows/pull-transifex.yml
@@ -67,6 +67,7 @@ jobs:
 
       - name: Cleanup download and ocstringstool
         run: |
+          tail -n 100 "ownCloud/Resources/Localizable.xcstrings"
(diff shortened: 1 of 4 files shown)


hosy (commented):
xcstrings files will be still reorderd on building.

felix-schwarz:
@hosy From what I saw, when I make a change, Xcode recreates the xcstrings files. On next build, `jq` reorders them, Xcode status displays that it's updating the xcstrings files, but then does not overwrite them. Granted, of course, `jq` is installed.

Can you reproduce it that way?

felix-schwarz:
Superseded/replaced by #1428

PR #1430, opened 2024-12-13
[feature/confidential-protection] Confidential Content Protection

hosy:
## Description
Confidential View and Screenshot Protection

This update introduces screenshot protection for confidentially marked views in the app and the share sheet extension. When a screenshot is taken, confidential views will not display their content in the screenshot.

To discourage taking photos of confidential views, these views now include a watermark displaying the account’s email and username. 

Screenshot protection can be enabled via an MDM parameter.

File or path names displayed in UI elements that are difficult to watermark are redacted after the third character. This ensures the user can identify the file or path name to some extent, but it remains unreadable for security.  

Additionally, to prevent sharing confidential data when screenshot protection or confidential view marking is enabled, the following features are disabled:
- File Provider access
- Shortcuts Intents
- System sharing dialog actions
- Copying files
- Text recognition on images

This behavior can be overridden using an MDM parameter; however, overriding it is not recommended.

### MDM Parameter

| MDM Parameter | Description |
|--------|--------|
| `confidential.allow-screenshots` | Controls whether screenshots are allowed or not. If not allowed confidential views will be marked as sensitive and are not visible in screenshots. |
| `confidential.mark-confidential-views` | Controls if views which contains sensitive content contains a watermark or not. |
| `confidential.allow-overwrite-confidential-mdm-settings` | Controls if confidential related MDM settings can be overwritten. |

### Confidential View

| File List | Content View | Spaces | File Provider | Shortcuts|
|--------|--------|--------|--------|--------|
| Simulator Screenshot - iPhone 16 Plus - 2024-12-16 at 16 05 16 | Simulator Screenshot - iPhone 16 Plus - 2024-12-16 at 14 46 09 | Simulator Screenshot - iPhone 16 Plus - 2024-12-16 at 14 18 19 | Simulator Screenshot - iPhone 16 Plus - 2024-12-16 at 14 18 51 | Simulator Screenshot - iPhone 16 Plus - 2024-12-16 at 14 19 39 | 

### Screenshot Protection

| App | Share Extension |
|--------|--------|
| IMG_0089 | IMG_0088 | 

## Testing

Add at least one of the following MDM parameters to the `Branding.plist`:

| Key | Type | Value |
|--------|--------|--------|
| `confidential.allow-screenshots` | `Boolean` | `NO` |
| `confidential.mark-confidential-views` | `Boolean` | `YES` | 

If you want to allow File Provider access or Actions set the key `confidential.allow-overwrite-confidential-mdm-settings` to `Boolean` `YES`.

Screenshot protection is only working on a physical device!

What to test:

- [ ] Light/dark appearance 
- [ ] File Provider access
- [ ] Share sheet
- [ ] Shortcuts
- [ ] All content and editing views
- [ ] Device rotation
- [ ] Multi window environment (iPad)
- [ ] Menu sheets
- [ ] Sidebar
- [ ] Screenshot protection should not show any file names, folders or document content
- [ ] Confidential views (file names, folders, or document content) should contain watermarks, or names should be redacted. 

QA:

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/iOS/Executions/Version%2012.4/Content%20Protection.md

Reports:

- [X] (1) Space name not redacted https://github.com/owncloud/ios-app/pull/1430#issuecomment-2604837287 [FIXED]
- [X] (2) Text to be redacted in spaces' names https://github.com/owncloud/ios-app/pull/1430#issuecomment-2604851540 [WONT FIX ftm]
- [X] (3) Not able to open the fp with the proper setup https://github.com/owncloud/ios-app/pull/1430#issuecomment-2605291563 [FIXED]
- [x] (4) problems in markup feature https://github.com/owncloud/ios-app/pull/1430#issuecomment-2607819318 [FIXED]

## Types of changes

## Checklist:


DeepDiver1975:
To fix CI I suggest to use this action https://github.com/marketplace/actions/mxcl-xcodebuild

In a second step let's define different combinations which we want to build and test in the future

jesmrec:
First, preliminary and exploratory testing, with commit `35fa8a981`

Assumed: protection only affects to:

- Files (name and other parameters)
- Folders name in navigation
- File contents
- Spaces (name and other parameters)

is that right? 

### 1. Screenshot protection:

Using the following set up:
```
<key>confidential.allow-screenshots</key>
<false/>
```

- Side menu is screen-shooted (shouldn't?). I caught a video, the content is blank due to the protection. I took two screenshots: the first one over the list of files is blank (correct ✅ ), the second one over the side menu shows the info

https://github.com/user-attachments/assets/957fec90-731b-42ae-bbc1-c6ba782e5361

are the username and the account sensible information?

the issue is even more visible in iPad, where the sidebar menu is always in foreground.

- It's not posible to share content with other users or by link. It's posible in the web and in Android app. That causes a regression.

### 2. Watermarking

no issues detected at this point.

### 3. `confidential.allow-overwrite-confidential-mdm-settings`

`true`: I wasn't able to test the correct access to the FP's because i always get the following error:

Screenshot 2024-12-19 at 10 35 10

`false`: `FP access has been disabled by the administrator`  ✅ 

that was just a first preliminary testing, final testing will be done when CR passed. This first step helped to design the feature test plan.

hosy:
@hosy There is a redraw error when switching the collection view layout from list to grid or other formats. This issue needs to be fixed.

ownCloudAppShared/User Interface/Confidential/SecureTextField.swift
```diff
@@ -0,0 +1,57 @@
...
+		
+		if let secureView = self.subviews.filter({ subview in
+			type(of: subview).description().contains("CanvasView")
+		}).first {
+			secureView.translatesAutoresizingMaskIntoConstraints = false
+			secureView.isUserInteractionEnabled = true
+			return secureView
+		}
+		
+		let view = UIView()
```
felix-schwarz, line 49: If this line of code is reached, no screenshot-protecting CanvasView could be found and there will not be any screenshot protection.

Returning a `UIView` here will keep the view structure intact, but nothing will prevent a screenshot from being taken.

Something to discuss: should the app have this fallback and continue to work even if the protection no longer is in place? Should the app crash? Should that decision be determined by us (and hardcoded) - or be configurable?

ownCloudAppShared/User Interface/More/MoreViewHeader.swift
```diff
@@ -86,6 +86,12 @@ open class MoreViewHeader: UIView {
 
 	private func render() {
 		cssSelectors = [.more, .header]
+		
+		let secureView = SecureTextField().secureContainerView
```
felix-schwarz, line 90: The suggestions about unifying the code to inject a secure container that I made in `BrowserNavigationViewController.swift` would also apply here.

ownCloudAppShared/Client/View Controllers/Location Picker/ClientLocationPicker.swift
```diff
@@ -172,6 +172,9 @@ public class ClientLocationPicker : NSObject {
 			}
 
 			allowedLocationFilter = { (targetLocation, context) in
+				OnMainThread {
+					self.headerView?.secureView(core: context?.core)
```
felix-schwarz, line 176: The use of `self` here might create a retain-loop.
Also view management code shouldn't be placed into this closure, if only because it could be called more than once and a new watermark view would be added on every call.

ownCloudAppShared/Client/Sharing/ShareViewController.swift
```diff
@@ -259,7 +259,18 @@ open class ShareViewController: CollectionViewController, SearchViewControllerDe
 
 	open override func viewDidLoad() {
 		super.viewDidLoad()
-
+		
+		let secureView = SecureTextField().secureContainerView
```
felix-schwarz, line 263: The suggestions about unifying the code to inject a secure container that I made in `BrowserNavigationViewController.swift` would also apply here.

ownCloudAppShared/Client/Collection Views/CollectionViewController.swift
```diff
@@ -167,10 +167,14 @@ open class CollectionViewController: UIViewController, UICollectionViewDelegate,
 	public override func loadView() {
 		super.loadView()
 
+		let secureView = SecureTextField().secureContainerView
```
felix-schwarz, line 170: The suggestions about unifying the code to inject a secure container that I made in `BrowserNavigationViewController.swift` would also apply here.

felix-schwarz (changes requested):

ownCloud/UI Elements/ImageScrollView.swift
```diff
@@ -196,6 +197,7 @@ extension ImageScrollView {
 	}
 
 	var imageInteractionsAllowed: Bool {
+		guard ConfidentialManager.shared.allowOverwriteConfidentialMDMSettings else { return false }
```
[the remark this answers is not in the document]
hosy: But does this mean we do not have an MDM keypath for setting it via MDM or the Branding.plist?

ownCloudAppFramework/File Provider Services/OCFileProviderSettings.m
```diff
@@ -46,6 +47,10 @@ + (OCClassSettingsMetadataCollection)classSettingsMetadata
 
 + (BOOL)browseable
 {
+	if ([[ConfidentialManager sharedConfidentialManager] allowOverwriteConfidentialMDMSettings] == false) {
```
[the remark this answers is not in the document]
hosy: But does this mean we do not have an MDM keypath for setting it via MDM or the Branding.plist?

ownCloud Intents/IntentSettings.swift
```diff
@@ -29,6 +30,10 @@ class IntentSettings: NSObject {
 	}
 
 	var isEnabled : Bool {
+		if !ConfidentialManager.shared.allowOverwriteConfidentialMDMSettings {
```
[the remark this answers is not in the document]
hosy: But does this mean we do not have an MDM keypath for setting it via MDM or the Branding.plist?

ownCloudAppShared/User Interface/More/MoreViewHeader.swift
hosy: I cannot get it working here, please can you have a look?

ownCloudAppShared/Client/Collection Views/CollectionViewController.swift
hosy: I cannot get it working here, please can you have a look?

ownCloudAppShared/User Interface/Confidential/SecureTextField.swift
hosy: Needs to be discussed. What do you think. Force crash?

hosy: I have added a force quit feature to prevent the application from reaching this point. However, it would be insecure if screenshot protection were not implemented.

felix-schwarz (approved):
I've implemented the remaining open findings and suggestions.

jesmrec:
is that QA-ready? @felix-schwarz @hosy

felix-schwarz:
@jesmrec Yes it is. I was just busy updating the iOS Team project and would've told you in a minute :)

jesmrec:
During QA phase it's assumed that protection only affects to:

- Files (name and other parameters)
- Folders name in navigation
- File contents
- Spaces (name and other parameters)

jesmrec:
### (1)

```
<key>confidential.mark-confidential-views</key>
<true/>
```
with that setup, i notice that the space name is not redacted here:

IMG_2542

it should be also redacted as the other red-rounded appearances.

Could you take a look @hosy @felix-schwarz 

iPhoneXR iOS18
``d3287bf0d`

jesmrec:
### (2) (question)

This is maybe a pretty corner case, or a feature (not a bug)

First three letters of space name are visible and the other letters are redacted. But, if the space name has just three letters (for example, an subject or name abbreviation), it won't be redacted at all. 

Should space names to be completely redacted? maybe reducing to two letters of visibilty will avoid more cases of short words or abbreviations that could be infered.

jesmrec:
@hosy, thanks for fixing (1). You gave me the clue to find other visible file names:

- The "Move" folder picker shows also the file name:

Screenshot 2025-01-21 at 17 45 58

- The "Sharing view", "Create link" and "Invite" shows the file name:

Screenshot 2025-01-21 at 17 46 46

DeepDiver1975:
conflict on submodulle .... @hosy FYI

jesmrec:
### (3)

with the following setup

```
<key>confidential.allow-overwrite-confidential-mdm-settings</key>
<true/>
```

i get the following error, not being able to open the fp's content:

https://github.com/user-attachments/assets/5ebf0501-6fbe-469c-80e9-9450b98da314

iPhoneXR iOS18
`9df71cfd1`

jesmrec:
@hosy @felix-schwarz in one of our meetings we talked about subtitling videos to avoid recording the audio under the `mark-confidential-view` set up. Was that finally implemented? any detail about it? i could not reproduce that.

DeepDiver1975:
> Was that finally implemented?

Only a POC was implemented but we left this out of this pull request to not overload this feature

jesmrec:
### (4)

About shortcuts:

As stated before:

> Additionally, to prevent sharing confidential data when screenshot protection or confidential view marking is enabled, the following features are disabled:

> ...
Shortcuts Intents
...

with the following setup:

```
<dict>
	<key>confidential.allow-screenshots</key>
	<false/>
	<key>confidential.mark-confidential-views</key>
	<false/>
</dict>
```

i am able to:

- Create and execute shortcuts from the `Shortcuts` app, once the feature is enabled in the `Settings`
- Create and execute shortcuts over files, folders and URLs inside the app

i was not sure which kind of shortcuts are suposed not to work anymore, but both work. Please, put light in front of my eyes to clarify the correct behaviour.

iPhoneXR  iOS 18.2
iPadOS iOS 17.4
`9df71cfd`

jesmrec:
### (4)

1. Open any image or PDF file with the `Markup` option

Current:

here i found some problems:

- files are not opened (seems to be a regression, so, feel free to fix here, it could be delayed to the final release regression test). I can reproduce this in iPhone (iOS18), but not in iPadOS (iOS17) where it works fine.

- In iPadOS17, the image is visible in markup view. But, taking a screenshot with `confidential.allow-screenshots` as false, shows the image in the screenshot (it shouldn't).

- Taking an screenshot, the file name on the top is visible when `confidential.allow-screenshots` is false

- With `mark-confidential-views` enabled, file name on the top is not redacted.

- If the image/pdf file is not downloaded, a download dialog appears on the bottom with the file name, that is not redacted either.

- With protection features enabled, the `open in` icon is shown on the top of markup screen. Open in is a banned operation

Could it be a good idea to remove the `Markup` as well as `Copy` or `Open in` ? on the one hand, it prevents all that errors, on the other hand, as payment feature, it's bad for the user to be prevented of a feature that was paid.


iPhoneXR iOS18
iPadAir2 , iOS 17.4
`9df71cfd1`

felix-schwarz:
> @hosy, thanks for fixing (1). You gave me the clue to find other visible file names:
> 
> * The "Move" folder picker shows also the file name:
> 
> Screenshot 2025-01-21 at 17 45 58
> * The "Sharing view", "Create link" and "Invite" shows the file name:
> 
> Screenshot 2025-01-21 at 17 46 46

I made sure these are also redacted as of https://github.com/owncloud/ios-app/pull/1430/commits/6f4101306f42e279096c86037c89039255b8ebde, but I'm wondering if this is really necessary. To quote the original description for the redact item name feature: "[…] UI elements that are difficult to watermark are redacted after the third character".

But since those areas already are watermarked, should they really also be redacted?

felix-schwarz:
Re (3): assuming this happened on an actual device, I'd think it's a typical memory limit crash & error message that's responsible for this. Please try again with the latest commits, which now include the latest mem-optimizations.

jesmrec:
> @jesmrec Re (3): assuming this happened on an actual device, I'd think it's a typical memory limit crash & error message that's responsible for this. Please try again with the latest commits, which now include the latest mem-optimizations.

(3) is fixed 👍 

(1) also fixed

(2) is a question, ftm we can keep 3 letters in the redacted names

only (4) is pending.

After that, i will do a second QA round to assure nothing was broken.

felix-schwarz:
Re (4): the markup feature is delivered by the `QLPreviewController` system component, which is implemented as remote UI (=> kind of like screen sharing the contents from another process, and sending input events to that other process). This can be seen in the view debugger in the view tree:
Bildschirmfoto 2025-01-24 um 17 31 19

Unfortunately, all attempts to add the protection features around it eventually failed. I documented it with this code commend:
```
The markup action could not be modified to implement protection mechanisms -
not even on the CALayer level - without interaction with the system-provided
view breaking and becoming unusable in different ways. A possible reason for
this is that the markup feature is delivered by the OS as Remote UI (visible
as QLRemoteUIHostViewController in the view hierarchy) and that otherwise working
approaches to "passing through" events are not usable with these.
```

I therefore added the markup action to the list of actions to disallow if protection features are used. It simply doesn't currently seem technically possible to offer that action under the protection umbrella.

From my POV this PR is now ready for a final QA round @jesmrec.

jesmrec:
after latest fix, markup feature was disabled if one of the new parameters or both, are disabled. So, all issues reported in (4) are out of scope from now on.

jesmrec:
Approved from my side, ready to go!

jesmrec (approved):

Issue #1422, opened 2024-11-08
[FEATURE REQUEST] Sharing NG on iOS


dj4oC:
Please scope to implement Sharing NG for iOS

\cc @DeepDiver1975 @hosy @felix-schwarz @marcelmockkw

jesmrec:
Done

Issue #1421, opened 2024-11-08
[FEATURE REQUEST] Space Management with iOS


dj4oC:
As a User I want to be able to perform Space Management on my iOS Device.

## Prio 1
- [ ] Create Space
- [ ] Rename Space
- [ ] Show details of the space
  - [ ] Show Managers
  - [ ] Show Quota
  - [ ] Show Size
- [ ] Members
  - [ ] Show members
  - [ ] Add members
  - [ ] Remove members
  - [ ] Change membership role
  - [ ] Show Access expire
  - [ ] Set Access expire
  - [ ] Remove Access expire
- [ ] Sharing via Link 
  - [ ] Set/Remove Public Link
  - [ ] Set/Change/Remove Expiration Date
  - [ ] Set/Change/Remove Password (consider Password Policy)
  - [ ] Set/Change Link Type
  - [ ] Copy Link & Copy Password

## Prio 2
- [ ] Duplicate Space
- [ ] Edit description
- [ ] Edit subtitle
- [ ] Edit image
- [ ] Set icon

## Prio 3
- [ ] Disable Space
- [ ] Delete / undelete disabled space
- [ ] Show Activities

\cc @DeepDiver1975  @hosy @felix-schwarz @marcelmockkw

felix-schwarz:
Implemented in #1435.

PR #1435, opened 2024-12-29
[feature/sharing-ng] Support for new ocis spaces/sharing capabilities

felix-schwarz:
## Description
App support for new spaces and sharing APIs.

## Screenshots
. | . | . | . | .
--|--|--|--|--
IMG_D1666B1433FF-1 | Screenshot 2025-02-27 at 11 33 51 PM | IMG_0071 | IMG_0070 | IMG_0069
IMG_0068 | IMG_0067 | IMG_0066 | IMG_7F6066237BD9-1 | Screenshot 2025-02-27 at 11 28 16 PM

## Related Issue
https://github.com/owncloud/ios-app/issues/1421
https://github.com/owncloud/ios-app/issues/1422

SDK portion: https://github.com/owncloud/ios-sdk/pull/135

### Prio 1
- [x] Create Space
- [x] Rename Space
- [x] Show details of the space
  - [x] Show Managers
  - [x] Show Quota
  - [x] Show Size
- [x] Members
  - [x] Show members
  - [x] Add members
  - [x] Remove members
  - [x] Change membership role
  - [x] Show Access expire
  - [x] Set Access expire
  - [x] Remove Access expire
- [x] Sharing via Link 
  - [x] Set/Remove Public Link
  - [x] Set/Change/Remove Expiration Date
  - [x] Set/Change/Remove Password (consider Password Policy)
  - [x] Set/Change Link Type
  - [x] Copy Link & Copy Password

### Prio 2
- [ ] ~~Duplicate Space~~ --> needs proper server-side API (currently would need client to create space and copy contents over)
- [x] Edit description
- [x] Edit subtitle
- [x] Edit image
- [x] Set icon

### Prio 3
- [x] Disable Space
- [x] Delete / undelete disabled space
- [ ] ~~Show Activities~~ --> outside scope of this PR, feature of its own

## Types of changes

### QA

Test Plan space management: https://github.com/owncloud/QA/blob/master/Mobile/iOS/Executions/Version%2012.5/Spaces%20Management.md

Test Plan sharing:
https://github.com/owncloud/QA/blob/master/Mobile/iOS/Executions/Version%2012.5/Sharing.md

Reports:

- [x] (1) Soft keyboard overlaps the `Create` button https://github.com/owncloud/ios-app/pull/1435#issuecomment-2754579837 [FIXED]
- [x] (2) 3-dot-button after reopening the app https://github.com/owncloud/ios-app/pull/1435#issuecomment-2754590888  [FIXED]
- [x] (3) Suggestion: add cleaner to name and subtitle https://github.com/owncloud/ios-app/pull/1435#issuecomment-2765844182 [DONE]
- [x] (4) Quota fixes https://github.com/owncloud/ios-app/pull/1435#issuecomment-2765946300 (DONE - MOVED WEB]
- [x] (5) Disable space does not work in the edit space menu https://github.com/owncloud/ios-app/pull/1435#issuecomment-2768337708 [FIXED]
- [x] (6) Edit description not showing icon in iPad https://github.com/owncloud/ios-app/pull/1435#issuecomment-2768349449  [FIXED]
- [x] (7) Unique manager of a space https://github.com/owncloud/ios-app/pull/1435#issuecomment-2769075643 [FIXED]
- [x] (8) Cleaner to be hidden in `Details` https://github.com/owncloud/ios-app/pull/1435#issuecomment-2772329590 [FIXED]
- [ ] (9) Beta endpoints https://github.com/owncloud/ios-app/pull/1435#issuecomment-2782455897
- [X] (10) Share with yourself, improve error handling https://github.com/owncloud/ios-app/pull/1435#issuecomment-2782471771 [FIXED]
- [X] (11) Bad error when sharing with an existing sharee https://github.com/owncloud/ios-app/pull/1435#issuecomment-2782521393 [FIXED]
- [X] (12) Cleaner in remove link name https://github.com/owncloud/ios-app/pull/1435#issuecomment-2782956322 [FIXED]
- [X] (13) Incorrect date in date picker https://github.com/owncloud/ios-app/pull/1435#issuecomment-2789210602 [WONT FIX]


hosy (approved):
Well done!

jesmrec:
### (1)

Software keyboard collapses with the `Create` button in the create space view. No way to submit the new space:

https://github.com/user-attachments/assets/84e7793b-9153-45fa-a465-932d9208b46a

iPhone XR iOS18.2
`b0516fdeb`

jesmrec:
### (2)

1. Open the `Spaces` view with an user with space permissions -> 3-dot-button on the top-right corner available
2. Kill the app
3. Open he app -> `Spaces` view in foreground

**Current:**

3-dot-button on the top-right corner is missing

**Expected:**

3-dot-button on the top-right corner available

iPhone XR iOS18.2
`b0516fdeb`

felix-schwarz:
@jesmrec I fixed (1) and (2) with the latest commits. Please have another look.

jesmrec:
(1) and (2) fixed

jesmrec:
### (3) (suggestion)

In the `Name` and `Subtitle` sections of the creation and edit space view, it'd be useful to add a cleaner button to remove the content in one click, like the one in the `Create folder` or similar:

IMG_2653

jesmrec:
### (4)

About the space's limit space. Check the difference between iOS app and web

| iOS | web |
|:-:|:-:|
|IMG_2654|Screenshot 2025-03-31 at 12 58 10|

Web's approach has just some default options in GB. It's true, you can type there other values, but always are less than 1PB:

Screenshot 2025-03-31 at 13 00 48

It should not be allowed in the iOS app to enter values higher than 1PB

OTOH, i realised that any option sent by the app with something that is not `GB`, sets in the web `No Restriction`. 

App sends:

```
{
    "name": "Test",
    "quota": {
        "total": 10000000000000
    }
}
```

server returns:

```
 "quota": {
        "remaining": 9999999952721,
        "state": "normal",
        "total": 10000000000000,
        "used": 47279
    },
```

i'm pretty sure this is a problem in web/backend side, just commenting in case i miss something or any fix or further action is required from our side.

`a9b14f294`
iPhone XR iOS18.2

jesmrec:
### (5)

1. Open `Spaces` list
2. In any space with admin permissions, click on the 3-dot-button
3. `Edit space`
4. `Disable space`

**Current:**

Nothing happens, it's like the button has no action attached

**Expected:**

Space to be disabled

`a9b14f294`
iPhone XR iOS18.2

jesmrec:
### (6)

The `Edit description` option does not show the icon in iPadOS 17,6, it does it iPhone.

|||
|:-:|:-:|
|Screenshot 2025-04-01 at 08 47 19|Screenshot 2025-04-01 at 08 47 41|

iPadOS 17.6
`a9b14f294`

felix-schwarz:
@jesmrec (4) is now addressed as well. If a higher quota is entered, it shows an alert when trying to create or save the space, asking the user to set a lower quota. Since this is a rare edge case, I think that solution should be sufficient in practice.

jesmrec:
(3), (4), (5) and (6) fixed.

I'll open issue in the web repo with the mentioned issue

jesmrec:
### (7) (improvements)

1. Create space (no matter if web or app)
2. Add some members as viewer or edit (not manager)
3. In the list of members, click on the space owner itself (it is there as manager)

For a better UX, if there is only one manager:

- The owner row in the list of members would not be clickable in case of no more managers are available, because it's not allowed to change the permissions for the unique manager.
- The `Expiration date` option should not be available for the unique manager.
- The unshare option should not be available for the unique manager. In this case, it shows a not-friendly error message

`a81c8014f`
iPhone XR 18.2

jesmrec:
Regarding (4), the following issue was created in web repo:

https://github.com/owncloud/web/issues/12388

jesmrec:
### (8)

About the membership expiration date, i notice that is like hardcoded as 8 days from the creation date. Would it be better to add a datepicker as the web does? 

i can't tell an specific use case but, it's not very flexible

`a81c8014f`
iPhone XR 18.2

jesmrec:
### (9)

About this [fix](https://github.com/owncloud/ios-app/pull/1435#issuecomment-2765844182)

The (x) is correct, but, there is one case in which it should be hidden: the `Details` view for spaces with no `Manager` permission. In such case, the title and subtitle are not editable, therefore the (x) has no action and effect.

`a81c8014f`
iPhone XR 18.2

felix-schwarz:
@jesmrec I've addressed (8) by making the text fields only show the "x" while they have focus (and content to clear) with the latest commit.

Regarding (7):

> * The owner row in the list of members would not be clickable in case of no more available managers, because it's not allowed to change the permissions for the unique manager.

I addressed that with the latest commit.

> * The `Expiration date` option should not be available for the unique manager.

I agree, but this should also be blocked on the server side. The server already prevents the removal of the last manager, but doesn't prevent this. Please let me know if I should file an issue.

I addressed this by not allowing bringing up the editing view for a solo manager.

> * The unshare option should not be available for the unique manager. In this case, it shows a not-friendly error message

I fixed this as well (via https://github.com/owncloud/ios-sdk/pull/141). You can still test this by 
- creating a space with two managers
- starting editing one of the manager members in the app
- remove the other manager via the web client
- with the editing view still open, using the Unshare option in the app

It should now show the error message returned from the server.

jesmrec:
> I agree, but this should also be blocked on the server side. The server already prevents the removal of the last manager, but doesn't prevent this. Please let me know if I should file an issue.

Server side does not allow to add an expiration date to the solo manager (neither iOS does), but, it does not allow to remove it if it was created before because a different permission level or other existing managers. Yeah, i think this should be fixed in ocis or web.

felix-schwarz:
Thanks!

I filed an ocis issue regarding being able to set an expiration date on the solo manager: https://github.com/owncloud/ocis/issues/11205

jesmrec:
### (9) 

i noticed that several requests are done over `v1beta1` endpoints:

https://github.com/owncloud/ios-sdk/blob/5399f6b7b57fa0dcf68acf080805105086732af5/ownCloudSDK/Connection/OCConnection.m#L139

(and the one in the following line)

is this really optimal/trustable? why is an endpoint in beta version included?

jesmrec:
### (10)

1. Select `Sharing` over any item in the list
2. `Invite`
3. Type the name of the current user
4. Choose any permission level
5. `Invite`

**Current:**

IMG_2655

**Expected:**

The error message is correct but not very friendly. Would it be posible to prevent the part of `bad request...` since this is pretty technical and not of the interest of final users. The best solution would be to prevent to show the current user in the list of potential sharees. Feasible?

iPhone XR V18.4
`0a09c3e5`

jesmrec:
### (11)

1. Select `Sharing` over any item in the list
2. `Invite`
3. Type the name of any user
4. Choose any permission level
5. Invite again the same user

**Current:**

IMG_2656

**Expected:**

Friendly error message

iPhone XR V18.4
`0a09c3e5`

felix-schwarz:
Re (9): these are the same endpoints used by ocis/web at the time. I'm not sure whether `v1beta1` went out of beta at some point.

@kobergj Do you know if the following endpoints have alternatives that are "out of beta"?
- `/graph/v1beta1/drives`
- `/graph/v1beta1/roleManagement/permissions/roleDefinitions`

hosy (approved):

felix-schwarz:
@jesmrec I fixed (10), (11) and (12) with the latest commits. I addressed issue (11) and (12) by filtering the results of the recipient search, so that the logged in user can't select him-/herself anymore. Also, those users or groups to whom the item is already shared no longer turn up in recipient search results either.

jesmrec:
(10), (11) and (12) fixed

jesmrec:
### (13)

1. Create a new link
2. Add an expiration date, i set one in the next month (12 May 2025)
3. Save the link
4. Open the link

**Current:**

Displayed date is `13 May 2025`

**Expected:**

Displayed date is `12 May 2025`

Checking the requests, the `POST` request sends in the payload:

```
{
    "expirationDateTime": "2025-05-12T10:25:00.000Z",
    "password": "...",
    "type": "view"
}

```

server responses with:

```
{
    "createdDateTime": "2025-04-09T10:26:35.254560161Z",
    "expirationDateTime": "2025-05-12T23:59:59Z",
    "hasPassword": true,
    "id": "qzVZkbaXjDeHJzD",
    "link": {
        "@libre.graph.displayName": "",
        "@libre.graph.quickLink": false,
        "preventsDownload": false,
        "type": "view",
        "webUrl": "https://192.168.1.66:9200/s/TavzFJdukShgxXH"
    }
}
```

both iOS client and server are under the same LAN, and for sure, under the same TZ. My guess is: whereas the server returns `"2025-05-12T23:59:59Z",`,  the iOS clients takes this as the following day.

Scenario footage (if helps):

https://github.com/user-attachments/assets/c11bc24b-4fdd-42bc-a6f5-869baf8728c2

is this fixable or just a race condition?

iPhone XR v18.4
`0c762b633`

felix-schwarz:
@jesmrec I looked into (13) and my findings confirm your observations:

- creation of the link with:
  ```
  {"type":"view","expirationDateTime":"2025-04-17T08:36:25.581Z","password":"…"}
  ```

- retrieval of the link as JSON:
  ```
  "expirationDateTime":"2025-04-17T23:59:59Z"
  ```
  and as XML:
  ```
  …<expiration>2025-04-17 23:59:59</expiration>…
  ```

As per [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#:~:text=Z%20is%20the%20zone%20designator,30Z%22%20or%20%22T0930Z%22.), the timezone indicated by `Z` is `UTC`.

Translated into my local timezone, that is 1:59 the following day, as becomes visible in the debugger:
L expirationDate = (Foundation Date) 2025-04-17 235959 UTC

So what happens here is that
- the client sends a correctly time-zoned expiration date to the server
- the server strips the time and just uses 23:59:59 instead
- the timezone is retained / UTC is still used
- converted from UTC to local timezones, the expiration date can shift into the next day

Whether or not the server expires the link at 23:59:59 UTC or 23:59:59 local time I haven't checked and don't know. If it expires at 23:59:59 local time, the expiration date returned by the server would be incorrect.

From my POV, however, the client works correctly, sending the server a correct UTC date/time and also showing the UTC date/time returned by the server correctly in local time.

jesmrec:
thanks @felix-schwarz. Researching a little bit about it, i will open an issue in ocis repo because the difference in the expiration date's handling in links and shares:

In Links, server returns:

`"expirationDateTime": "2025-05-12T23:59:59Z",` (always last minute of the day, in our TZ it will do +2, jumping to the next day as reported)

In shares, server returns:

`"expirationDateTime": "2025-05-12T07:37:00Z",` (same timestamp as the share creation, not the last minute of the expiration date)

both could be valid, but also both should behave the same.

That means, (13) is a won't fix

jesmrec:
This is approved on my side. Good job!! 💯

jesmrec (approved):

Issue #1334, opened 2024-04-05
[a11y] 11.1.4.3 Contrast (minimum)


tschiebel:
audit: https://infinite.owncloud.com/s/qSvYucPltEYUhhc

https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de
https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf

Note 1: For texts (also in the focused state) under 18 points (or
bold text below 14 point), the contrast ratio between foreground and background color should be at least
background color should be at least 4.5:1. For large fonts, the contrast ratio should be
should be 3:1 or greater.
Note 2: Since the contrast measurements in the test were carried out using screenshots
the test results may differ from the actual color values. It is
possible that the colors are displayed differently on the screenshot than on the
screen of the device. For this reason, the test results should be regarded as guide values.
should be regarded as guidelines.

### Personal Space

- ~~[ ] Both the size information and the time information on the page do not meet the contrast do not meet the contrast requirements for texts, as the contrast ratio is currently
is currently well below 4.5:1.
The text at the bottom of the page, which shows the number of directories and
files also does not meet the contrast requirements for text.~~
![image](https://github.com/owncloud/ios-app/assets/132444655/ed704054-a10c-459d-a483-d10d11d89cc4)

**-> The colors for the brighter text and the icons are using standard system-provided colors used by iOS throughout the system. Enabling the option `Increase Contrast` (`Kontrast erhöhen`) in System Settings also increases the contrast here:**

Without `Increase Contrast` | With `Increase Contrast`
---|----
![Image](https://github.com/user-attachments/assets/4be989ad-5fe7-4ae4-a84a-fa41b95b5c9d) | ![Image](https://github.com/user-attachments/assets/8351a218-4bb7-4c65-9dcc-3ff5425a5d97)

- ~~[ ] The button labels (see screenshot) do not meet the contrast requirements for text.
contrast requirements for text.~~
![image](https://github.com/owncloud/ios-app/assets/132444655/0646228a-8ca2-496e-a759-a5aba9e744df)

**-> The color for the button label is using a standard system-provided color used by iOS throughout the system. Enabling the option `Increase Contrast` (`Kontrast erhöhen`) in System Settings also increases the contrast here:**

Without `Increase Contrast` | With `Increase Contrast`
---|----
[...]

PR #1451, opened 2025-03-10
[fix/a11y-1334] Add Spaces text background color if Increase Contrast is on

felix-schwarz:
## Description
- ThemeCollection: remove superfluous code cell
- DriveListCell: hide subtitle label if subtitle is not set or empty
- DriveHeaderCell: make sensitive to high contrast mode, adding a dark background if detected or changed to

## Related Issue
#1334 

## Types of changes

### DIFF
--- a/ownCloudAppShared/Client/Collection Views/Cells/DriveHeaderCell.swift
+++ b/ownCloudAppShared/Client/Collection Views/Cells/DriveHeaderCell.swift
@@ -38,6 +38,34 @@ class DriveHeaderCell: DriveListCell {
 		coverObservation?.invalidate()
 	}
 
+	override init(frame: CGRect) {
+		super.init(frame: frame)
+		setupHighContrastMode()
+		updateForContrastMode()
+	}
+	
+	@MainActor required init?(coder: NSCoder) {
+		fatalError("init(coder:) has not been implemented")
+	}
+
+	var contrastObservation: NSObjectProtocol?
+
+	func setupHighContrastMode() {
+		contrastObservation = NotificationCenter.default.addObserver(forName: UIAccessibility.darkerSystemColorsStatusDidChangeNotification, object: nil, queue: .main, using: { [weak self] _ in
+			self?.updateForContrastMode()
+		})
+	}
+
+	func updateForContrastMode() {
+		if UIAccessibility.isDarkerSystemColorsEnabled {
+			titleLabel.backgroundColor = UIColor(white: 0, alpha: 0.8)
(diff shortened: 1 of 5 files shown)


ownCloudAppShared/Client/Collection Views/Cells/DriveHeaderCell.swift
```diff
@@ -38,6 +38,34 @@ class DriveHeaderCell: DriveListCell {
...
+	}
+	
+	@MainActor required init?(coder: NSCoder) {
+		fatalError("init(coder:) has not been implemented")
+	}
+
+	var contrastObservation: NSObjectProtocol?
+
+	func setupHighContrastMode() {
+		contrastObservation = NotificationCenter.default.addObserver(forName: UIAccessibility.darkerSystemColorsStatusDidChangeNotification, object: nil, queue: .main, using: { [weak self] _ in
```
hosy, line 54: remove observer is missing on deinit

hosy (changes requested):

ownCloudAppShared/Client/Collection Views/Cells/DriveHeaderCell.swift
felix-schwarz: This is no longer necessary as observers added via this method are automatically removed once the returned `observation` object is deallocated. As the returned `contrastObservation` object is released and deallocated at the time `DriveHeaderCell` is deinited/deallocated, the observation is automatically removed.

hosy (approved):

felix-schwarz:
@jesmrec Thanks for the analysis. I have since made changes to increase contrast significantly for those elements when High Contrast is enabled in System Settings. Please have another look.

jesmrec:
Thanks for the update @felix-schwarz. Added a column in the table above with the contrast checks in the latest commit `47eba9a9`. Every text over white background is already in. 

Just one exception: the green/red buttons in the shared with me view (`Decline` / `Accept` ). These ones are still below 4,5. Is there something else we could do?

felix-schwarz:
@jesmrec That contrast issue should be fixed now with the latest commit. The buttons now use a different style if High Contrast mode is enabled at the time the view is invoked by the user.

jesmrec:
> @jesmrec That contrast issue should be fixed now with the latest commit. The buttons now use a different style if High Contrast mode is enabled at the time the view is invoked by the user

Thanks for the fix. We can guess that the user is not switching on and off the "increase contrast" option steadily, so, it's a good solution. 

This is approved on my side.

jesmrec (approved):

Issue #1339, opened 2024-04-05
[a11y] 11.2.5.1 Pointer gestures


tschiebel:
audit: https://infinite.owncloud.com/s/qSvYucPltEYUhhc

https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de
https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf

### Image Viewer

- [x] The carousel elements (images) can only be moved using swipe gestures. There is
There is no alternative input option using simple pointer gestures.

**Fixed in #1453 by adding previous/next buttons that become visible when hovered with a trackpad or mouse pointer.**

PR #1453, opened 2025-04-03
[fix/a11y-1339-carousel] Add previous/next buttons to image viewer carousel

felix-schwarz:
## Description
Adds previous/next buttons to the image viewer / carousel to provide an simpler pointer alternative to swipe gestures (both with a finger on the display - or two fingers on the trackpad).

## Related Issue
#1339 

Screenshot 2025-04-03 at 10 18 23 PM

## Types of changes

### DIFF
--- a/ownCloud/Client/Viewer/DisplayHostViewController.swift
+++ b/ownCloud/Client/Viewer/DisplayHostViewController.swift
@@ -151,6 +151,19 @@ class DisplayHostViewController: UIPageViewController {
 
 		addKeyCommand(UIKeyCommand.ported(input: UIKeyCommand.inputLeftArrow, modifierFlags: .shift, action: #selector(keyCommandPreviousItem), discoverabilityTitle: OCLocalizedString("Previous item", nil)))
 		addKeyCommand(UIKeyCommand.ported(input: UIKeyCommand.inputRightArrow, modifierFlags: .shift, action: #selector(keyCommandNextItem), discoverabilityTitle: OCLocalizedString("Next item", nil)))
+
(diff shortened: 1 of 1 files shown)


jesmrec:
Is this straight: install and go? or any setup is required?

After installing the current branch in iPhone and iPad, and opening an imagen inside a folder with multiple images, the preview does not show the left-right buttons.

felix-schwarz:
@jesmrec The arrows only show up when you hover over the image viewer with a pointer - via a mouse or trackpad connected to the iPad. If you don't have an iPad case with trackpad at hand, connecting a mouse to the iPad via USB or Bluetooth should make this testable, too.

jesmrec:
Thanks for the response. Two comments here:

- This is available just for iPad, right? not for iPhone
- After testing it with a physical trackpad, i noticed that `svg` files does not show correctly the prompt arrows. Check the following video out (name and type of the displayed image in the top bar):

https://github.com/user-attachments/assets/6f3f878c-3d34-454a-a288-722415124649

is this something fixable?

`1911cd056`
iPadAir, iPadOS 18.4

felix-schwarz:
@jesmrec "Fixed" with the latest commit: SVGs are no longer included in the carousel, as they are shown via QuickLook - which uses a remote view that doesn't play well with gesture recognition by the pagination control itself (=> not possible to go back to the previous image or forward to the next) as well as the hover gesture recognizer.

jesmrec:
Perfect! it works excluding `svg`. With any other kind of image or media file, it works.

jesmrec (approved):

Issue #1332, opened 2024-04-05
[a11y] 11.1.3.2 Meaningful sequence


tschiebel:
audit: https://infinite.owncloud.com/s/qSvYucPltEYUhhc

https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de
https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf

### Personal Space

- [x] If multiple selection has been activated, it is not possible with the screen reader to
to focus on the checkboxes themselves. Instead, directory and
and file names as well as date entries are focused, but not the checkboxes themselves.
As no roles are displayed, users are not informed about the presence of checkboxes.
of checkboxes. See https://a11yguidelines.orange.com/en/mobile/ios/development/#grouping-elements for more information.

PR #1454, opened 2025-04-07
[fix/a11y-1332] Fix 11.1.3.2 Meaningful sequence

felix-schwarz:
## Description
This addresses the issue brought up in #1332 to the extent possible. A few notes on that:

Already previously, the ownCloud iOS app presented the same trait (`button` / "Taste" in German) as the Apple Files app in the same situation:

Files App | OC iOS app
------------|-------------
Bildschirmfoto 2025-04-07 um 10 16 55 | Bildschirmfoto 2025-04-07 um 10 20 16

Already previously, the ownCloud iOS app, like the Apple Files app, communicated a selected item by prepending the word "selected" to the VoiceOver of the line.

So, in summary, the OC app already behaved like the Apple Files app.

A `checkbox` trait does not exist on iOS. The closest to a `checkbox` is `toggleButton`, introduced in iOS 17. In multi-select contexts, the app will therefore now use `toggleButton` instead of `button` on iOS 17+.

## Related Issue
#1332 

## Types of changes


hosy (approved):

jesmrec:
@felix-schwarz:

> Already previously, the ownCloud iOS app, like the Apple Files app, communicated a selected item by prepending the word "selected" to the VoiceOver of the line.

within the current branch, the "selected" word is not narrated by VoiceOver anymore. Is that intended? It could be indicative for a visual impaired person to know whether the item was selected or not. Moreover, if the item becomes unselected, it should be narrated too.

jesmrec:
With the new implementation, in multiselection mode:

- If a cell is selected, the word `selected` is narrated by VoiceOver first ✅ 
- If a cell is not selected, VoiceOver narrates nothing. By default, a visual impaired user could know that is not selected (therefore unselected) 

Would it be more consistent in multiselection mode, that VoiceOver narrates always the selected/unselected status of the cell first? 

For sure this is not a bad behaviour, just an idea as improvement. Not a blocker at all.

felix-schwarz:
> With the new implementation, in multiselection mode:
> 
> * If a cell is selected, the word `selected` is narrated by VoiceOver first ✅
> * If a cell is not selected, VoiceOver narrates nothing. By default, a visual impaired user could know that is not selected (therefore unselected)
> 
> Would it be more consistent in multiselection mode, that VoiceOver narrates always the selected/unselected status of the cell first?

I wondered about the same thing.

`selected` is added automatically by Apple for VoiceOver. So the omission of `unselected` for unselected rows appears to be a choice made by Apple, which I currently don't see a way to override.

jesmrec:
Then, it's also a system restriction not to be skipped.

This is OK from my side, then.

Approved

jesmrec (approved):

Issue #1364, opened 2024-06-17
[BUG] Accessibility, Large Text: Setup Account


hosy:
- Enable Accessibility Large Text
- Enable largest text size
- Add new Account > Enter URL
- Login Buttons are not reachable

![Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-06-17 at 15 34 46](https://github.com/owncloud/ios-app/assets/736109/bfdcae5b-4f33-43dc-adbb-8d13a5c98fab)

PR #1460, opened 2025-04-22
[fix/a11y-1364-setupwizard] Fix Setup Wizard issue with largest Dynamic Type setting

felix-schwarz:
## Description
Make layout of setup wizard also usable when using the largest Dynamic Type text size by moving the content into a scrollview that comes into effect if the content of a step doesn't fit in the available vertical space.

## Related Issue
#1364

## Types of changes


hosy (approved):

jesmrec:
The problem of the authentication is fixed. But, the app could need more adjustments to handle correctly the large text mode.

jesmrec (approved):

PR #1468, opened 2025-05-26
[fix/confidential-office-pan-zoom] Fix pan/zoom issue when using confidential overlay options

felix-schwarz:
## Description
PreviewViewController/WebViewDisplayViewController: if watermark/confidential overlays are used, QLPreviewController no longer reacts to scroll and zoom events, so use WebKit's support for QuickLook formats instead in those cases.

Testing note: test with multi-page/multi-slide documents to reproduce the issue.

## Related Issue
https://github.com/owncloud/enterprise/issues/7224

## Types of changes

### DIFF
--- a/ownCloud/Client/Viewer/QuickLook/PreviewViewController.swift
+++ b/ownCloud/Client/Viewer/QuickLook/PreviewViewController.swift
@@ -7,18 +7,19 @@
 //
 
 /*
-* Copyright (C) 2018, ownCloud GmbH.
-*
-* This code is covered by the GNU Public License Version 3.
-*
-* For distribution utilizing Apple mechanisms please see https://owncloud.org/contribute/iOS-license-exception/
-* You should have received a copy of this license along with this program. If not, see <http://www.gnu.org/licenses/gpl-3.0.en.html>.
-*
-*/
+ * Copyright (C) 2018, ownCloud GmbH.
+ *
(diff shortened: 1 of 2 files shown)


jesmrec:
### (1)

I noticed the following behaviour about zooming txt files:

- current branch with no watermark ->  txt files can not be zoomed
- current branch with watermark -> txt files can be zoomed

that means, zooming is only posible in txt if watermark is screened. For other type of files, works fine in both modes (with and without watermark)

CC @felix-schwarz

felix-schwarz:
@jesmrec Text files couldn't be zoomed with the QuickLook text viewer before the change. With watermark enabled, WebKit is used instead of QuickLook - and WebKit appears to support zoom for text files.

Since this is not a regression, I'd leave it as is. If text should be zoomable in general, I'd suggest to move that to an issue and implement it separately.

jesmrec:
> @jesmrec Text files also couldn't be zoomed with the QuickLook text viewer before the change. With watermark enabled, WebKit is used instead of QuickLook - and WebKit appears to support zoom for text files.
Since this is not a regression, I'd leave it as is. If text should be zoomable in general, I'd suggest to move that to an issue and implement it separately.

works for me, that broke nothing and ftm, i think there is no an specific request about zooming files. The comment was about the different behaviour. We'll let it be.

approved on my side

PR #1470, opened 2025-06-04
Fix: Add sharing action to folder navigation bar menu

### LINKED ISSUE #1469: [BUG] Sharing action missing from folder navigation bar menu
### Steps to reproduce
1. Navigate to any folder in the file list
2. Tap on the three-dots menu button (•••) in the top-right corner of a folder list item 
3. Observe that "Sharing" appears as the first option in the action menu
4. Navigate into that same folder
5. Tap on the three-dots menu button (•••) in the navigation bar
6. Observe that "Sharing" is missing from this action menu

### Expected behavior
The "Sharing" action should be available in both locations:
- In the action menu when tapping the three-dots button on a folder item in the list
- In the action menu when tapping the three-dots button in the navigation bar when inside that folder

This would provide consistent access to sharing functionality regardless of where the user accesses the folder actions from.

### Actual behavior
The "Sharing" action is only available in the folder list item's action menu (`.moreItem` location) but missing from the navigation bar's action menu when inside the folder (`.moreFolder` location).

### Analysis
The `CollaborateAction` class is not registered for the `.moreFolder` location. It only includes:
- `.keyboardShortcut`
- `.contextMenuSharingItem`  
- `.moreItem`
- `.moreDetailItem`
- `.accessibilityCustomAction`

But not `.moreFolder`, which is used for the navigation bar's more button when viewing folder contents.

### Client
**iOS version:** 18.5

**ownCloud app version:** 12.5.0

**Device model:** iPhone

### Server configuration
Reproducible with both:
- ownCloud 10 (classic WebDAV)
- ownCloud Infinite Scale (with Spaces)

### Proposed solution
Add `.moreFolder` to the supported locations in `CollaborateAction.swift`:

```swift
override class var locations : [OCExtensionLocationIdentifier]? { 
    return [.keyboardShortcut, .contextMenuSharingItem, .moreItem, .moreFolder, .moreDetailItem, .accessibilityCustomAction] 
}
```

Additional consideration: For root folders, sharing behavior should be consistent with server permissions and drive types (e.g., personal spaces vs project spaces).

michaelstingl:
## Description
This PR adds the "Sharing" action to the folder navigation bar's more menu (•••) to ensure consistency with the action menu shown for folder items in the file list.

## Related Issue
Fixes #1469

## Motivation and Context
Users expect the same actions to be available whether they access a folder's actions from:
- The three-dots menu on a folder item in the file list
- The three-dots menu in the navigation bar when inside that folder

Previously, the "Sharing" action was only available in the file list context menu but missing from the navigation bar menu.

## How Has This Been Tested?
- Tested on iPhone with iOS 18.5
- Verified with both ownCloud 10 and ownCloud Infinite Scale servers
- Confirmed that:
  - Sharing appears in regular folder navigation bar menus
  - Sharing is correctly hidden for OC10 root folders
  - Sharing is correctly hidden for oCIS virtual and personal space roots
  - Sharing is available for oCIS project space roots (when allowed by server)

N/A - UI remains the same, just adds the missing menu item

## Types of changes

## Checklist:

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -40,6 +40,7 @@ disabled_rules:
     - nesting
     - comma
     - large_tuple
+    - force_cast
 custom_rules:
   empty_line_after_guard_statement:
     included: ".*\\.swift"
--- a/ownCloud.xcodeproj/project.pbxproj
+++ b/ownCloud.xcodeproj/project.pbxproj
@@ -342,6 +342,11 @@
 		DC59087D2AA8B82200BFF393 /* BookmarkSetupStepPrepopulateViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC59087C2AA8B82200BFF393 /* BookmarkSetupStepPrepopulateViewController.swift */; };
 		DC59087F2AA8D25400BFF393 /* CertificateSummaryView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC59087E2AA8D25400BFF393 /* CertificateSummaryView.swift */; };
 		DC5C48A32918FB7400EBC053 /* CollectionSidebarViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC5C48A22918FB7400EBC053 /* CollectionSidebarViewController.swift */; };
+		DC5D397F2DDC649700EEDFE7 /* RecentLocationStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC5D397D2DDC649700EEDFE7 /* RecentLocationStore.swift */; };
+		DC5D39812DDC64A400EEDFE7 /* RecentLocation.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC5D39802DDC64A400EEDFE7 /* RecentLocation.swift */; };
+		DC5D39862DDE6D4E00EEDFE7 /* RecentLocationCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC5D39852DDE6D4E00EEDFE7 /* RecentLocationCell.swift */; };
+		DC5D398B2DDE743300EEDFE7 /* RecentLocation+Interactions.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC5D398A2DDE743300EEDFE7 /* RecentLocation+Interactions.swift */; };
+		DC5D398D2DE06B2600EEDFE7 /* RecentLocation+UniversalItemListCellContentProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC5D398C2DE06B2600EEDFE7 /* RecentLocation+UniversalItemListCellContentProvider.swift */; };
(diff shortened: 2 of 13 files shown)


michaelstingl:
Added a simple unit test in the second commit to verify the location is included. Happy to adjust or remove if you prefer a different approach!

jesmrec:
Thanks Michael for your contribution!! let's take a look into there ;)

ownCloud/Client/Actions/Actions+Extensions/CollaborateAction.swift
```diff
@@ -24,13 +24,29 @@ class CollaborateAction: Action {
...
 	override class var name : String { return OCLocalizedString("Sharing", nil) }
-	override class var locations : [OCExtensionLocationIdentifier]? { return [.keyboardShortcut, .contextMenuSharingItem, .moreItem, .moreDetailItem, .accessibilityCustomAction] }
+	override class var locations : [OCExtensionLocationIdentifier]? { return [.keyboardShortcut, .contextMenuSharingItem, .moreItem, .moreFolder, .moreDetailItem, .accessibilityCustomAction] }
 	override class var keyCommand : String? { return "S" }
 	override class var keyModifierFlags: UIKeyModifierFlags? { return [.command] }
 
 	// MARK: - Extension matching
 	override class func applicablePosition(forContext: ActionContext) -> ActionPosition {
 		if forContext.items.count == 1, let core = forContext.core, core.connectionStatus == .online, core.connection.capabilities?.sharingAPIEnabled == 1, let item = forContext.items.first, item.isShareable {
+			// For .moreFolder location: Check special cases at root level
```
felix-schwarz, line 34: The checks carried out here should not be necessary as the cases it tries to handle should already be prevented by checking `item.isShareable`, which relies on the permissions returned by the server.

The action should also not show up for root folders even without that code. Can you verify/reproduce?

ownCloudTests/Actions/CollaborateActionTests.swift
```diff
@@ -0,0 +1,51 @@
+//
```
felix-schwarz, line 1: Thanks for the effort, but I think a unit test for such a small change is not necessary.

ownCloud/Client/Actions/Actions+Extensions/CollaborateAction.swift
```diff
@@ -24,13 +24,29 @@ class CollaborateAction: Action {
 	override class var identifier : OCExtensionIdentifier? { return OCExtensionIdentifier("com.owncloud.action.collaborate") }
 	override class var category : ActionCategory? { return .normal }
 	override class var name : String { return OCLocalizedString("Sharing", nil) }
-	override class var locations : [OCExtensionLocationIdentifier]? { return [.keyboardShortcut, .contextMenuSharingItem, .moreItem, .moreDetailItem, .accessibilityCustomAction] }
+	override class var locations : [OCExtensionLocationIdentifier]? { return [.keyboardShortcut, .contextMenuSharingItem, .moreItem, .moreFolder, .moreDetailItem, .accessibilityCustomAction] }
```
felix-schwarz, line 27: I think the change in this line should be the only change necessary to implement the change. The rest should not be needed.

jesmrec:
That's a good catch ;)

Fixed and approved from my side.

jesmrec (approved):

jesmrec:
Done!

Issue #1473, opened 2025-06-27
Add automatically generated SBOM


felix-schwarz:
This issue explores automatic SBOM generation for this app.

## Existing solutions
- a direct integration like used in https://github.com/owncloud/android/pull/4599 is not available for Xcode projects using SwiftPM
- I couldn't find any third-party project fit for the purpose. The closest I could find is [a project generating CycloneDX files for CocoaPods](https://github.com/CycloneDX/cyclonedx-cocoapods) - something which we no longer use

## Findings and a possible solution
- a JSON list of Swift packages is maintained inside the Xcode project at ownCloud.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```json
{
  "originHash" : "cc8e62b9f24062ff6e87f5e91e6f6cb41653dcf8250ce8973635e9ed0a5c561d",
  "pins" : [
    {
      "identity" : "down",
      "kind" : "remoteSourceControl",
      "location" : "https://github.com/johnxnguyen/Down",
      "state" : {
        "branch" : "master",
        "revision" : "e754ab1c80920dd51a8e08290c912ac1c2ac8b58"
      }
    },
    {
      "identity" : "openssl",
      "kind" : "remoteSourceControl",
      "location" : "https://github.com/krzyzanowskim/OpenSSL.git",
      "state" : {
        "revision" : "8729614a93e6b20307db3c3359aed81f030facaa",
        "version" : "3.3.2000"
      }
    },
    {
      "identity" : "plcrashreporter",
      "kind" : "remoteSourceControl",
      "location" : "https://github.com/microsoft/plcrashreporter.git",
      "state" : {
        "revision" : "8c61e5e38e9f737dd68512ed1ea5ab081244ad65",
        "version" : "1.12.0"
      }
    },
    {
      "identity" : "pocketsvg",
      "kind" : "remoteSourceControl",
      "location" : "https://github.com/pocketsvg/PocketSVG.git",
      "state" : {
        "revision" : "10c6885e427e88d35fbb64010fbd6385cb58fef7",
        "version" : "2.7.3"
      }
    }
  ],
  "version" : 3
}
```
- this information could be rearranged to match the CycloneDX JSON format as seen f.ex. [here](https://github.com/CycloneDX/bom-examples/blob/master/SBOM/laravel-7.12.0/bom.1.4.json).
- problem: no license information is available in Package.resolved
- solution:
[...]

PR #1474, opened 2025-06-30
[chore/sbom] Auto-generate SBOM using cdxgen

felix-schwarz:
## Description
Auto-generate SBOM using cdxgen

## Related Issue
#1473

### DIFF
--- a/.github/workflows/generate-sbom.yml
+++ b/.github/workflows/generate-sbom.yml
@@ -32,7 +32,8 @@ jobs:
         shell: bash
         run: |
           export FETCH_LICENSE=true
-          cdxgen -o ./sbom.json -t swift
+          export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
+          cdxgen -o ./sbom.json -t swift .
       - name: Format JSON
         shell: bash
         run: |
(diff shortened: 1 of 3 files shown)


DeepDiver1975 (changes requested):
- [x] readable json would be great - ` | json_pp`
- [ ] github.com/krzyzanowskim/OpenSSL is listed twice with two different version - needs a second look if wrong data or mixup of dependencies in the projects

DeepDiver1975 (changes requested):
- [ ] let's name it sbom.json

felix-schwarz:
> * github.com/krzyzanowskim/OpenSSL is listed twice with two different version - needs a second look if wrong data or mixup of dependencies in the projects

The SDK actually pointed to an older version than the app, which only became visible because this script asks Xcode to resolve packages for both the SDK and the app project individually. In practice, Xcode used the later version for both. I've created a new PR (https://github.com/owncloud/ios-sdk/pull/144) to make the SDK use the same version of the OpenSSL  Swift PM package, so there's no longer a difference even in the resolved package lists.

DeepDiver1975 (approved):

PR #1489, opened 2025-08-25
[feature/confidential-configurability] Option to exempt actions from auto-disallow 

felix-schwarz:
## Description
Adds a new MDM option `confidential.exempted-actions` to exempt one or more actions from being automatically disallowed when enabling Confidential Protection.

Example to allow the open in action:
```xml
<key>confidential.exempted-actions</key>
<array>
	<string>com.owncloud.action.openin</string>
</array>
```

## Related Issue
https://github.com/owncloud/enterprise/issues/7296

## Types of changes

### DIFF
--- a/changelog/12.6.1_2025_xx_xx/1489
+++ b/changelog/12.6.1_2025_xx_xx/1489
@@ -1,0 +1,13 @@
+Change: Option to exempt actions from auto-disallow
+
+Adds a new MDM option `confidential.exempted-actions` to exempt one or more actions from being automatically disallowed when enabling Confidential Protection.
+
+Example to allow the open in action:
+```xml
+<key>confidential.exempted-actions</key>
+<array>
+	<string>com.owncloud.action.openin</string>
+</array>
+```
+
+https://github.com/owncloud/ios-app/pull/1489


felix-schwarz:
@jesmrec The option `confidential.exempted-actions` can only exempt actions that would otherwise be automatically disallowed when enabling Confidential Protection. Those actions are:

- `com.owncloud.action.copy`
- `com.owncloud.action.markup`
- `com.owncloud.action.openin`

Those are documented as allowed/supported values for `confidential.exempted-actions` in [configuration.adoc](https://github.com/owncloud/ios-app/blob/a5132fb842afd830ff56b32e818aa57a54e57e9a/doc/configuration.adoc).

Right now, the array is validated as a whole: if any element in it is not passing validation, the entire array is rejected and the option will not be applied. This will happen here for an invalid value like `com.owncloud.action.cutpasteboard`.

Unlike many other validations, however, this one does not yet log a message in case the validation fails. I'll add one later today.

Sidenote: the Cut action is not included in that list since it only works internally within the app (f.ex. to move files) by copying just an internal reference to the file to the clipboard in an OC-specific format. Unlike Copy, it doesn't download and offer the file itself via the clipboard.

jesmrec:
Approved.

PR #1531, opened 2026-04-08
[chore/update-gh-workflows] Update GitHub workflows

felix-schwarz:
## Description
Updating GitHub workflows to meet https://central.owncloud.org/t/upcoming-changes-for-developers-what-you-need-to-know/65650 requirements.

### DIFF
--- a/.github/workflows/configuration-documentation.yml
+++ b/.github/workflows/configuration-documentation.yml
@@ -24,8 +24,9 @@ jobs:
       - name: Run Generate Docs
         run: ./generate_docs.sh
         working-directory: ./tools/GenerateDocs/
-      - name: Commit files
-        uses: stefanzweifel/git-auto-commit-action@04702edda442b2e678b25b537cec683a1493fcb9 # v7.1.0 (previously v4)
+      - uses: GuillaumeFalourd/git-commit-push@205c043bca2f932f7a48a28a8d619ba30eb84baf # v1.3
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/configuration-documentation.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         run: ./generate_docs.sh
         working-directory: ./tools/GenerateDocs/
       - name: Commit files
-        uses: stefanzweifel/git-auto-commit-action@v4
+        uses: stefanzweifel/git-auto-commit-action@04702edda442b2e678b25b537cec683a1493fcb9 # v7.1.0 (previously v4)
```
DeepDiver1975: GuillaumeFalourd/git-commit-push is already white listed - reads like a similar actions .....

is it possible to use GuillaumeFalourd/git-commit-push in this case as well? THX

DeepDiver1975 (changes requested):

.github/workflows/configuration-documentation.yml
felix-schwarz: IIRC there was an issue with commit rights with regards to a token. Let's try with the other, though. If it turns out not to work, we know what to fall back to. Will make the changes in a bit.

DeepDiver1975 (approved):

DeepDiver1975 (approved):

PR #1563, opened 2026-06-29
[chore/update-same-branch-actions] Use verified commits for same-branch committing workflows

felix-schwarz:
## Description
- add reusable action to commit changes as verified commits using the GitHub API
- update existing workflows to use new reusable action for verified commits in the same branch


.github/actions/verified-api-commit/action.yml
```diff
@@ -0,0 +1,230 @@
...
+#           commit-message: "chore: auto-format documents"
+#           retry-on-conflict: 'true'
+#           files: |
+#             doc/
+#             src/config.json
+#             package.json
+
+name: 'Verified API Commit'
+description: 'Commits specifically changed files and folders via the GitHub API to generate a Verified signature.'
+author: 'Your Name'
```
Matthias-Huehne-Kiteworks: Is there an author needed?

dj4oC (changes requested):
Adds a new composite GitHub Action (`.github/actions/verified-api-commit`) that commits changes via GitHub's GraphQL `createCommitOnBranch` API to produce signed "Verified" commits, and switches `configuration-documentation.yml` and `pull-ocis-icons.yml` from a static SSH deploy key + third-party `git-commit-push` action to this action, authenticated via a per-run GitHub App token.

### Findings

**Security** (comment, not blocking)
- Net positive direction: replacing the long-lived `DEPLOYMENT_SSH_KEY` deploy key with per-run, revocable GitHub App tokens plus `persist-credentials: false` reduces the checkout step's blast radius. `FILES_INPUT`/`COMMIT_MESSAGE` are passed via `env:` (not inline `${{ }}` interpolation in the `run:` body) and safely escaped into the GraphQL payload via `jq --arg`, avoiding the classic Actions script-injection pattern. Both workflows trigger on `push` only, so a forked PR can't reach these secrets.
- `TRANSLATION_APP_ID`/`TRANSLATION_APP_PRIVATE_KEY` are scoped for Transifex automation elsewhere in the repo (e.g. `pull-transifex.yml`, which only needs `contents: read`). This PR reuses the same App's credentials to directly push (`contents: write`) to release-adjacent branches for unrelated icon-pulling/doc-generation automation. Please confirm the App's actual installed permission scope, and consider whether a dedicated App/secret for commit automation would avoid scope confusion later.
- The new composite action's header notes it was generated by/with an AI tool. It's new and handles credential/commit-flow logic with nontrivial retry/rebase branches — worth an explicit human sign-off on the script itself rather than treating AI authorship as pre-vetted.

**Stability** (one item should be fixed before merge)
- `.github/actions/verified-api-commit/action.yml`: the conflict-retry path (`git stash push` → `git fetch` → `git rebase FETCH_HEAD` → `git stash pop || true`) has no failure handling on the rebase or the stash pop. If a real conflict occurs there — exactly the race this retry logic exists to handle — the working tree is left with raw `<<<<<<<`/`=======`/`>>>>>>>` conflict markers, and the next loop iteration's `git status --porcelain` + base64-encode step will pick that file up and push it as a "Verified" commit, silently corrupting `doc/CONFIGURATION.json` or an icon file on a release branch. Recommend `git rebase FETCH_HEAD || exit 1` and checking `git stash pop` for conflicts instead of swallowing with `|| true`.
- Related: `$RESPONSE` from `gh api graphql` is only pattern-matched for two known failure strings; any other failure mode (auth error, network error, missing `jq`) falls through to the "success" branch and reports green. Consider checking the `gh api` exit code directly rather than `2>&1 ... || true` + string-grepping combined output.
- `pull-ocis-icons.yml` now triggers on `push` to `release/*` and itself pushes to that ref. This causes one extra, wasted-but-not-infinite self-triggered run per icon update (the second run's `git status --porcelain` will be empty and hit the no-op `exit 0` path) — minor, just flagging.
- `submodules: true` added to `pull-ocis-icons.yml`'s checkout looks unused (the job never touches the `ios-sdk` submodule) — harmless since it's a public HTTPS submodule, but adds needless clone time.
- Please confirm the `milestone/*` → `release/*` trigger-branch change in `configuration-documentation.yml` is intentional — docs will stop auto-updating on `milestone/*` pushes.

**Performance** — Nothing blocking. The per-file `jq` re-serialization and in-memory base64 encoding are technically O(n²)/full-buffer patterns, but at current file counts (~33 icons, 2 doc files ≤107KB) the added cost is milliseconds against a multi-minute macOS-runner job.

**Coverage** — N/A: CI/YAML-only change, no Swift/UIKit code, so the XCTest/XCUITest coverage rule doesn't apply. No evidence in the PR description of a dry-run (e.g. `workflow_dispatch` against a scratch branch) exercising the retry/rebase path — worth doing once before/after merge. No changelog entry needed, consistent with other `.github/`-only chore PRs in this repo.

### Note
Build verification skipped (Xcode not available in cloud review environment). Code analysis only.

### Verdict
Changes requested

---
_Generated by [Claude Code](https://claude.ai/code)_

Matthias-Huehne-Kiteworks (changes requested):

.github/actions/verified-api-commit/action.yml
```diff
@@ -0,0 +1,230 @@
...
+  files:
+    description: 'A multiline list of files and/or folders to check and commit'
+    required: true
+  commit-message:
+    description: 'The commit message to use'
+    required: true
+  github-token:
+    description: 'The GITHUB_TOKEN to use for the API call'
+    required: true
+    default: ${{ github.token }}
```
jesmrec, line 61: If the `github-token` is required, not sure if `default` is useful. `default` covers a non-mandatory option in order not to keep the value empty.

.github/actions/verified-api-commit/action.yml
```diff
@@ -0,0 +1,230 @@
...
+                  # Linux (GNU) requires -w 0 to disable line breaks and accepts the file directly
+                  B64_CONTENT=$(base64 -w 0 "$FILE_PATH")
+                fi
+                
+                # Use jq to safely append the object to our JSON array
+                ADDITIONS_JSON=$(echo "$ADDITIONS_JSON" | jq --arg path "$FILE_PATH" --arg contents "$B64_CONTENT" '. + [{path: $path, contents: $contents}]')
+              else
+                echo "Warning: $FILE_PATH is not a file or was deleted. Skipping addition."
+              fi
+            done <<< "$(git status --porcelain "$ITEM" 2>/dev/null)"
```
jesmrec: I'd set this one with the `--` option to avoid paths starting with `-` to be interpreted as git commands:

`done <<< "$(git status --porcelain -- "$ITEM" 2>/dev/null)"`

.github/actions/verified-api-commit/action.yml
```diff
@@ -0,0 +1,230 @@
...
+                repo: $repo,
+                branch: $branch,
+                oid: $oid,
+                msg: $msg,
+                files: $files
+              }
+            }' | gh api graphql --input - 2>&1) || true
+          
+          # 6. Handle Concurrency Mismatch
+          if echo "$RESPONSE" | grep -q "STALE_DATA" || echo "$RESPONSE" | grep -q "Expected branch to point to"; then
```
jesmrec, line 195: Just as an improvement, before checking whether the `$RESPONSE` contains such strings, you can take advantage of `API_STATUS=$?` to verify that the `gh api` command returned success. If it did,  `API_STATUS=$?` will return `0`, otherwise return something `!=0`. You'll avoid wrong positives in case of an api failure.

jesmrec (changes requested):

.github/actions/verified-api-commit/action.yml
felix-schwarz: It's indeed [optional](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax), but I'll give it a proper value with the next commit.

.github/actions/verified-api-commit/action.yml
felix-schwarz: The problem here is that if I don't set `required` to `true`, it could be read as "a GitHub token is optional" when it is actually necessary.

On the other hand, if I don't provide a default value, all users of the action will have to add that option in their workflow - when really the default value should be what's used in almost all cases.

Since, as per [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#example-specifying-inputs) `Actions using required: true will not automatically return an error if the input is not specified.`, both can be used together to indicate a requirement while also affording the convenience of providing a default value.

.github/actions/verified-api-commit/action.yml
felix-schwarz: Very good catch! Thanks! I'll fix this with the next commit.

.github/actions/verified-api-commit/action.yml
felix-schwarz: Thanks for the hint. However, since the code needs to handle three different cases based on the contents of the response, I'm afraid the `API_STATUS` may not be sufficient for differentiation.

felix-schwarz:
@hosy @Matthias-Huehne-Kiteworks @jesmrec I addressed your findings. Please have another look.

.github/actions/verified-api-commit/action.yml
jesmrec: It's confusing that the `required` does not force to use the parameter by the caller workflow, but it is how it works, so... 👍 👍

.github/actions/verified-api-commit/action.yml
jesmrec: 👍

jesmrec (approved):

felix-schwarz:
@dj4oC 

> * Please confirm the `milestone/*` → `release/*` trigger-branch change in `configuration-documentation.yml` is intentional — docs will stop auto-updating on `milestone/*` pushes.

This is intentional. We've not been using `milestone/*` branches for a while.

Matthias-Huehne-Kiteworks (approved):

PR #1572, opened 2026-07-31
[feature/webapp-url-delegation] Configuration options for in-app web-app URL opening

felix-schwarz:
## Description
Adds a new option `action.in-app-web-app-link-open-mode` to allow configuring behavior when in-app web apps want to open a URL in a new window.

Possible values:
- `with-default-browser`: ask the user if the link should be opened in the default browser **(default)**
- `navigate`: ask the user if the in-app browser hosting the web apps should navigate to the link
- `choice`: ask the user whether to delegate link opening to the default browser or perform an in-app navigation 
- `deny`: deny opening the link and tell the user
- `silent-deny`: silently deny opening the link (effectively turning it into a no-op) and log a debug message (not including the URL)

## Related Issue
OCSRE-2029
OCSRE-2031

Screenshot 2026-07-31 at 16 36 04

## Types of changes

### DIFF
--- a/ownCloudAppShared/Client/Actions/Implementations/ClientWebAppViewController.swift
+++ b/ownCloudAppShared/Client/Actions/Implementations/ClientWebAppViewController.swift
@@ -62,10 +62,10 @@ open class ClientWebAppViewController: UIViewController, WKUIDelegate {
 			var openAction, navigateAction: UIAlertAction?
 
 			title = OCLocalizedString("Web app requests to open link", nil)
-			openAction = UIAlertAction(title: "Open", style: .default) { (_) in
+			openAction = UIAlertAction(title: OCLocalizedString("Open", nil), style: .default) { (_) in
 				OCAuthenticationBrowserSessionCustomScheme.open(url) // Open in default browser
 			}
-			navigateAction = UIAlertAction(title: "Navigate", style: .default) { (_) in
+			navigateAction = UIAlertAction(title: OCLocalizedString("Navigate", nil), style: .default) { (_) in
 				webView.load(navigationAction.request) // Navigate
 			}
 
--- a/ownCloudAppShared/Client/Actions/Implementations/ClientWebAppViewController.swift
+++ b/ownCloudAppShared/Client/Actions/Implementations/ClientWebAppViewController.swift
@@ -53,6 +53,60 @@ open class ClientWebAppViewController: UIViewController, WKUIDelegate {
 		return configuration
 	}
 
+	public func webView(_ webView: WKWebView, createWebViewWith configuration: WKWebViewConfiguration, for navigationAction: WKNavigationAction, windowFeatures: WKWindowFeatures) -> WKWebView? {
+		if navigationAction.targetFrame == nil, let url = navigationAction.request.url,
+		   let openModeRaw = OpenInWebAppAction.classSetting(forOCClassSettingsKey: .inAppWebAppLinkOpenMode) as? String,
+		   let openMode = InAppWebAppLinkOpenMode(rawValue: openModeRaw) {
+			// Web app wants to open a link in a new window
+			var title: String
+			var openAction, navigateAction: UIAlertAction?
+
+			title = OCLocalizedString("Web app requests to open link", nil)
+			openAction = UIAlertAction(title: "Open", style: .default) { (_) in
+				OCAuthenticationBrowserSessionCustomScheme.open(url) // Open in default browser
+			}
+			navigateAction = UIAlertAction(title: "Navigate", style: .default) { (_) in
+				webView.load(navigationAction.request) // Navigate
+			}
+
+			switch openMode {
+				case .withDefaultBrowser:
+					navigateAction = nil
+
+				case .navigate:
+					openAction = nil
+
+				case .choice:
+					// allow all actions
+					break
+
+				case .deny:
+					title = OCLocalizedString("The web app attempted to open the URL below. Opening was denied by configuration.", nil)
+					openAction = nil
+					navigateAction = nil
+
+				case .silentDeny:
+					Log.debug("ClientWebAppViewController denied opening a link.")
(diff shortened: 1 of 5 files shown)


ownCloudAppShared/Client/Actions/Implementations/ClientWebAppViewController.swift
```diff
@@ -53,6 +53,60 @@ open class ClientWebAppViewController: UIViewController, WKUIDelegate {
...
+	public func webView(_ webView: WKWebView, createWebViewWith configuration: WKWebViewConfiguration, for navigationAction: WKNavigationAction, windowFeatures: WKWindowFeatures) -> WKWebView? {
+		if navigationAction.targetFrame == nil, let url = navigationAction.request.url,
+		   let openModeRaw = OpenInWebAppAction.classSetting(forOCClassSettingsKey: .inAppWebAppLinkOpenMode) as? String,
+		   let openMode = InAppWebAppLinkOpenMode(rawValue: openModeRaw) {
+			// Web app wants to open a link in a new window
+			var title: String
+			var openAction, navigateAction: UIAlertAction?
+
+			title = OCLocalizedString("Web app requests to open link", nil)
+			openAction = UIAlertAction(title: "Open", style: .default) { (_) in
```
JohnKuan: "Open" and "Navigate" are not localized, are these intentional?

felix-schwarz: Good catch, thanks! I'll make sure it, too, is localized.

felix-schwarz: I now addressed that in the latest commit. Please have another look.

felix-schwarz: I also rebased against `master` and signed my commits.

JohnKuan (dismissed):

jesmrec:
Reproduced by opening the link in the inline preview and in Collabora.

`Branding.plist`

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>action.in-app-web-app-link-open-mode</key>
	<string></string>
</dict>
</plist>
```


1. `with-default-browser`

Screenshot 2026-08-20 at 12 54 21

- Open with default (Safari) ✅ 
- Change default to Chrome -> open with Chrome ✅ 
- Change default to Opera -> open with Opera ✅ 
- Cancel ✅  

2. `navigate`

Screenshot 2026-08-20 at 13 00 49

With browsers open, web is displayed in the inline ownCloud browser ✅ 

3. `choice`

Screenshot 2026-08-20 at 13 05 05

- Open -> Behaves like `with-default-browser` ✅ 
- Navigate -> Behaves like `navigate` ✅ 
- Cancel ✅ 

4. `deny`

Screenshot 2026-08-20 at 13 15 14

That's probably the one to set by the customer.

No way to open the web, just cancelling ✅ 

5. `silent-deny`

Behaves as `deny` without dialog ✅

jesmrec:
Question:

The branding option `action.in-app-web-app-link-open-mode` rules the opening mode from the in-web providers (collabora, onlyoffice, microsoft...).

But, the inline previewer always allows to open the links. For example, using `deny`, opening a docx file with a link inside: `deny` does not prevent the link to be opened from there, but, by opening with `Collabora` it is correctly denied (file `testOffice.docx`):

https://github.com/user-attachments/assets/572bb02a-2405-4b3e-84d6-d94a1107a55d

That's a concept question, non technical. If the idea is preventing that users don't open links embedded inside of files, should it be prevented everywhere? not only in the web

The feature works fine technically.

@felix-schwarz

felix-schwarz:
@jesmrec Good point! The new option is really only meant for links opened from within web apps presented in-app.

Background: the customer asking for this feature uses proxy/filter software that can only control which web pages can be opened inside Safari. The proxy/filter software can't limit what can be shown in embedded, in-app webviews like those used by the OC app to present web apps in-app.

It's therefore important for the customer to be able to specifically control what happens to links tapped inside the web app view. Other areas opening links are fine with the customer - provided these links are opened in the iOS default browser, where the customer can exert control.

jesmrec:
If i understood correctly:

- Opening webs in browser (Safari): customer proxy controls
- Opening links inside files opened in web: customer proxy does not control, new feature (`action.in-app-web-app-link-open-mode`) controls
- Opening links inside files in-app: customer does not matter

Is that right @felix-schwarz ?

felix-schwarz:
@jesmrec:

> If i understood correctly:
> 
> * Opening webs in browser (Safari): customer proxy controls

Yes.

> * Opening links inside files opened in web: customer proxy does not control, new feature (`action.in-app-web-app-link-open-mode`) controls

Yes, the new feature makes sure the link is opened in Safari, where the customer proxy controls kick in.

> * Opening links inside files in-app: customer does not matter

Yes, because the links are opened in Safari, where - again - the customer proxy controls kick in.

jesmrec:
Good. My last question is:

is an assumption that Safari is always the default browser? just in case, the customers' proxy does not work in other browsers (it's easy to switch the default one).

felix-schwarz:
@jesmrec If a URL is opened externally, it's always opened in the default browser - which, on the devices managed by the client requesting this feature, is Safari.

JohnKuan (approved):

jesmrec:
> @jesmrec If a URL is opened externally, it's always opened in the default browser - which, on the devices managed by the client requesting this feature, is Safari.

thanks for all clarifications @felix-schwarz 

Then, the current and the expected behaviour match.

That's ok for me to go on.

jesmrec (approved):

